PR #17, opened 2019-10-21
imxrt: flash driver.

HBuczynski:


agkaminski:
Move "imxrt: flash driver" to separate branch and then make pull request

PR #29, opened 2020-01-15
Disable UART1 and SPI1 by default. Having both UART1 and UART3 enable…

Maxez:
…d causes kernel not to print init info. Having both SPI1 and SPI3 enabled causes imxrt-flashsrv to not initialize correctly and return with ENXIO error.

### DIFF
--- a/multi/imxrt-multi/config.h
+++ b/multi/imxrt-multi/config.h
@@ -29,7 +29,7 @@
 /* UART */
 
 #ifndef UART1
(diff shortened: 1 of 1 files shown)


agkaminski:
UARTs and SPIs can be disabled by passing e.g. -DSPI1=0 in CFLAGS. If having a SPI enabled causes some code to malfunction, it meas that we need to treat a cause, not a symptom.

PR #32, opened 2020-03-15
flashsrv: added argument check

kemonats:


storage/imx6ull-flash/flashsrv.c
```diff
@@ -660,6 +661,7 @@ static int flashsrv_partition(size_t start, size_t size)
 	TRACE("partition allocated, start: %u, a:t id %d", start, idtree_id(&p->node));
 	mutexUnlock(flashsrv_common.lock);
 
+	free(p);
```
agkaminski: We shouldn't free this partition, it is added to the tree without being copied.

PR #34, opened 2020-03-15
nandtool: options handling changed

kemonats:


pawelpisarczyk:
This tool will be removed and the functionality will be added to the flash driver.

PR #37, opened 2020-04-08
pc-ata: cleaned up driver implementation

Maxez:
Reworked and cleaned up ATA driver implementation.


storage/pc-ata/Makefile
```diff
@@ -4,13 +4,10 @@
...
 
-$(PREFIX_A)libatadrv.a: $(PREFIX_O)storage/pc-ata/atadrv.o
+$(PREFIX_A)libata.a: $(PREFIX_O)storage/pc-ata/ata.o
 	$(ARCH)
 
-$(PREFIX_H)atasrv.h: storage/pc-ata/atasrv.h
-	$(HEADER)
-
-all: $(PREFIX_PROG_STRIPPED)pc-ata $(PREFIX_A)libatadrv.a $(PREFIX_H)atasrv.h
+all: $(PREFIX_PROG_STRIPPED)pc-ata $(PREFIX_A)libata.a
```
agkaminski, line 13: Why we don't want to install header?

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+	uintptr_t reg_base = (uintptr_t)base;
+
+	if (reg_base & 0x1) {
+		/* Read from IO-port */
+		reg_base &= ~0x3;
+		reg_base += reg;
+
+		switch (size) {
+		case 1:
+			return inb((void *)(reg_base));
```
agkaminski, line 46: No need for putting reg_base in brackets. This applies to the whole commit.

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+}
+
+
+static void ata_resetbus(ata_bus_t *bus)
+{
+	void *ctrl = bus->ctrl;
+
+	/* Software reset all drives on the bus */
+	ata_writereg(ctrl, REG_CTRL, CTRL_SOFTRST | CTRL_NIEN, 1);
+	/* Short delay */
```
agkaminski, line 131: This comment is rather redundant ;) I would rather know why we are waiting.

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+static ssize_t ata_pio(ata_dev_t *dev, uint32_t nsectors, uint8_t *buff, uint8_t dir)
+{
+	ata_bus_t *bus = dev->bus;
+	void *base = bus->base;
+	uint32_t i, j;
+	ssize_t err, ret = 0;
+
+	for (i = 0; i < nsectors; i++) {
+		/* Wait until BSY clears and DRQ sets */
+		if ((err = ata_wait(bus, STATUS_BSY, STATUS_DRQ)))
```
agkaminski, line 223: if (err = ata_wait(bus, STATUS_BSY, STATUS_DRQ)) or if ((err = ata_wait(bus, STATUS_BSY, STATUS_DRQ)) != 0)

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+			for (j = 0; j < dev->secsize / 2; j++)
+				ata_writereg(base, REG_DATA, ((uint16_t *)(buff + ret))[j], 2);
+			break;
+		}
+
+		ret += dev->secsize;
+	}
+
+	/* Wait until DRQ clears and RDY sets */
+	if ((err = ata_wait(bus, STATUS_DRQ, STATUS_RDY)))
```
agkaminski, line 244: As above

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+		return err;
+
+	return ret;
+}
+
+
+static ssize_t ata_access(ata_dev_t *dev, uint64_t lba, uint16_t nsectors, uint8_t cmd, uint8_t *buff)
```
agkaminski, line 251: We have convention for functions that should be guarded by mutex outside - these should be prefixed with _
static ssize_t _ata_access(...

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+	}
+
+	/* Wait until BSY clears and DRQ sets */
+	if ((err = ata_wait(bus, STATUS_BSY, STATUS_DRQ)))
+		return err;
+
+	/* Read device info */
+	for (i = 0; i < 256; i++) {
+		data = (uint16_t)ata_readreg(base, REG_DATA, 2);
+		((uint16_t *)info)[i] = ((data & 0xFF) << 8) | ((data >> 8) & 0xFF);
```
agkaminski, line 398: You can cause alignment issue by casting char * to bigger sizeof type. You need to changed endianness anyway, so I don't see any perfomance gain by doing this this way. I would suggest accessing _info_ bytewise

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+	if ((err = ata_wait(bus, STATUS_DRQ, STATUS_RDY)))
+		return err;
+
+	if (dev->type == ATA) {
+		dev->mode = ((info[98] >> 1) & 1) + ((info[166] >> 2) & 1);
+		dev->cylinders = (info[3]   << 0) | (info[2]  << 8);
+		dev->heads     = (info[7]   << 0) | (info[6]  << 8);
+		dev->sectors   = (info[13]  << 0) | (info[12] << 8);
+		dev->secsize   = (info[235] << 0) | 
+						 (info[234] << 8) |
```
agkaminski, line 411: Use spaces for fancy formatting.

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+		case AMODE_CHS:
+			dev->size =
+				(uint64_t)(dev->cylinders) *
+				(uint64_t)(dev->heads) *
+				(uint64_t)(dev->sectors);
+			break;
+
+		case AMODE_LBA28:
+			dev->size =
+				(info[121] <<  0)|
```
agkaminski, line 428: Please add space before | operator

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+	for (i = 0; i < nsectors; i++) {
+		/* Wait until BSY clears and DRQ sets */
+		if ((err = ata_wait(bus, STATUS_BSY, STATUS_DRQ)))
+			return err;
+
+		switch (dir) {
+		case DIR_READ:
+			/* Read sector */
+			for (j = 0; j < dev->secsize / 2; j++)
+				((uint16_t *)(buff + ret))[j] = (uint16_t)ata_readreg(base, REG_DATA, 2);
```
agkaminski, line 230: There are two problems with this code. It depends on endianess of the target and is de facto undefined behavior. Don't cast char * to other types.

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+			break;
+
+		case DIR_WRITE:
+			/* Write sector */
+			for (j = 0; j < dev->secsize / 2; j++)
+				ata_writereg(base, REG_DATA, ((uint16_t *)(buff + ret))[j], 2);
```
agkaminski, line 236: As above

storage/pc-ata/ata.c
```diff
@@ -0,0 +1,533 @@
...
+	}
+
+	if (bus->devs[MASTER] == NULL && bus->devs[SLAVE] == NULL)
+		return -ENXIO;
+
+	if (mutexCreate(&bus->lock) < 0) {
+		if (bus->devs[MASTER] != NULL)
+			free(bus->devs[MASTER]);
+		if (bus->devs[SLAVE] != NULL)
+			free(bus->devs[SLAVE]);
```
agkaminski, line 490: free(NULL) is safe

storage/pc-ata/Makefile
Maxez: We have a circular dependency. Ext2 filesystem depends on PC-ATA (atasrv_read and atasrv_write to be specific). On the other hand PC-ATA driver depends on the filesystem libraries (in our case on libext2.a). The default build order is building filesystems first (PC-ATA makefile haven't triggered yet, the atasrv header is not installed), then the drivers.

agkaminski: Ok. I think that ext2 should not depend on pc-ata or any other storage driver (at least that should be our goal). Ext2 should get read/write callbacks on initialization and use them without knowledge of any particular hardware. @pawelpisarczyk do you agree with me on that?

pawelpisarczyk: Yes

storage/pc-ata/ata.c
Maxez: I tried to go for the first option. The additional parentheses prevent warning when compiling with -Wall (warning: suggest parentheses around assignment used as truth value [-Wparentheses]).
Should I remove them?

agkaminski: Understood. Leave them as-is

storage/pc-ata/ata.c
Maxez: REG_DATA is a 16-bit register. I think I have to read it with inw(), 2 bytes at a time. I tried to use inb() but I receive incorrect data when doing so (also the driver hangs on polling right after - DRQ bit doesn't get cleared and RDY bit doesn't get set).
I'm not sure how I can make this code portable to big endian systems. Perhaps I should use functions like ntohl() for conversion?

agkaminski: No, no, that's not what I meant. I am aware that you need to read 16 bit chunks. I suggest below solution:
info[i] = ((data >> 8) & 0xFF);
info[i + i] = ((data & 0xFF) << 8);

or something like that. So info is only accessed byte-wise.

You'll need to adapt the for loop to this.

Maxez: Ok, I get it now.

storage/pc-ata/ata.c
Maxez: Fixed.

storage/pc-ata/ata.c
Maxez: Fixed.

storage/pc-ata/ata.c
Maxez: Fixed.

storage/pc-ata/ata.c
Maxez: Fixed.

storage/pc-ata/ata.c
Maxez: Fixed.

storage/pc-ata/ata.c
Maxez: Fixed.

storage/pc-ata/ata.c
Maxez: I changed the code to this:
`/* Read one sector */
for (j = 0; j < dev->secsize; j += 2) {
	data = (uint16_t)ata_readreg(base, REG_DATA, 2);
	(buff + ret)[j]       = (data >> 0) & 0xFF;
	(buff + ret)[j + 1] = (data >> 8) & 0xFF;
}`
However, it still doesn't solve the endianess problem. How can I make it endianess independent?

storage/pc-ata/ata.c
Maxez: Fixed.

storage/pc-ata/ata.c
Maxez: I think that the IDE bus through which we transfer the data is little-endian. I've added byte order conversions (to and from little-endian to host byte order) to the register access functions (ata_readreg() and ata_writereg()). I think that the code should work on big-endian systems too.

PR #39, opened 2020-04-23
arm/flashdrv: fix unaligned access to flash_context_t struct

Cinek28:


nalajcie:
Why 8-byte alignment is required? I don't see any field with 8-byte alignment inside nested structs.

IMHO this should not be packed.

agkaminski:
There is very good reason for packing this struct - it is send between processes. We had in the past cases when one process made different paddings in struct.

Real solution, which should be made here - do not change struct definition. Fix all instances of getting pointer to struct member (use for example temporary variable, copy value to it and the pass pointer to this variable, instead of directly do the struct member)

agkaminski:
Comment to future comment: yeah, this struct now has no weird options for padding. But if someone modifies it in the future, it might have.

Cinek28:
All right, will do just that

nalajcie:
> There is very good reason for packing this struct - it is send between processes. We had in the past cases when one process made different paddings in struct.

How this can happen? Assuming the same host architecture (you're talking about processes so I assume we're not sending through the network) I think it's only possible if the other process redefines the type or messes with `#pragma` - and that's just a programming error :).

What I'm trying to tell - using `packed` should be avoided at all costs as it introduces run-time penalty most of the people are not aware of. Consider the following example:
```C
typedef struct {
    uint32_t val;
}  __attribute__((packed)) int_packed_t;

void inc_struct(int_packed_t* x) {
    x->val = x->val + 1;
}
```
On armv7-32 the `inc_struct` function will take **16 assembler instructions**.  Dropping `packed` attribute will make it 4 (ARM gcc 8.2 with `-O2` optimizations; you can fiddle with it here: https://godbolt.org/z/f8dD24) - with absolutely no change to the structure size/layout itself!

The rule of thumb should be to always hand-pack the structs if needed (and if not - just reorder the fields for optimal layout and let compiler handle the rest).

I know that it's out of scope of this fix, but should be taken care of across all our code in the near future.

> Real solution, which should be made here - do not change struct definition. Fix all instances of getting pointer to struct member (use for example temporary variable, copy value to it and the pass pointer to this variable, instead of directly do the struct member)

As You said - there is no padding added in this struct on 32-bit architecture so in this case the `padding` should be dropped (the `sizeof(flash_context_t) = 556` regardless of the `packed` attribute) - it's already well-packed by hand.

Adding fields to this structure in the future should be done with care - but this applies to any code addition/deletion. What I'm saying - let's not harm the run-time performance because we're afraid someone will be a sloppy programmer in the future :).

agkaminski:
Imho sending message should be treated like sending eth packet. We should serialize/deserialize data. Kernel won't do that for you, it knows nothing about this structure layout.

PS we actually should perform endianess management in addition to serialization

nalajcie:
> Imho sending message should be treated like sending eth packet. We should serialize/deserialize data. Kernel won't do that for you, it knows nothing about this structure layout. 

I completely agree, but it doesn't automatically mean we have to use `packed` attribute. Here we're talking about `flash_context_t` struct which I assume is run-time context for the flash operations. The serialization/deserialization pattern should not impact it.

I know that the usual pattern is to `memcpy` the data - which is fine for me - unpacked (or hand-packed) struct will have the same layout in other process/host (assuming it will not use arch-dependant types like `int` which is the case here :)).

I don't have deep knowledge of phoenix-rtos, but I don't see a pattern when serializing by `memcpy` for unpacked structs would fail not due to programming error - maybe You can point me in the right direction?

Sorry for being stubborn, this has unwillingly changed into discussion not directly related to this PR - I've just seen the solution which might not be valid (just silences the compiler) to the problem which should not exist in the first place :). My concern is not related to this PR, just to the general code quality.

> PS we actually should perform endianess management in addition to serialization.

:+1: That would be nice to have.

Maybe it would be enough to have functions for message serialization/deserialization as part of the API (some of the modules already do it by providing "message sending wrapper" lib with function interface). If the intended use cases include sending messages between different archs - they should handle endianess internally.

> I am talking about interfaces. Inside of the process - yeah, I agree, packing is often unnecessary and causes performance penalty

Yep, interfaces should be well defined to be portable - but it's enough to use const-length types inside the struct (they have the same alignment requirements on every arch, so the padding would be the same). If You use pointers it won't be possible to pass them to other host without custom serialization/deserialization pattern so the struct won't be directly memcopied but reconstructed differently in the recipient.

pawelpisarczyk:
Guys, what about this?

Cinek28:
The flash_context_t has been changed so that packed attribute is no longer needed. Closing pull request.

PR #53, opened 2020-08-26
add uart16550 (wip)

youki0:


pawelpisarczyk:
Julia, please do it in another way. The function for reading UART registers should be introduced and I recommend to define it in separate file specific for the target board e.g. uart-riscv.c, uart-pc.c. This module should be responsible for low-level initialization of device register access and should provide specific descriptor for upper layers to let them operate on the device registers. Functions like iopl() or mmap() should be located and used in this file.

PR #54, opened 2020-08-30
uart16550: fixed uart_get()

kemonats:
Now:
- the function returns 0 instead of the size of the array if neither condition is true
- wrong calculation of the number of array elements

I suggest to use an expression 
`
sizeof(array) / sizeof(array[0])
`
to calculate the size of the array

### DIFF
--- a/tty/uart16550/uart16550.c
+++ b/tty/uart16550/uart16550.c
@@ -165,7 +165,7 @@ uint8_t uart_get(oid_t *oid)
 {
 	unsigned int i;
 
-	for (i = 0; i < sizeof(uarts) / sizeof(uart_t); i++) {
+	for (i = 0; i < sizeof(uarts) / sizeof(uart_t *); i++) {
 		if ((uarts[i]->oid.id == oid->id) && (uarts[i]->oid.port == oid->port))
 			return i;
 	}
@@ -315,7 +315,7 @@ int main(void)
 	for (n = 0; n < 4; n++) {
 		if (_uart_init(n, B115200, &uarts[n]) < 0)
 			continue;
-		
+
 		if (portRegister(port, "/dev/ttyS0", &uarts[n]->oid) < 0) {
 			fprintf(stderr, DRIVER ": Can't register port %d\n", port);
 			return -1;


pawelpisarczyk:
Your PR is partially correct. Please leave these returns in uart_get() without changes because it was incorrectly assumed in WRITE and READ messages handling that uart_get() function should always return valid index even when oid points to uart which is not present.

tty/uart16550/uart16550.c
```diff
@@ -165,11 +165,11 @@ uint8_t uart_get(oid_t *oid)
 {
 	unsigned int i;
 
-	for (i = 0; i < sizeof(uarts) / sizeof(uart_t); i++) {
+	for (i = 0; i < sizeof(uarts) / sizeof(uart_t *); i++) {
```
agkaminski, line 168: Perhaps sizeof(uarts) / sizeof(uarts[0])?

pawelpisarczyk: This part proposed by kemonats is good. I commented these returns. He changed first return with break and proposed to changed the return 0 located at the function end with return i. This change will lead us to incorrect behavior when wrong oid will be passed in WRITE and READ messages.

kemonats:
> Your PR is partially correct. Please leave these returns in uart_get() without changes because it was incorrectly assumed in WRITE and READ messages handling that uart_get() function should always return valid index even when oid points to uart which is not present.

What will happen in this situation

`uarts[0]->oid.id != oid.id` or `uarts[0]->oid.port != oid.port`

`uart_get(`) will return `0` and `uarts[0]` will be used.
Should it be so or should a bug be reported?

The sizeof(uarts) / sizeof(uart_t) error also occurred in pc_uart.
Function `uart_get()` always returned 0.
Therefore, it is safer to use 
`sizeof(array) / sizeof(array[0])`
 to calculate the size of the array.

pawelpisarczyk:
Kemonats, you are right that current division is incorrect. I agree with you. The problem is in changing the semantics of uart_get(). It was in the past incorrectly assumed that this function every time returns correct index of uart structure. If you like to change this and you provide error handling in WRITE, READ message handling please don't hesitate :)

pawelpisarczyk:
I've corrected uart_get() according to kemonats suggestions. I've add error handling when no uart is found.

PR #58, opened 2020-09-14
stm32lx-multi: Add DMA support, make SPI use DMA

Darchiv:
This change introduces DMA to stm32lx-multi driver (both STM32L1 and STM32L4 are supported). Currently only SPI peripheral is handled, avoiding the overhead of sharing DMA channels between multiple peripherals. DMA transfers are used implicitly by the driver when a certain transfer length threshold is exceeded - processes using multi driver have no notion of DMA at all.

Hardware NSS output is now enabled, eliminating the need of managing NSS GPIO in software. The pin still needs to be explicitly mapped as alternate function, so the change is backward compatible. NSS is always asserted when SPI peripheral is enabled, so the enable/disable part has to be done directly before and after the transaction.

The following enhancements, which may be added to this request, are to be discussed:

- Replace active-wait for the interrupt-set flag `ready` in `_spi_readwrite()` with waiting for `SPIx_SR.RXNE==1` directly. Active-wait is already employed, and this enhancement will remove interrupt overhead completely.
- For time-critical operations, allow multiple transactions to be scheduled in one multi driver message (by e.g. an array of buffer pointers in `multi_i_t`'s union). This would significantly reduce the time required to complete small consecutive transactions without having to bypass the driver's message IPC.
- The `spi_address16msb` flag was added to allow using 2-byte addresses before transactions and to not break compatibility with existing software using `spi_address` alone. If this compatibility is not needed, `spirw_t.flags` could have a `addrlen` subfield instead of `spi_address*` to make the addressing feature more flexible.

### DIFF
--- a/multi/stm32l1-multi/stm32-multi.h
+++ b/multi/stm32l1-multi/stm32-multi.h
@@ -162,7 +162,11 @@ typedef struct {
 enum { spi1 = 0, spi2, spi3 };
 
 
-enum { spi_cmd = 0x1, spi_address = 0x2, spi_dummy = 0x4 };
+#define SPI_ADDRSHIFT 3
+#define SPI_ADDRMASK 0x3
+
+
+enum { spi_cmd = 0x1, spi_dummy = 0x2, /* 3-bits for SPI_ADDR* ,*/ spi_addrlsb = 0x20 };
 
 
 typedef struct {
--- a/multi/stm32l4-multi/stm32-multi.h
+++ b/multi/stm32l4-multi/stm32-multi.h
@@ -113,7 +113,11 @@ typedef struct {
 enum { spi1 = 0, spi2, spi3 };
 
 
-enum { spi_cmd = 0x1, spi_address = 0x2, spi_dummy = 0x4 };
+#define SPI_ADDRSHIFT 3
+#define SPI_ADDRMASK 0x3
+
+
+enum { spi_cmd = 0x1, spi_dummy = 0x2, /* 3-bits for SPI_ADDR* ,*/ spi_addrlsb = 0x20 };
 
 
 typedef struct {
--- a/multi/stm32l1-multi/Makefile
+++ b/multi/stm32l1-multi/Makefile
@@ -4,7 +4,7 @@
 # Copyright 2018 Phoenix Systems
 #
 
-MULTIDRV_OBJS = stm32-multi.o uart.o rcc.o gpio.o adc.o i2c.o lcd.o rtc.o flash.o spi.o exti.o
+MULTIDRV_OBJS = stm32-multi.o uart.o rcc.o gpio.o adc.o i2c.o lcd.o rtc.o flash.o spi.o exti.o dma.o
 
 $(PREFIX_PROG)stm32-multi: $(addprefix $(PREFIX_O)multi/stm32l1-multi/, $(MULTIDRV_OBJS))
 	$(LINK)
--- a/multi/stm32l1-multi/README.md
+++ b/multi/stm32l1-multi/README.md
@@ -243,22 +243,25 @@ Structure of below format:
 
 	typedef struct {
 		int spi;
-		char cmd;
 		unsigned int addr;
 		unsigned int flags;
+		char cmd;
 	} spirw_t;
 
 Is used for read/write transactions with SPI device.
 
 - spi - from pool of enum { spi1 = 0, spi2, spi3 };
 - cmd - command to SPI device, first byte always written during transaction
-- addr - address, second and third byte written during transaction 
-- flags - bit mask of enum { spi_address = 0x1, spi_dummy = 0x2 };
+- addr - address to be written after cmd during transaction
+- flags - bit mask of enum { spi_cmd, spi_dummy, spi_addrlsb } and/or (addrsz << SPI_ADDRSHIFT)
+
 
 where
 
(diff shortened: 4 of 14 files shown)


agkaminski:
1. That's fine for me.
2. I think this complicates interface without real benefit. You can prepare longer buffer with all needed transactions within it.
3. It is needed - used by meterfs

agkaminski:
Does DMA have advantage for bufflens >= 2 over active waiting/interrupt?

Darchiv:
With bufflen == 2, complete transaction time (NSS low time) is:
- with DMA: 80 µs
- without DMA: 390-520 µs

So yes, it has.

Darchiv:
>     2. I think this complicates interface without real benefit. You can prepare longer buffer with all needed transactions within it.

The problem is that a transaction is marked by high-to-low and low-to-high edges of NSS. These edges **must** be present, so simply passing a longer buffer is not an option. If the driver is to support multi-transactions, there has to be a way to infer the lengths of individual transactions, e.g. 1) by prepending each transaction in a buffer with a length field, and passing such a lengthy buffer (not nice), or 2) passing an array of pointers (still needs lengthy buffer(s)). Is this even viable?

>     3. It is needed - used by meterfs

So the `spi_address` has to remain. But `spi_address16msb` could be replaced with a 2-bit "field": 0x0 = 3 bytes (compatibility), 0x1 = 1 byte, 0x2 = 2 bytes, 0x3 = 3 bytes, no `spi_address` flag -> 0 bytes.

agkaminski:
There are even devices with 4 byte address, so it would be better to handle them too.

It would be better to patch meterfs rather than making this 0x0 compability. This field could indicate number of address bytes (including 0 bytes, no address). Perhaps one bit extra for lsb/msb.

Darchiv:
Two solutions for the address size were considered: an extra field in `spirw_t`, `flags` as bit-fields, and masking the multi-bit size in `flags`. The former two would require changes in every application making use of SPI messages. The last one would not - only meterfs-srv has to be patched (https://github.com/phoenix-rtos/phoenix-rtos-filesystems/pull/21) - and thus I have incorporated the address size "field" into `flags`.

multi/stm32l4-multi/stm32-multi.h
```diff
@@ -113,7 +113,7 @@ typedef struct {
 enum { spi1 = 0, spi2, spi3 };
 
 
-enum { spi_cmd = 0x1, spi_address = 0x2, spi_dummy = 0x4 };
+enum { spi_cmd = 0x1, spi_dummy = 0x2, spi_addrshift = 2, spi_addrmask = 0x7, spi_addrlsb = 0x20 };
```
agkaminski, line 116: spi_addrshift and spi_addrmask should not really be a part of the same enum as other flags - these two symbols have very different meaning. I think it is justified to make them defines; SPI_ADDRSHIFT and SPI_ADDRMASK

multi/stm32l1-multi/stm32-multi.h
```diff
@@ -162,7 +162,7 @@ typedef struct {
 enum { spi1 = 0, spi2, spi3 };
 
 
-enum { spi_cmd = 0x1, spi_address = 0x2, spi_dummy = 0x4 };
+enum { spi_cmd = 0x1, spi_dummy = 0x2, spi_addrshift = 2, spi_addrmask = 0x7, spi_addrlsb = 0x20 };
```
agkaminski, line 165: spi_addrshift and spi_addrmask should not really be a part of the same enum as other flags - these two symbols have very different meaning. I think it is justified to make them defines; SPI_ADDRSHIFT and SPI_ADDRMASK

agkaminski (changes requested):

multi/stm32l4-multi/stm32-multi.h
Darchiv: I agree that they have different meaning, but should we at least insert something like a `/* SPI_ADDR */` placeholder comment to explicitly mark these bits reserved or is `spi_addrlsb = 0x20` enough to deter one from using these?

multi/stm32l1-multi/stm32-multi.h
Darchiv: I agree that they have different meaning, but should we at least insert something like a `/* SPI_ADDR */` placeholder comment to explicitly mark these bits reserved or is `spi_addrlsb = 0x20` enough to deter one from using these?

agkaminski: Sure, that's great idea to place comment there.

multi/stm32l4-multi/stm32-multi.h
Darchiv: Done

multi/stm32l1-multi/stm32-multi.h
Darchiv: Done

Darchiv:
I have added a few missing `volatile` keywords, taken care of multi-byte FIFOs of L4, and removed hardware management of NSS. It is impossible to assert NSS in hardware without creating any issues. In a simple single-master arrangement the NSS signal is always asserted when the peripheral is enabled and high-Z otherwise. SPI lines (MOSI, etc.) are also put in high-Z when the peripheral is disabled, so obscure glitches can happen when NSS is not used (e.g. with GPIO SS).

This PR seems to be complete and mergeable, along with mentioned PRs.

PR #68, opened 2020-10-14
imxrt-multi: added uart.c dependency on libtty.h

kemonats:
This fixes the error that is possible at compile time if the `-j` option is used.


nalajcie (approved):
IMHO we should fix (again) the build system to autogenerate and import build dependencies, but this is good as a stop-gap for now.

PR #71, opened 2020-11-19
imxrt106x: include adc & oled driver

HBuczynski:


nalajcie:
I see that these drivers will be also built for armv7m7-imxrt106x (evalkit) target on phoenix-rtos-project. Is this desirable? Shouldn't we have some mechanism to select which drivers should be compiled for each target?

(this is not a topic or blocker for this PR, just a generic discussion - do You maybe have an idea how to do it in a generic way?)

HBuczynski:
One of the solution might be declaring variables in `phoenix-rtos-project` in target's file:  https://github.com/phoenix-rtos/phoenix-rtos-project/blob/master/_targets/build.project.armv7m7-imxrt106x. Based on these parameters, specific drivers would be then chosen and compiled.

@nalajcie what do you think about it ?

pawelpisarczyk:
oled and adc sdrivers are specific for meter project and they should be compiled using make <driver name> invocation in _targets/<target> file specific for the project.

PR #72, opened 2020-11-19
flashdrv: fix long busy wait on flash write

Cinek28:


nalajcie:
it would be great if commit message had some indication regarding the platform which is affected by the changes

PR #73, opened 2020-11-19
imxrt/ad7779: add new capabilities to device api

Cinek28:
Added device error status check, hardware reset,
unused channels disabling. Fixed memory deallocation
when initialization is unsuccessful.

### DIFF
--- a/adc/ad7779/ad7779-driver.c
+++ b/adc/ad7779/ad7779-driver.c
@@ -47,6 +47,8 @@
 #define ADC_BUFFER_SIZE             (4 * AD7779_NUM_OF_CHANNELS * _PAGE_SIZE)
 #define NUM_OF_BUFFERS              (16)
 
+#define MAX_INIT_TRIES              (4)
+
 typedef volatile struct {
 	uint32_t VERID; //0x00
 	uint32_t PARAM; //0x04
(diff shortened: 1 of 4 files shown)


adc/ad7779/ad7779.c
```diff
@@ -632,55 +672,95 @@ static int ad7779_gpio_init(void)
...
+
+	do {
+		if (!ad7779_get_status(status)) {
+			if (status[16] & 0x10)
+				return AD7779_OK;
+		}
+
+		++iter;
+		usleep(100000);
+	} while (iter < 4);
```
agkaminski, line 757: Wouldn't `for` loop be better here?

adc/ad7779/ad7779-driver.c
```diff
@@ -682,8 +696,8 @@ int main(void)
 	while (lookup("/", NULL, &root) < 0)
 		usleep(10000);
 
-	if (init())
-		return -EIO;
+	while (init())
+		usleep(100000);
```
agkaminski, line 700: Maybe we should give up after N tries?

adc/ad7779/ad7779-driver.c
```diff
@@ -436,13 +432,7 @@ static int dev_read(void *data, size_t size)
...
 		return -EIO;
 
-	mutexLock(common.irq_lock);
-
-	condWait(common.irq_cond, common.irq_lock, 0);
-
-	*(uint32_t **)data = ((uint32_t *)&common.current);
-
-	mutexUnlock(common.irq_lock);
+	*(uint32_t **)data = (uint32_t *)&common.current;
```
agkaminski, line 435: So we don't need synchronization with the interrupt via conditional?

adc/ad7779/ad7779.c
Cinek28: Right, for loop would do better. Will fix that.

adc/ad7779/ad7779-driver.c
Cinek28: All right, will change that.

adc/ad7779/ad7779-driver.c
Cinek28: Well, with synchronization via conditional, the driver caused the thread reading it to be blocked
when no data was received or there were communication problems with AD7779. The common.current
is a read-only variable indicating the number of buffers read from the start. I have another concern here.
As it is passed by pointer (in order to send message only once and then simply checking the value), maybe
it would be more secure to pass it as value (but then again, the value will need to be checked by message
each time)?

adc/ad7779/ad7779-driver.c
Cinek28: Done

adc/ad7779/ad7779.c
Cinek28: Done

adc/ad7779/ad7779-driver.c
Cinek28: Added timeout on cond_wait

PR #74, opened 2020-11-20
imxrt106x: add BOARD_CONFIG to CFLAGS

HBuczynski:


nalajcie:
please move this to the main Makefile, as the current approach is misleading (the `multi/imxrt-multi/Makefile` is included in the main Makefile so the `CFLAGS` will actually be changed for all devices compilation).

Makefile
```diff
@@ -9,6 +9,7 @@
 SIL ?= @
 MAKEFLAGS += --no-print-directory
 
+
```
nalajcie, line 12: why? :P

nalajcie (approved):

nalajcie:
As this is a not-destructive change, I'm merging this right away

Makefile
HBuczynski: I moved `CFLAGS` below line `include ../phoenix-rtos-build/Makefile.common`
Currently, in Makefile.common `CFLAGS` are reset.

This new line is a residue :P

PR #78, opened 2020-12-18
imxrt-multi: UART baudrate calculation based on sdk

gerard5:
The reason for the change is that with 230k and 460k and sending long frames I continuously observed numerous system freezes, after changing to the algorithm proposed the problem disappeared.

I've changed the UART baud rate calculation to an idea taken from the SDK. Please take a look at the attached comparison table below. I've checked the algorithm on hardware with ~500 octet long frames (the most common baud rates I use are 230400 and 460800).
Additionally, I added a line that disables TX and RX interrupts like in imx6ull (sticky conds), out of the condWait loop.

https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/05008b584058340204413cba3ef27a4fe55a5d0c/tty/imx6ull-uart/imx6ull-uart.c#L161-L168

Maybe this will help to stop receiving FE and Noisy interrupts (RX), even if they are disabled in ctrlr register (bit 25, 26).

image

### DIFF
--- a/multi/imxrt-multi/uart.c
+++ b/multi/imxrt-multi/uart.c
@@ -182,33 +182,37 @@ static void set_cflag(void *_uart, tcflag_t* cflag)
 
 static uint32_t calculate_baudrate(speed_t baud)
 {
-	int osr, sbr, bestSbr = 0, bestOsr = 0, bestErr = 1000, t, baud_rate = libtty_baudrate_to_int(baud);
+	uint32_t osr, sbr, bestSbr = 0, bestOsr = 0, bestDiff, t, tDiff, baud_rate;
 
-	if (!baud_rate)
+	bestDiff = baud_rate = (uint32_t)libtty_baudrate_to_int(baud);
+
+	if ((int)baud_rate <= 0) {
 		return 0;
-
-	for (osr = 3; osr < 32; ++osr) {
-		sbr = UART_CLK / (baud_rate * (osr + 1));
-		sbr &= 0xfff;
-		t = UART_CLK / (sbr * (osr + 1));
-
-		if (t > baud_rate)
-			t = ((t - baud_rate) * 1000) / baud_rate;
-		else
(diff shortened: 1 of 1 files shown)


multi/imxrt-multi/uart.c
```diff
@@ -121,6 +121,9 @@ static void uart_intrThread(void *arg)
 			condWait(uart->cond, uart->lock, 0);
 		}
 
+		/* diable RX & TX interrupts */
+		*(uart->base + ctrlr) &= ~((1 << 23) | (1 << 21));
+
```
agkaminski, line 126: I don't get why you would want to disable RX irq. Where is it reenabled?

gerard5: Don't know, RX and TX are disabled in interrupt handler, after returning from IRQ handler UART thread wakes up. Somehow in handling the IRQ, if the framing error and noise bits are set (both not one), the RX interrupt does not turn off, and if a lot of data comes quickly, there is a deadlock between the IRQ and UART thread, the libtty does not empty the FIFO on time and overflow occurs, RX IRQ arrives with a framing error and noisy flags set, RX can't be turned off in the interrupt there, wake up the thread. And again, again. System dies in a loop.

This was an attempt to limit this behavior, before rewriting the baudrate calculation (I noticed that the deviation from the nominal 460k and 230k speeds I use may introduce framing errors), I left this line for discussion, some sort of solution to the problem I described. Currently the behaviour has gone with proposed baud rate calculation, but I'm using 230k/460k most of the time. Similar problems may appear in the case of opto-ports, but at lower speeds, e.g. 9600 or 19200, and deviations from the nominal transmission speed of foreign devices (readers, optical heads). I can remove this line (as I also see the typo in the comment :)

gerard5:
I've removed the line disabling TX/RX interrupt outlined above for now and added similar change to plo here https://github.com/phoenix-rtos/plo/pull/24

PR #79, opened 2021-01-25
stm32l4-multi: Fix building race, simplify Makefile

anglov:
Rarely build fails due to following error:

> CC  multi/stm32l4-multi/lib/libuart.c
> CC  multi/stm32l4-multi/stm32-multi.c
> In file included from multi/stm32l4-multi/rtc.h:18,
>                  from multi/stm32l4-multi/stm32-multi.c:32:
> multi/stm32l4-multi/stm32-multi.h:18:10: fatal error: stm32l4-multi/libuart.h: No such file or directory
>    18 | #include <stm32l4-multi/libuart.h>
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~
> 

It's because header installation is not forced before building sources. 
I added explicit dependency between MULTIDRV_OBJS and SRCHEADERS. It's not ideal solution as when we apply automatic header dependency tracking (gcc -M  -MD -MP -MF) this hardcoded dependency will pull all headers . In the other hand, I suppose more changes will be needed in that case. For now, I don't see any better alternatives to fix the race.

I also think about following changes:
in sources:
rename lib to libmulti
in install dir
rename include/stm32l4x6-multi/libXYZ.h to include/libmulti/libXYZ.h

pros
- will be similar to stm32-multi.h header (we don't have different stm32l4x6-multi.h and stm32l15x-multi.h headers) 
- simplify Makefile
- simplify rules to IDE where to search headers
cons
- includes need to be refactored

If you agree I could do changes.


agkaminski (approved):

agkaminski:
I'm ok with your idea about changes with headers location

anglov:
Ok, I provided suggested changes.

PR #82, opened 2021-02-11
Fix usb on imxrt106x, add notifications and event handling

gerard5:
This change fixes `usb-cdc` and `usb-hid` operation on `imx rt1064-evk`.

Added:
- `PHY` initialization in device mode and
- replaced `OCRAM2` usage with uncached memory - `DTCM`
- basic event and notification handling (connected, disconnected, reset, class setup, etc)
- class setup was moved from desc_manager to CDC and HID client respectively.
- minor bug fixes


usb/imx-usbc/client.c
```diff
@@ -132,6 +131,22 @@ static void usbclient_irqThread(void *arg)
...
+		/* Initialize endpoints */
+		if (imx_common.dc.op == DC_OP_INIT) {
+			ctrl_initQtd();
+			for (i = 1; i < ENDPOINTS_NUMBER; ++i) {
+				if (ctrl_endptInit(i, &imx_common.data.endpts[i]) < 0) {
+					imx_common.dc.op = DC_OP_INIT_ERR;
+					break;
+				}
+			}
+			if (imx_common.dc.op != DC_OP_INIT_ERR) {
```
HBuczynski, line 144: I think that user should get information that endpoints initialization has failed. What do you think about invoking event with appropriate information?

usb/imx-usbc/desc_manager.c
```diff
@@ -216,27 +219,35 @@ static int desc_ReqGetConfig(const usb_setup_packet_t *setup)
...
+		if ((setup->wValue & 0xff) == 0) {
+			strDesc = (usb_string_desc_t *)desc_common.str0;
+			ctrl_execTransfer(0, desc_common.str0, MIN(strDesc->bLength, setup->wLength), USB_ENDPT_DIR_IN);
+		}
+		else if ((setup->wValue & 0xff) == 1) {
+			strDesc = (usb_string_desc_t *)desc_common.strMan;
+			ctrl_execTransfer(0, desc_common.strMan, MIN(strDesc->bLength, setup->wLength), USB_ENDPT_DIR_IN);
+		}
+		else if ((setup->wValue & 0xff) == 2) {
+			strDesc = (usb_string_desc_t *)desc_common.strProd;
```
HBuczynski, line 240: In targets with MMU it doesn't work. `desc_common.str0`, `desc_common.strMan` and `desc_common.strProd` point to physical addresses. The best solution would be keep virtual addresses and make translation to physical address while transfer is executed.

usb/imx-usbc/controller.c
```diff
@@ -297,7 +327,7 @@ int ctrl_hfIrq(void)
 	int endpt = 0;
 
 	if ((ctrl_common.dc->setupstat = *(ctrl_common.dc->base + endptsetupstat)) & 0x1) {
-		/* trip winre set */
+		/* Trip wire set */
 		while (!((ctrl_common.dc->setupstat >> endpt) & 1))
 			endpt++;
```
HBuczynski, line 332: Usually host is responsible for initiating transmission with the device. So that, there is a check in line 299 whether setup transaction is received on control endpoint and then other endpoints are checked in the while loop.
I think that there is very little chance of that interrupts is received and no setup packet is transmitted on any of the endpoints.

However, if we ended up with endpt > 15, there would be definitely a bad memory access. For this reason it would be safer to check endpt value before using it.

usb/imx-usbc/client.c
gerard5: Good point, either/or:
- `cdc_init ()` should wait for this event (with a reasonable timeout e.g. 200ms) and succeed/fail upon receiving a success/failure event after completing the initialization step (not yet descriptor exchange between the host and device), thread worker need to exit, that way `cdc_init()` may call `cdc_destroy()` in itself, and return error.
- `cdc_init ()` works as it is now, but only the appropriate event is sent to the user.

The first approach may be better but requires an another (separate) PR, the second is simpler to implement and may be an amendment to the current PR, but from my personal point of view the first is how to do it (the right way), what do you think?

usb/imx-usbc/desc_manager.c
gerard5: That's right, I definitely need to fix this.

usb/imx-usbc/controller.c
gerard5: Oops. You're right, it could go crazy.

usb/imx-usbc/client.c
HBuczynski: I agree with you. The first approach covers case when the device is not connected to the host and this solution with timeout is recommended.
If you consider that this approach is more complicated you can add the second one with appropriate information in comment that it needs to be changed in the future.

usb/imx-usbc/controller.c
gerard5: Done, also added USB interrupt clear routine.

usb/imx-usbc/client.c
gerard5: Done, according to the first approach.

usb/imx-usbc/desc_manager.c
gerard5: Because `va2pa()` cannot be used in the interrupt handler, I used an intermediate container with a physical address and a virtual address as a pointer to usb_string_desc_t.

agkaminski (approved):

PR #83, opened 2021-02-15
imxrt-multi: add uart rx overrun, noisy and framing error handling

gerard5:
During debugging with jtag, when the processor is halted and receiving data with UART (e.g. keyboard typing on the console serial port)  - writing to the hardware FIFO (more characters than the FIFO size) causes the UART to overrun and locks it (until the  FIFO register is flushed and status flag is cleared by W1C), so no more data can be received after the processor is resumed - the attached commit resolves this problem, additionally adds clear status flags for noisy and framing errors.
A similar change is also added to the plo (separate PR):
https://github.com/phoenix-rtos/plo/pull/28

### DIFF
--- a/multi/imxrt-multi/uart.c
+++ b/multi/imxrt-multi/uart.c
@@ -87,7 +87,17 @@ static int uart_handleIntr(unsigned int n, void *arg)
 {
 	uart_t *uart = (uart_t *)arg;
 
-	*(uart->base + ctrlr) &= ~((1 << 23) | (1 << 21));
+	*(uart->base + ctrlr) &= ~((1 << 27) | (1 << 26) | (1 << 25) | (1 << 23) | (1 << 21));
+
+	/* RX overrun - invalidate fifo and clear status flag */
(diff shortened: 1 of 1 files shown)


multi/imxrt-multi/uart.c
```diff
@@ -87,7 +87,17 @@ static int uart_handleIntr(unsigned int n, void *arg)
...
+	if (*(uart->base + statr) & (1 << 19)) {
+		*(uart->base + fifor) |= (1 << 14);
+		*(uart->base + statr) |= (1 << 19);
+	}
+
+	/* clear RX framing and noise error */
+	if (*(uart->base + statr) & (1 << 18) | (1 << 17))
+		*(uart->base + statr) |= (1 << 18) | (1 << 17);
 
 	return 1;
```
agkaminski, line 102: Maybe we should wake up interrupt thread if we handled an error (return -1 that case)?

gerard5: What about flushing the RX FIFO and then return -1, in the quoted above case FIFO still contains data.

PR #87, opened 2021-03-04
imxrt/ad7779: change initialization and hardware reset pinout

Cinek28:


agkaminski:
Can You connect it with some JIRA task?

Cinek28:
Created and added jira task label

PR #88, opened 2021-03-15
imxrt: add optional baudrate defines

Cinek28:
The default baudrate equals 115200. Optional values set as UARTX_BAUDRATE.

### DIFF
--- a/multi/imxrt-multi/uart.c
+++ b/multi/imxrt-multi/uart.c
@@ -680,6 +680,8 @@ int uart_init(void)
 
 	uart_initPins();
 
+	const speed_t default_baud[] = { UART_BAUDRATES };
+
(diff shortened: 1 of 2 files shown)


multi/imxrt-multi/uart.c
```diff
@@ -680,6 +680,8 @@ int uart_init(void)
 
 	uart_initPins();
 
+	const speed_t default_baud[] = { UART_BAUDRATES };
```
kemonats, line 683:
```suggestion
	const speed_t default_baud[sizeof(uartConfig) / sizeof(uartConfig[0])] = { UART_BAUDRATES };
```
If you don't use _Static_asset(), an explicit size array declaration might be better in this case .
The same for `info[]` array.

agkaminski: Added an issue (https://github.com/phoenix-rtos/phoenix-rtos-devices/issues/89) as this is not the scope of this PR

PR #92, opened 2021-03-29
ad7779: fix log types and levels

Cinek28:
JIRA: BES-82

### DIFF
--- a/adc/ad7779/ad7779-driver.c
+++ b/adc/ad7779/ad7779-driver.c
@@ -36,8 +36,14 @@
 #define COL_CYAN    "\033[1;36m"
 #define COL_NORMAL  "\033[0m"
 
-#define LOG_TAG "ad7779-drv: "
+#define LOG_TAG "ad7779: "
+
+#ifdef NDEBUG
+#define log_debug(fmt, ...)
+#else
 #define log_debug(fmt, ...)     do { printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
+#endif
+
 #define log_info(fmt, ...)      do { printf(LOG_TAG COL_CYAN fmt COL_NORMAL "\n", ##__VA_ARGS__); } while (0)
 #define log_error(fmt, ...)     do { printf(LOG_TAG COL_RED fmt COL_NORMAL "\n", ##__VA_ARGS__); } while (0)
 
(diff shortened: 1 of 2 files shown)


adc/ad7779/ad7779-driver.c
```diff
@@ -36,8 +36,14 @@
 #define COL_CYAN    "\033[1;36m"
 #define COL_NORMAL  "\033[0m"
 
-#define LOG_TAG "ad7779-drv: "
-#define log_debug(fmt, ...)     do { printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
+#define LOG_TAG "ad7779: "
+
+#if !NDEBUG
```
nalajcie, line 41: use `#ifndef` or `#if !defined(NDEBUG)`. This code would fail with `-Wundef` compilation parameter.

I would switch the contents of if-else not to use "double negation" (right now it's "if not no debug")

nalajcie (changes requested):

adc/ad7779/ad7779-driver.c
Cinek28: Done

adc/ad7779/ad7779-driver.c
```diff
@@ -36,8 +36,14 @@
...
 #define COL_NORMAL  "\033[0m"
 
-#define LOG_TAG "ad7779-drv: "
-#define log_debug(fmt, ...)     do { printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
+#define LOG_TAG "ad7779: "
+
+#ifdef NDEBUG
+#define log_debug(fmt, ...)
+#else
+#define log_debug(fmt, ...)	 do { printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
```
nalajcie, line 44: [nitpicking] tab instead of spaces before `do` which makes it not aligned with `do`'s below.

Not a merge blocker but if You have the strength to fix it - then do so. Otherwise LGTM.

(btw, the same has happened in ad7779.c with `log_info` and `log_error` some time ago)

Cinek28: Done

nalajcie: thanks for the fixes

nalajcie (approved):

PR #93, opened 2021-04-07
host-flash: improved hostflash_write() function

Diabliz:


storage/host-flash/host-flash.c
```diff
@@ -51,28 +52,37 @@ ssize_t hostflash_read(unsigned int addr, void *buff, size_t bufflen)
...
+
+        stat = hostflash_read(addr + wrote, tempTab, chunkLen);
         if (stat <= 0)
             return stat;
 
-        currentByte = currentByte & *((char *)buff + wrote);
-        stat = pwrite(hostflash_common.filefd, &currentByte, 1, addr + wrote);
+        for (i = 0; i < stat; i++) {
+            tempTab[i] = tempTab[i] & *((char *)buff + wrote + i);
+        }
```
HBuczynski, line 75: According to our coding convention:
_Opening and closing braces should be used only if the body of the conditional instruction is longer than one line_

storage/host-flash/host-flash.c
```diff
@@ -51,28 +52,37 @@ ssize_t hostflash_read(unsigned int addr, void *buff, size_t bufflen)
...
 
     if (addr + bufflen > hostflash_common.flashsz)
         return -EINVAL;
 
     while (wrote != bufflen) {
-        stat = hostflash_read(addr + wrote, &currentByte, 1);
+        if ((bufflen - wrote) / sizeof(tempTab))
+            chunkLen = sizeof(tempTab);
+        else
+            chunkLen = (bufflen - wrote) % sizeof(tempTab);
```
HBuczynski, line 67: What do you think about defining size of `tempTab` as local const variable or define? Then you don't have to invoke to `sizeof` function each time.

storage/host-flash/host-flash.c
```diff
@@ -51,28 +52,37 @@ ssize_t hostflash_read(unsigned int addr, void *buff, size_t bufflen)
...
         if (stat <= 0)
             return stat;
 
-        currentByte = currentByte & *((char *)buff + wrote);
-        stat = pwrite(hostflash_common.filefd, &currentByte, 1, addr + wrote);
+        for (i = 0; i < stat; i++) {
+            tempTab[i] = tempTab[i] & *((char *)buff + wrote + i);
+        }
+
+        stat = pwrite(hostflash_common.filefd, &tempTab, stat, addr + wrote);
```
anglov, line 77: When pwrite return less bytes than requested (eg.  interrupted by signal), you obviously return to writing data, but will do additional read and once more you will anding readed values with data to write. I think you should avoid unnecesarry operations  by writing chunk to the end if something will interrupt.

storage/host-flash/host-flash.c
nalajcie: just FYI: sizeof is not a function, it's an operator which is resolved at compile time to constant value - so in terms of performance it won't get any faster.

I would advise to introduce intermediate value only if it improves readability, eg doing repeatedly:
```c
// multiple similar loops
for (unsigned int i = 0; i < sizeof(tab) / sizeof(tab[0]); ++i)
    // [...]
```
I would change it to
```c
const unsigned int tabLen =  sizeof(tab) / sizeof(tab[0]);

for (unsigned int i = 0; i < tabLen; ++i)
    // [...]
```

In the above case the performance would be exactly the same (const value constraint at compile time)

HBuczynski: OK

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
```diff
@@ -51,28 +52,39 @@ ssize_t hostflash_read(unsigned int addr, void *buff, size_t bufflen)
...
+            return readed;
+
+        for (i = 0; i < readed; i++)
+            tempTab[i] = tempTab[i] & *((char *)buff + wrote + i);
+
+        len = 0;
+        while (len != readed) {
+            stat = pwrite(hostflash_common.filefd, &tempTab, readed - len, addr + wrote + len);
+            if (stat < 0) {
+                if (errno == -EINTR)
```
gerard5, line 80:
```suggestion
                if (errno == EINTR)
```
`errno` is defined by the ISO C standard, valid error numbers are all positive numbers.

Diabliz: Done

kemonats:
Change indentation to tabs in the `host-flash.c` file.

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
+	char tempTab[256];
+	size_t wrote = 0, i;
+	int len;
+	ssize_t stat, readed;
+
+	if (addr + bufflen > hostflash_common.flashsz)
+		return -EINVAL;
+
+	while (wrote != bufflen) {
+		if ((bufflen - wrote) / sizeof(tempTab))
```
agkaminski, line 64: I think that `if ((bufflen - wrote) >= sizeof(tempTab))` would be more readable.

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
-        }
-
-        wrote++;
-    }
-
-    return (ssize_t)wrote;
+	char tempTab[256];
+	size_t wrote = 0, i;
+	int len;
+	ssize_t stat, readed;
```
agkaminski, line 58: `readed` is not a valid word in english :) The past tense of read is read

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
-
-    return;
+	char tempTab[256];
+	ssize_t len = sizeof(tempTab);
+	int erased = 0;
+	unsigned int sectorNum;
+	unsigned int sectorAddr;
+	int stat;
+
+	memset(tempTab, 0xFF, sizeof(tempTab));
```
agkaminski, line 103: We use lower case letters to write hex numbers

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
+
+	if (addr < hostflash_common.flashsz) {
+		sectorNum = addr / hostflash_common.sectorsz;
+		sectorAddr = sectorNum * hostflash_common.sectorsz;
+	}
+	else {
+		return;
+	}
```
agkaminski, line 111:
```
if (addr >= hostflash_common.flashsz)
  return;

sectorAddr = (addr / hostflash_common.sectorsz) * hostflash_common.sectorsz;
```

One variable less, 4 lines of code less, same result

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
+
+	while (erased != hostflash_common.sectorsz) {
+		stat = pwrite(hostflash_common.filefd, tempTab, len, sectorAddr + erased);
+		if (stat < 0) {
+			if (errno == EINTR)
+				continue;
+
+			return;
+		}
+		else {
```
agkaminski, line 121: In the `if` before we either `continue` or `return`, no need for `else`

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
+
+	while (erased != hostflash_common.flashsz) {
+		stat = pwrite(hostflash_common.filefd, tempTab, len, erased);
+		if (stat < 0) {
+			if (errno == EINTR)
+				continue;
+
+			return;
+		}
+		else {
```
agkaminski, line 150: In the `if` before we either `continue` or `return`, no need for `else`

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
-
-        erased += stat;
-    }
-
-    return;
+	char tempTab[256];
+	size_t erased = 0, len = sizeof(tempTab);
+	ssize_t stat;
+
+	memset(tempTab, 0xFF, sizeof(tempTab));
```
agkaminski, line 140: We use lower case letters to write hex numbers

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
+	while (erased != hostflash_common.flashsz) {
+		stat = pwrite(hostflash_common.filefd, tempTab, len, erased);
+		if (stat < 0) {
+			if (errno == EINTR)
+				continue;
+
+			return;
+		}
+		else {
+			len = len - stat;
```
agkaminski, line 151: You can type this as `len -= stat;` but this a matter of taste, you can leave it is, if you prefer it.

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,167 @@
...
+	while (erased != hostflash_common.sectorsz) {
+		stat = pwrite(hostflash_common.filefd, tempTab, len, sectorAddr + erased);
+		if (stat < 0) {
+			if (errno == EINTR)
+				continue;
+
+			return;
+		}
+		else {
+			len = len - stat;
```
agkaminski, line 122: You can type this as `len -= stat;` but this a matter of taste, you can leave it is, if you prefer it.

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
Diabliz: english language is so hard sometimes :) Done

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
Diabliz: Done

storage/host-flash/host-flash.c
Diabliz: I prefer `len -= stat;` :) Done

storage/host-flash/host-flash.c
Diabliz: I prefer `len -= stat;` :) Done

agkaminski (approved):
LGTM

storage/host-flash/host-flash.c
```diff
@@ -15,155 +15,161 @@
...
+	ssize_t stat, readsz;
+
+	if (addr + bufflen > hostflash_common.flashsz)
+		return -EINVAL;
+
+	while (wrote != bufflen) {
+		if ((bufflen - wrote) >= sizeof(tempTab))
+			len = sizeof(tempTab);
+		else
+			len = (bufflen - wrote) % sizeof(tempTab);
```
anglov, line 67: Is this modulo needed? 
bufflen - wrote < sizeof(tempTab)  => (bufflen - wrote) % sizeof(tempTab) === bufflen - wrote
Am I right?

anglov (changes requested):
One more little mistake found. Rest looks good.

One side note – we silently return in case of failing of sector erase. Not nice, but I know it's needed for meterfs.

storage/host-flash/host-flash.c
Diabliz: Right, modulo isn't needed.

Diabliz: Done

anglov (approved):

PR #96, opened 2021-04-21
stm32l4_tty: send logs to tty only when UART_CONSOLE points to tty

xvuko:
Currently logs are always sent to TTY when any TTY is enabled. After this change TTY can be used (ie. for optoport) without log redirection.

I've checked stm32l4 projects on gerrit and this would not change current behavior.

### DIFF
--- a/multi/stm32l4-multi/stm32-multi.c
+++ b/multi/stm32l4-multi/stm32-multi.c
@@ -163,7 +163,7 @@ static void handleMsg(msg_t *msg)
 
 static ssize_t console_write(const char *str, size_t len, int mode)
 {
-#if TTY1 || TTY2 || TTY3 || TTY4 || TTY5
+#if CONSOLE_IS_TTY
 	tty_log(str);
 	return (ssize_t)len;
 #else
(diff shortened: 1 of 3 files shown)


multi/stm32l4-multi/stm32-multi.c
```diff
@@ -163,7 +163,7 @@ static void handleMsg(msg_t *msg)
 
 static ssize_t console_write(const char *str, size_t len, int mode)
 {
-#if TTY1 || TTY2 || TTY3 || TTY4 || TTY5
+#if TTY_CONSOLE
```
agkaminski, line 166: Where is `TTY_CONSOLE` defined?

xvuko: It;s not. I missed this during rewrite. This is CONSOLE_IS_TTY now.

agkaminski (approved):
Have you checked the phoenix-rtos-project usage too? (with psh, console on tty). If if works then it LGTM

xvuko:
> Have you checked the phoenix-rtos-project usage too?

Yes, phoenix-rtos-project is not affected.

anglov (approved):
Looks good but I not tested it.

PR #98, opened 2021-04-28
makes: use templates for binaries and static libs compilation

nalajcie:
**DO NOT MERGE**

Various approaches included to be discussed, changes done locally in one repo.

The main source of race condition is that we don't have the component-to-component dependencies (example: `libtty` and `pc-tty`), moreover one component needs to be finished before the compilation of the dependent component starts (so the dependency needs to be defined for object files, not for target binary).

IMHO the simplest and best solution is to create `PHONY` component targets and describe relations between them explicitly (done for `pc-tty` in the commented out code).

While experimenting I encountered multiple recurring patterns, so I'm toying with the idea of templating the most standard use cases (creating static lib, creating binary). That way it's easy to make global changes to the compilation process (as a sample - automatic file-to-file dependencies were reintroduced).

The same idea is applied in:
 - OpenBSD/NetBSD - https://github.com/openbsd/src/blob/master/bin/ls/Makefile
 - Android project - https://developer.android.com/ndk/guides/android_mk#local_export_cflags

(note: checks fail because `riscv`-tty drivers haven't been touched and showcase compilation order fiasco every time :))


binary.mk
```diff
@@ -0,0 +1,33 @@
...
+$(PREFIX_PROG)$(NAME): $(OBJS) $(RESOLVED_LIBS)
+	$(LINK)
+
+# create phony target for dependencies
+.PHONY: $(NAME)
+$(NAME): $(PREFIX_PROG_STRIPPED)$(NAME)
+
+
+# TODO: should we always add targets to all?
+# the other approach would be to include always all Makefiles and add them to "all" on per-TARGET basis
```
anglov, line 31: What I like in all per TARGET basis is that this will looks much simpler (especially with PHONY targets).
But wonder if you would include all modules Makefiles explicitly or want to write wildcard rule (not sure if it's possilble and advisable)? 
Also – don't it increase Makefile processing time?

On the other hand maybe could work better with tab completion?

tty/pc-tty/Makefile
```diff
@@ -3,9 +3,26 @@
...
+
+#OBJS := $(addprefix $(PREFIX_O)tty/pc-tty/, ttypc.o ttypc_vt.o ttypc_vtf.o ttypc_vga.o ttypc_kbd.o ttypc_bioskbd.o)
+
+#$(OBJS) : libtty
+
+#$(PREFIX_PROG)pc-tty: $(OBJS) $(PREFIX_A)libtty.a
+#	$(LINK)
+
+#pc-tty: $(PREFIX_PROG_STRIPPED)pc-tty
+
```
anglov, line 27: probably in this manual way you forgot to add
.PHONY: pc-tty

anglov (commented):
Ok. I suppose that in main Makefile you change rule to always include all components for current TARGET because
make pc-tty
should build libtty so you can't include just one Makefile.

I think both approach (included binary making rules and explicit target rules) will resolve racing issues and I'm perfectly ok with both of them: explicit way give you much more control, but .mk approach is simpler.

anglov:
One more thing
In case of tests you should probably have something like some common source file and one test file. 
In .mk approach you will need to define variable for each test file and then including .mk. 
In explicit approach you could create nice wildcard for this.

Am I wrong?

tty/pc-tty/Makefile
nalajcie: yep, actually there are more things missing, but I wanted to provide minimal working example.

I'll try to rewrite all `phoenix-rtos-devices` and will see what I learn along the way.

By using template (and including all Makefiles) I would like to add other automatic targets, like:
- `$(NAME)-clean` -> removes binaries (or libs), objects and dependency files for single component
- `$(NAME)-install` -> installs binary to predefined rootfs location
- `$(NAME)-uninstall` :question:

nalajcie:
> One more thing
> In case of tests you should probably have something like some common source file and one test file.
> In .mk approach you will need to define variable for each test file and then including .mk.
> In explicit approach you could create nice wildcard for this.
> 
> Am I wrong?

Yes and no :). If there would be a need we can always add other templates (like `multibin.mk`) which will iterate over variables (and internally use `binary.mk`). I tried to find such an example in NetBSD and they use variables with dot notation for that:

https://github.com/NetBSD/src/blob/trunk/tests/lib/libc/sys/Makefile

I don't know if it works in GNU make, but "associative arrays" work, so we could write something like:

```Makefile
NAMES := test1 test2 test3

SRCS := common.c

# or assume that for test X `X.c` always exists
SRCS[test1] := test1.c
SRCS[test2] := test2.c
SRCS[test3] := test3.c

#note: also possible if it would be needed
EXTRA_CFLAGS[test1] := -DDEBUG -g -O0
EXTRA_LDFLAGS[test2] := --wrap=strcmp 

include multibin.mk
```

I'll think about possible solutions when I'll encounter the real-world problem :)

kemonats:
Your suggestion probably does not solve the main problem of missing the `libtty.h` file in include directory when compiling e.g. the `uart16650.c` file.
I soved this problem a few months ago in two ways ( without manually creating dependences ):
1. the Makefile showed the changes:
`Makefile`
```
else

.PHONY: depend
all: depend

depend:

include _targets/Makefile.$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)

endif
```
`tyy/libtty/Makefile`
```
$(PREFIX_H)libtty.h: tty/libtty/libtty.h
	$(HEADER)

all: $(PREFIX_A)libtty.a

depend: $(PREFIX_H)libtty.h
```
This solution does not give 100% certatinty that at the time of compilation of the `*.c` file the needed file `*.h` will be in the include directory. I tested it with `-j` option and there were no errors.

2. The solution than gives 100% certainty is copying needed `*.h` files using `$(SHELL ...)` function. However this requires more changes to the Makefiles.

nalajcie:
Small update - I tried to include all Makefiles but it turns out we have multiple ways to produce the same binary, eg. `flashsrv`, `stm32-multi` - depending on the target. Some are obviously invalid (two separate Makefiles for mostly the same soruce files), but some of them are completely separate - and should be kept that way? (libflashdrv.a ?)


```bash
usb/imxrt-usbc/Makefile:11: warning: overriding recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/lib/libusbclient.a'
usb/imx6ull-usbc/Makefile:11: warning: ignoring old recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/lib/libusbclient.a'
usb/imxrt-usbc/Makefile:14: warning: overriding recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/include/usbclient.h'
usb/imx6ull-usbc/Makefile:14: warning: ignoring old recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/include/usbclient.h'
multi/stm32l1-multi/Makefile:10: warning: overriding recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/prog/stm32-multi'
multi/stm32l4-multi/Makefile:18: warning: ignoring old recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/prog/stm32-multi'
storage/imxrt-flash/Makefile:14: warning: overriding recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/lib/libflashdrv.a'
storage/imx6ull-flash/Makefile:11: warning: ignoring old recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/lib/libflashdrv.a'
storage/imxrt-flash/Makefile:17: warning: overriding recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/include/flashsrv.h'
storage/imx6ull-flash/Makefile:14: warning: ignoring old recipe for target '/home/wm/src/phoenix-rtos-project/_build/ia32-generic/include/flashsrv.h'
```

I don't see a clear and consistent solution for now, I'll need to experiment more :)

nalajcie:
**Changes**:
 - always include all `Makefiles` (make targets tab completion works)
 - explicitly define dependencies between components (fix compilation
order fiasco)
 - provide per-component targets (for compilation and cleaning)
 - hide implementation details from the end-user
 - (re)introduce auto-generated file-to-file dependencies
 - some component names needed to be changed:
   - `libusbclient` - create explicit component with single `Makefile` and optional sources
   - `libflashdrv` (for `imx6ull` and `imxrt`) - actually 2 different libs with different API -> renamed do `libflashdrv-{imx6ull,imxrt}`
   - `stm32-multi` - the same binary for L0 and L4, for now renamed L0 to `stm32l0-multi`, @agkaminski should we rename L4 to `stm32l4-multi`?
 - remove some hacks (empty lib.a to install header, compile objects from outside directories)
 - force `.NOTPARALLEL` when using cleaning (the intent is to almost never use clean because incremental build will work now)
 - provide explicit dependencies for `libphoenix.a`

This PR won't build because `libflashdrv.a` changed name (`psd` will fail).

**Implementation approach**:
 - add new *.mk files to the `phoenix-rtos-build` but keep also "old" approach, not to break everything
 - incrementally move to the new approach across subrepositories

**@anglov , @agkaminski , @HBuczynski -> please check the `Makefiles` written by You if they are readable and provide some tips for possible enhancements.**

Some `Makefiles` vary because I wanted to test different approaches to define sources and other values, I don't know which are the easiest to use - so for now I've kept them all.

I consider the current approach nearly-mergeable, but I want to move the template files to the `phoenix-rtos-build` before the actual merge (thus the file paths are provided by make variables).

Makefile
```diff
@@ -9,38 +9,34 @@
...
-	include _targets/Makefile.$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)
+ifneq ($(filter %clean,$(MAKECMDGOALS)),)
+.NOTPARALLEL:
 endif
+
+# should define DEFAULT_COMPONENTS and target-specific variables
+include _targets/Makefile.$(TARGET_FAMILY)-$(TARGET_SUBFAMILY)
+
+# read out all components
+ALL_MAKES := $(wildcard */*/Makefile) $(wildcard */*/*/Makefile)
```
kemonats, line 35: Wouldn't it be better to use the `find` command :
```
ALL_MAKES := $(filter-out ./Makefile,$(shell find -name Makefile))
```

anglov (commented):
I really like your changes. It seems to be a big step to clean up build system, make it robust  and simplify makefiles greatly.

What is really needed for projects is passing CFLAGS to specific component instead of whole packages in the repository – for example for providing some additional defines basing on env.

About a .NOTPARALLEL – did I understand it correctly:
When you use `make clean all you` will compile everything like with -j1 but when do `make clean && make all` you do everything like before but with pararellisation?
It's ok for me, but as we use clean all strongly It's important to properly point that.

_targets/Makefile.armv7m3-stm32l152xd
```diff
@@ -8,4 +8,4 @@
 
 CFLAGS += -DTARGET_STM32 -DTARGET_STM32L1
 
-include multi/stm32l1-multi/Makefile
+DEFAULT_COMPONENTS := stm32l1-multi
```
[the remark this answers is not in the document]
nalajcie: ok, will do

Makefile
nalajcie: Yeah, builtin should be quicker.

IMHO this is a minor issue, I went wit the easiest approach and will see if it would work in other repos.

I don't know if including "child" Makefiles by other Makefiles is the way to go - if somebody would reintroduce `*/*/*/Makefile` here (or go with `find` approach) we will be including one makefile multiple times.

multi/stm32l4-multi/Makefile
```diff
@@ -4,23 +4,18 @@
...
+DEPS := libtty
 
-$(MULTIDRV_OBJS): $(SRCHEADERS) $(PREFIX_H)libtty.h
+include $(static-lib.mk)
 
-$(PREFIX_PROG)stm32-multi: $(MULTIDRV_OBJS) $(PREFIX_A)libstm32l4-multi.a $(PREFIX_A)libtty.a
-	$(LINK)
+NAME := stm32-multi
+SRCS := $(wildcard $(LOCAL_PATH)*.c)
+LOCAL_HEADERS := stm32-multi.h
```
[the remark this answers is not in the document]
nalajcie: yeah, actually `LOCAL_LIBS` was introduced at the beginning, I later added `LOCAL_SRCS/HEADERS` when I started rewriting Makefiles to get rid of duplication.

Hmm... maybe `DEP_LIBS` would be better? (still `LIBS`, but from which we're directly dependent?)

static-lib.mk
```diff
@@ -0,0 +1,70 @@
...
+# - LOCAL_HEADERS - headers to be installed (relative to current makefile)
+# - HEADERS       - headers to be installed (relative to project root)
+#
+# - DEPS - list of components from current repo to be completed before starting this one
+#
+
+### FIXME: should be available in phoenix-rtos-build for usage in all projects
+
+# directory with current Makefile - relative to the repository root
+CLIENT_MAKES := $(filter-out $(binary.mk) $(static-lib.mk) %.c.d,$(MAKEFILE_LIST))
```
nalajcie, line 15: `static-lib.mk` and `binary.mk` share much common code, I'm thinking about introducing some common files for them, but that's just implementation detail

nalajcie (commented):
Some more notes:
- I stalled the per-target `CFLAGS` experiment as I thought I need to introduce changes in `phoenix-rtos-build`. I will try to do it here and if it would fail, move the template files to `phoenix-rtos-build`
- regarding the `clean` -> I intend to remove it from the targets passed from `phoenix-rtos-build/build.sh` (passing `clean` to the build script already removes the whole build dir, giving it as a target to makes is redundant). The main intended use case is to rebuild one component if the necessary change would not be triggered by auto-deps (eg. changing ENV variable or `build.projec`, etc.)

nalajcie:
Small update regarding custom per-target `CFLAGS/LDFLAGS`, it can easily be added by hand, eg by writing:

```Makefile
$(NAME): CFLAGS:=$(CFLAGS) -DSPECIAL_DEFINE
```
The downside is that the custom value is kept also for all dependent targets (unless somebody also would redefine it) - so also dependent libraries :).

I'll add `LOCAL_CFLAGS` / `LOCAL_LDFLAGS` for ease of use and restrict the variable only to the sane targets:

```Makefile
$(OBJS.$(NAME)): CFLAGS:=$(CFLAGS) $(LOCAL_CFLAGS)
$(PREFIX_PROG)$(NAME): LDFLAGS:=$(LDFLAGS) $(LOCAL_LDFLAGS)
```

nalajcie:
Changes:
- rebased on top of current master
- added `LOCAL_CFLAGS`/`LOCAL_LDFLAGS`
- renamed `LOCAL_LIBS` to `DEP_LIBS`
- (separate commit) mark external `Makefile`'s by renaming to `Makefile.ext` (cm4 bare-metal component)

nalajcie:
Waiting for the checks to finish and merging this.

I will also propagate it to some of the other repos.

Please note that we will have separate `.clang-format` in each of the repos (to be able to check the PRs without the headache of checking-out full `phoenix-rtos-project`), so any coding style changes will need to be propagated to other repos also.

PR #101, opened 2021-05-07
[DO NOT MERGE] Feature/clang format - testing PR use case

nalajcie:
**DO NOT MERGE**


nalajcie:
Showcase PR/direct commit with `clang-format` auto-checks (with suggested changes). These commits were checked twice because they are direct-push'es to "feature/clang-format" and (now) pull request content.

For now added some basic `.clang-format` file, please verify if it aligns well with current coding standard. The best way to check is to:
```bash
clang-format -i $(find . -name \*.c -o -name \*.h)
git diff # or git-cola
```

Parts of current code will need to be wrapped in:
```c
// clang-format off
enum {a, s,d,        f};   //////// I don't care about coding style!
// clang-format on
```

agkaminski:
I love it!

nalajcie:
Waiting for the checks to finish and merging this.

I will also propagate it to some of the other repos.

Please note that we will have separate `.clang-format` in each of the repos (to be able to check the PRs without the headache of checking-out full `phoenix-rtos-project`), so any coding style changes will need to be propagated to other repos also.

anglov:
> I found it kind of inconsistent in current code with slight favour of indenting with one extra tab. It should be easy to change by setting `ContinuationIndentWidth` to 16 (two tab widths). The continuation should be done now by tabs, It can be changed to spaces if You would prefer to
> 
> >

It should be tabs.


> sorry, I don't really see the difference in the preview (apart from the indentation :))
> 
> >

It's indentation issue :)



> This is not settable only per-enum, we would need to align all `=`, so that might be bad. Also - because of the "inlined enums" style, it will be pretty common to disable `clang-format` for enums so I consider this a non-issue :)

OK, maybe it should be then:
status_abcdefABCDEFGHIJ =          1U << 0,


> Please experiment with the options, I would gladly welcome PR which would reformat most of the existing code :).
> 
> >

Thanks for configuration options ideas.

PR #105, opened 2021-05-18
ade7913: add API and test for ADE7913 device

Cinek28:
JIRA: NIL-117

### DIFF
--- a/adc/ade7913/ade7913-test.c
+++ b/adc/ade7913/ade7913-test.c
@@ -1,0 +1,82 @@
+/*
+ * Phoenix-RTOS
+ *
+ * i.MX RT1170 ADE7913 test application
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Marcin Baran
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include <sys/msg.h>
+
+#include "ade7913.h"
+
+
+int main(int argc, char **argv)
+{
+	int i;
+	oid_t ade7913_spi;
+	ade7913_burst_reg_t sample;
+
+	int devcnt = 1;
+
+	if (argc >= 2)
+		devcnt = atoi(argv[1]);
+
+	printf("Starting ADE7913 test (%d devices)\n", devcnt);
+
+	if (devcnt > 4 || devcnt <= 0) {
+		printf("Incorrect ADE7913 device count (4 max)\n");
+		return -1;
+	}
+
+	while (lookup("/dev/spi1", NULL, &ade7913_spi) < 0)
+		usleep(5000);
+
+	printf("SPI1 initialized\n");
+
+	for (i = 0; i < devcnt; ++i) {
+		printf("Configuring ADE7913 device nr %d\n", i);
+
+		while (ade7913_init(&ade7913_spi, i, i < devcnt - 1 ? 1 : 0) < 0) {
+			printf("Failed to initialize ade7913 nr %d\n", i);
+			usleep(500000);
+		}
+
+		if (ade7913_lock(&ade7913_spi, i) < 0)
+			printf("Could not lock ADE7913 nr %d\n", i);
+		if (ade7913_enable(&ade7913_spi, i) < 0)
+			printf("Could not enable ADE7913 nr %d\n", i);
+	}
+
+	printf("Reading ADE7913 registers in burst mode:\n");
+
+	while (1) {
+		for (i = 0; i < devcnt; ++i) {
+			if (ade7913_sample_regs_read(&ade7913_spi, i, &sample) < 0) {
+				printf("Failed reading sample registers from device %d\n", i);
+				continue;
+			}
+
+			printf("IWV[%d]: %x\n", i, sample.iwv);
+			printf("V1WV[%d]: %x\n", i, sample.v1wv);
+			printf("V2WV[%d]: %x\n", i, sample.v2wv);
+			printf("ADC_CRC[%d]: %x\n", i, sample.adc_crc);
+			printf("CNT_SNAPSHOT[%d]: %x\n", i, sample.cnt_snapshot);
+			printf("STATUS0[%d]: %x\n", i, sample.status0);
+		}
+
+		printf("\n");
+		usleep(500000);
+	}
+
+	return 0;
+}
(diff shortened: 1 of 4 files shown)


adc/ade7913/ade7913.c
```diff
@@ -0,0 +1,304 @@
...
+	msgSend(device->port, &msg);
+
+	return omsg->err;
+}
+
+
+static int ade7913_read_reg(oid_t *device, int cs,
+	uint8_t addr, uint8_t *data, size_t size)
+{
+	uint8_t buff[size + 1];
```
kemonats, line 104: VLA are allowed?

adc/ade7913/ade7913-test.c
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		for (i = 0; i < devcnt; ++i) {
+			printf("IWV[%d]: %x\n", i, sample.iwv);
+			printf("V1WV[%d]: %x\n", i, sample.v1wv);
+			printf("V2WV[%d]: %x\n", i, sample.v2wv);
+			printf("ADC_CRC[%d]: %x\n", i, sample.adc_crc);
+			printf("CNT_SNAPSHOT[%d]: %x\n", i, sample.cnt_snapshot);
+			printf("STATUS0[%d]: %x\n", i, sample.status0);
+		}
```
kemonats, line 75: From which device are the values printed.
What is printed if the function `ade7913_sample_regs_read()` returns an error.

adc/ade7913/ade7913.c
Cinek28: VLA is on, not sure it is best approach. @agkaminski , @nalajcie what do you guys think?

adc/ade7913/ade7913-test.c
Cinek28: Fixed, values are now printed for each device. If ade7913_sample_regs_read() returns error, error message is printed and values are ommited.

adc/ade7913/ade7913.c
agkaminski: Based on my knowledge VLA are C99 extension. We try to adhere to ANSI C (C89) standard as much as possible, because of possibility that we'll need to compile our code for some obscure target in the future.

It would be best to allocate this buffer by some other means.

adc/ade7913/ade7913.c
```diff
@@ -0,0 +1,304 @@
...
+{
+	uint8_t reg;
+
+	if (ade7913_read_reg(device, cs, ADE7913_STATUS0, &reg, sizeof(reg)))
+		return -EIO;
+
+	if (!(reg & ADE7913_STATUS0_RESET_ON))
+		return ADE7913_READY;
+
+	return !ADE7913_READY;
```
agkaminski, line 180: This define `ADE7913_READY` is really no needed. It's only used here to return either zero or one.

adc/ade7913/ade7913.c
```diff
@@ -0,0 +1,304 @@
...
+static int ade7913_write_reg(oid_t *device, int cs,
+	uint8_t addr, uint8_t value)
+{
+	int res = 0;
+	uint8_t buff[2] = { (addr << ADE7913_ADDR_OFFS), value };
+
+	/* Check if register is writtable */
+	if ((addr >= ADE7913_IWV && addr <= ADE7913_CNT_SNAPSHOT) ||
+		addr > ADE7913_EMI_CTRL || addr == ADE7913_STATUS0)
+		return -EINVAL;
```
agkaminski, line 133: Line-broken `if` condition should be indented by two levels

agkaminski:
```suggestion
	if ((addr >= ADE7913_IWV && addr <= ADE7913_CNT_SNAPSHOT) ||
			addr > ADE7913_EMI_CTRL || addr == ADE7913_STATUS0)
		return -EINVAL;
```

adc/ade7913/ade7913.c
```diff
@@ -0,0 +1,304 @@
...
+#define ADE7913_CNT_SNAPSHOT 0x7
+#define ADE7913_CONFIG       0x8
+#define ADE7913_STATUS0      0x9
+#define ADE7913_LOCK         0xa
+#define ADE7913_SYNC_SNAP    0xb
+#define ADE7913_COUNTER0     0xc
+#define ADE7913_COUNTER1     0xd
+#define ADE7913_EMI_CTRL     0xd
+#define ADE7913_STATUS1      0xf
+#define ADE7913_TEMPOS       0x18
```
agkaminski, line 43: By convention, when creating device drivers, we define registers offsets using enum

agkaminski:
```suggestion
enum { ade7913_iwv = 0, ade7913_v1wv, ade7913_v2wv, ade7913_adc_crc = 4, ade7913_ctrl_crc,
ade7913_cnt_snapshot = 7, ade7913_config, ade7913_status0, ade7913_lock, ade7913_sync_snap,
ade7913_counter0, ade7913_counter1, ade7913_emi_ctrl, ade7913_status1, ade7913_tempos = 24 };
```

adc/ade7913/ade7913.c
```diff
@@ -0,0 +1,304 @@
...
+
+	for (i = 0; i < 4; ++i) {
+		if (ade7913_ready(device, cs))
+			break;
+
+		usleep(10000);
+	}
+
+	if (i == 4)
+		return -EBUSY;
```
agkaminski, line 196: my suggestion

agkaminski:
```suggestion
	i = 0
	while (!ade7913_ready(device, cs)) {
		if (++i > 4)
			return -EBUSY;
		
		usleep(10000);
	}
```

agkaminski:
@Cinek28  I know that this code is WIP, so if you would like to fix above remarks in the future PR, we can merge this one now. Just let me know.

adc/ade7913/ade7913.c
Cinek28: Removed define

adc/ade7913/ade7913.c
Cinek28: Changed to enum

adc/ade7913/ade7913.c
Cinek28: Changed according to suggestion

Cinek28:
All comments aside from VLA usage fixed. I will change buffer allocation along with the changes I am working on right now.

PR #107, opened 2021-05-24
imxrt-flash: fixed sector erase error.

Diabliz:


HBuczynski:
I don't think so that it is a good approach. 

For example, user will invoke this function with offset located in the middle of the sector and not the part of the sector will be erased but the whole one. It should be handle by the higher level.
If it is a temporary solution, please add some comment.

agkaminski:
If someone uses this function, most likely they know how NOR flash works. Best approach would be to pass just the sector number (i.e. MSB bits of address) instead, because LSBs doesn't really matter in this operation.

If we treat LSBs of address as "doesn't matter", approach of this PR is perfectly ok

HBuczynski:
> If someone uses this function, most likely they know how NOR flash works. Best approach would be to pass just the sector number (i.e. MSB bits of address) instead, because LSBs doesn't really matter in this operation.
> 
> If we treat LSBs of address as "doesn't matter", approach of this PR is perfectly ok

Ok, it depends on the approach. When i was writing `flashsrv & flashdrv`, I was taking example from `imx6ull-flashsrv & flashdrv` where such arguments to functions are checked.

PR #111, opened 2021-05-25
ade7913: add CS pins ordering as arg, fix power enable

Cinek28:
ADE7913 devices can be connected to SPI CS pins in many different
orders, thus giving CS pin orders as argument like:
ade7913-test;2301 - means main ADE7913 device (source of the clock)
is connected to CS pin 2, next in chain is connected to CS
pin 3, etc. This argument ensures correct initialization.

JIRA: NIL-117

### DIFF
--- a/adc/ade7913/ade7913-test.c
+++ b/adc/ade7913/ade7913-test.c
@@ -14,26 +14,65 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <unistd.h>
+#include <string.h>
 
 #include <sys/msg.h>
+#include <imxrt-multi.h>
 
 #include "ade7913.h"
 
 
+#define ADE7913_DEFAULT_LIST "1230"
+
+
+static int lpspi_config(oid_t *device, int spi)
+{
+	msg_t msg;
+	multi_i_t *imsg = (multi_i_t *)msg.i.raw;
+
+	msg.type = mtDevCtl;
+	msg.i.data = NULL;
+	msg.i.size = 0;
+	msg.o.data = NULL;
+	msg.o.size = 0;
+
+	imsg->id = id_spi1 + spi;
+	imsg->spi.type = spi_config;
+	imsg->spi.config.cs = 0;
+	imsg->spi.config.mode = spi_mode_0;
+	imsg->spi.config.endian = spi_msb;
+	imsg->spi.config.sckDiv = 4;
+	imsg->spi.config.prescaler = 3;
+
+	return msgSend(device->port, &msg);
+}
+
+
 int main(int argc, char **argv)
 {
-	int i;
+	int i, devnum, devcnt;
+	const char *order;
 	oid_t ade7913_spi;
 	ade7913_burst_reg_t sample;
 
-	int devcnt = 1;
+	if (argc != 2) {
+		printf("No device list given. Using default: %s\n", ADE7913_DEFAULT_LIST);
+		devcnt = strlen(ADE7913_DEFAULT_LIST);
+		order = ADE7913_DEFAULT_LIST;
+	}
+	else {
+		order = argv[1];
+		devcnt = strlen(argv[1]);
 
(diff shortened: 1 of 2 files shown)


adc/ade7913/ade7913-test.c
```diff
@@ -14,64 +14,108 @@
...
-
-	if (argc >= 2)
-		devcnt = atoi(argv[1]);
-
-	printf("Starting ADE7913 test (%d devices)\n", devcnt);
+	if (argc != 2) {
+		printf("No device list given. Using default: %s\n", ADE7913_DEFAULT_LIST);
+		devcnt = strlen(ADE7913_DEFAULT_LIST);
+		order = ADE7913_DEFAULT_LIST;
+	} else {
```
agkaminski, line 62: clang-format-pr

adc/ade7913/ade7913-test.c
```diff
@@ -14,64 +14,108 @@
...
 	for (i = 0; i < devcnt; ++i) {
-		printf("Configuring ADE7913 device nr %d\n", i);
+		devnum = (int)(order[i] - '0');
+
+		usleep(500000);
+		printf("Configuring ADE7913 device nr %d\n", devnum);
 
-		while (ade7913_init(&ade7913_spi, i, i < devcnt - 1 ? 1 : 0) < 0) {
-			printf("Failed to initialize ade7913 nr %d\n", i);
+		while ( ade7913_init(&ade7913_spi, devnum,
```
agkaminski, line 90: clang-format-pr

adc/ade7913/ade7913-test.c
```diff
@@ -14,64 +14,108 @@
...
+		printf("Could not initialize SPI1\n");
+		return -1;
+	}
+
 	printf("SPI1 initialized\n");
 
+	/* Start init from device with xtal */
 	for (i = 0; i < devcnt; ++i) {
-		printf("Configuring ADE7913 device nr %d\n", i);
+		devnum = (int)(order[i] - '0');
```
agkaminski, line 85: You should check if you got` 0 >= devnum > devcnt`. If I were to pass "abcd" as dev list, this wouldn't end well

agkaminski (commented):
I like the idea with access order list being passed via digits string, quite neat. But you should sanitaze this input.

Two formatting errors

adc/ade7913/ade7913-test.c
```diff
@@ -14,64 +14,115 @@
...
+
+		usleep(500000);
+		printf("Configuring ADE7913 device nr %d\n", devnum);
 
-		while (ade7913_init(&ade7913_spi, i, i < devcnt - 1 ? 1 : 0) < 0) {
-			printf("Failed to initialize ade7913 nr %d\n", i);
+		while ( ade7913_init(&ade7913_spi, devnum,
+				   devnum == order[devcnt - 1] - '0' ? 0 : 1) < 0) {
```
agkaminski, line 98: Still formatting error

agkaminski:
```suggestion
		while (ade7913_init(&ade7913_spi, devnum,
				devnum == order[devcnt - 1] - '0' ? 0 : 1) < 0) {
```

adc/ade7913/ade7913-test.c
Cinek28: Fixed

adc/ade7913/ade7913-test.c
Cinek28: Fixed

adc/ade7913/ade7913-test.c
Cinek28: Added check for order (from '0' to '3').

adc/ade7913/ade7913-test.c
Cinek28: Now should be ok

PR #113, opened 2021-05-26
stm32-multi: Fix invalid header installation

anglov:
There was a conflict between rules for stm32l1 and stm32l4 header
version of stm32-multi.h. Changes unify name usage to avoid problematic
stm32-multi repeat for different components.

JIRA: ISC-105


anglov:
Some more background:
When I build project, I found that stm32l1 header was installled instead of stm32l4. It's because duplication of header installation rule probably. 

To avoid that I renamed both headers. To not confuse further I renamed also binary of stm32-mult for l4 to stm32l4-multi. Before it was stm32l1-multi for l1 and stm32-multi for l4. It was not needed, but I think it will be better and less confusing that way,

nalajcie:
I like the change, I didn't do it myself because I was afraid of disrupting some internal projects.

Needs additional changes in `phoenix-rtos-project`, otherwise LGTM.

anglov:
> I like the change, I didn't do it myself because I was afraid of disrupting some internal projects.
> 
> Needs additional changes in `phoenix-rtos-project`, otherwise LGTM.

Need to add something more than 
https://github.com/phoenix-rtos/phoenix-rtos-project/pull/88
(i cannot add reviewers there)

nalajcie:
> Need to add something more than
> [phoenix-rtos/phoenix-rtos-project#88](https://github.com/phoenix-rtos/phoenix-rtos-project/pull/88)
> (i cannot add reviewers there)


Should be enough, but commit hash might change after merging this PR.

I think we can merge this now and update the other PR if needed.

nalajcie (approved):

PR #126, opened 2021-08-04
imx6ull-flashsrv: fix flashsrv_write() overwriting metadata

Maxez:
## Description

Passing `aux == NULL` to `flashdrv_write()` means that we're going to perform partial page programming (data only). Therefore not programmed area (metadata and its ECC) should be set to `0xff`.
We trick BCH controller into thinking that this area contains metdata only without ECC (so that BCH won't calculate new ECC for the metadata).

## Motivation and Context

[JIRA: DTR-109](https://jira.phoenix-rtos.com/browse/DTR-109)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a7-imx6ull + extended `test_meta()` in `imx6ull-flash-test.c`

## Checklist:

## Special treatment

### DIFF
--- a/storage/imx6ull-flash/flashdrv.c
+++ b/storage/imx6ull-flash/flashdrv.c
@@ -693,13 +693,19 @@ int flashdrv_write(flashdrv_dma_t *dma, uint32_t paddr, void *data, char *aux)
 {
 	int chip = 0, channel = 0, sz;
 	char addr[5] = { 0 };
-	int err;
+	int skipMeta = 0, err;
 	memcpy(addr + 2, &paddr, 3);
 
-	if (data != NULL)
-		sz = flashdrv_common.info.writesz + flashdrv_common.info.metasz;
-	else
+	if (data == NULL) {
 		sz = flashdrv_common.rawmetasz;
+	}
+	else {
+		sz = flashdrv_common.info.writesz + flashdrv_common.info.metasz;
+		if (aux == NULL) {
+			aux = (char *)flashdrv_common.uncached_buf;
+			skipMeta = 1;
+		}
+	}
 
 	dma->first = NULL;
 	dma->last = NULL;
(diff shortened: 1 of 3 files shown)


Maxez:
Changed the fix so that we no longer need to read the metadata to correctly program the page.

storage/imx6ull-flash/flashdrv.c
```diff
@@ -720,6 +724,14 @@ int flashdrv_write(flashdrv_dma_t *dma, uint32_t paddr, void *data, char *aux)
 		*(flashdrv_common.bch + bch_flash0layout1) &= ~(0xffff << 16);
 		*(flashdrv_common.bch + bch_flash0layout1) |= flashdrv_common.rawmetasz << 16;
 	}
+	else if (aux == (char *)flashdrv_common.uncached_buf) {
```
nalajcie, line 727: I would prefer to add explicit variable (eg. `skipMeta` or something similar) instead of condition based on pointer value

storage/imx6ull-flash/imx6ull-flash-test.c
```diff
@@ -511,11 +511,10 @@ void test_meta(void)
 	if ((err = flashdrv_erase(dma, paddr) < 0))
 		printf("erase() failed: %d\n", err);
 
-	/* test writing page data + meta */
+	/* test writing meta + data */
 	memset(data, 0xff, pagemapsz);
-	for (i = 0; i < sizeof(aux->metadata); ++i) {
+	for (i = 0; i < sizeof(aux->metadata); ++i)
```
nalajcie, line 516: not relevant change, also: against MISRA (probably not yet relevant but in the near future will be - 
Rule 15.6 - The body of an iteration-statement or a selection-statement shall be a compound-statement).

Please revert all compound-statements removal

nalajcie (commented):
seems good, I think readability could be improved by introducing additional local variable.

storage/imx6ull-flash/flashdrv.c
Maxez: Changed as suggested - I've added `skipMeta` variable.

storage/imx6ull-flash/imx6ull-flash-test.c
Maxez: Thanks for pointing this out. I wasn't aware of this rule. I've restored iteration-statement braces but left out original selection-statements (`if` without braces). Should I add braces to the one-line `if` statements too?

nalajcie (approved):

PR #132, opened 2021-08-30
imxrt1170: quickfix of SPI and eDMA clock performance

gerard5:
This change is intended to make the `imxrt1176` SPI driver behave similar to SPI driver on `imxrt1064` target (clock setup).

Additionally this PR fixes poor eDMA performance on `imxrt1176`.

JIRA: NIL-139, NIL-140

## Description

The imxrt-multi implementation on imxrt117x resets the default SPI clocks settings while turning on, this is different from the imxrt1064 implementation, where the clocks when turned on do not affect the rest of the configuration, i.e. dividers, multipliers, PLL's, etc.

A similar bug has also occurres in eDMA library and this PR fixes it.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (nil-imxrt1176).

## Checklist:

## Special treatment

### DIFF
--- a/multi/imxrt-multi/imxrt1170.h
+++ b/multi/imxrt-multi/imxrt1170.h
@@ -308,16 +308,32 @@
 
 static inline int common_setClock(int clock, int div, int mux, int mfd, int mfn, int state)
 {
+	int res;
 	platformctl_t pctl;
 
-	pctl.action = pctl_set;
+	pctl.action = pctl_get;
 	pctl.type = pctl_devclock;
 	pctl.devclock.dev = clock;
-	pctl.devclock.div = div;
-	pctl.devclock.mfd = mfd;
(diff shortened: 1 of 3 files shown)


multi/imxrt-multi/imxrt1170.h
```diff
@@ -308,16 +308,32 @@
 
 static inline int common_setClock(int clock, int div, int mux, int mfd, int mfn, int state)
```
gerard5, line 309: Maybe the function should not be inlined ? I don't know why this tendency is to `static inline` everything here, especially since this function could be reused by calling rather than by inlining.

nalajcie (dismissed):
not too much into imxrt117x but this is a bugfix so LGTM.

nalajcie (approved):

PR #135, opened 2021-09-15
imx6ull-gpio: add text write mode

nalajcie:
## Description

We need easy way to toggle leds from shell scripts. This is a kind-of hackish approach (single file which understands `text` and `binary` write mode), but I can use other approach if preferable:
 - separate files for `text mode` (`textdir`, `textport`? :rofl:)
 - `export` file which would allow us to dynamically export specific pin from specific port as a dir with files for different options (the linux way)

For now went whit the easiest path without creating additional files at all.

## Functional description

The `dir` and `port` files support text write mode for easier usage from shell scripts.

The syntax is `[+/-][pin]`, eg `+28`, `-7`, where:
 - `+` means _value high_ for `port` and _output_ for `dir`
 - `-` means _value low_ for `port` and _input_ for `dir`
 - `pin` is the numeric pin value in <0, 31> range

This interface allows changing only one pin value / direction at a time (state of other pins won't be changed).

### Example

```bash
# set direction to output
echo -n "+9" > /dev/gpio2/dir
# set output value to LOW
echo -n "-9" > /dev/gpio2/port
```
## Motivation and Context
JIRA: DTR-11, DTR-47

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/gpio/imx6ull-gpio/imx6ull-gpio.c
+++ b/gpio/imx6ull-gpio/imx6ull-gpio.c
@@ -38,8 +38,6 @@ enum { dr = 0, gdir, psr, icr1, icr2, imr, isr, edge };
 struct {
 	struct {
 		volatile uint32_t *base;
-		id_t port;
-		id_t dir;
 		handle_t lock;
 	} gpio[5];
 
@@ -54,7 +52,7 @@ static const int clocks[] = { pctl_clk_gpio1, pctl_clk_gpio2, pctl_clk_gpio3, pc
 int init(void)
 {
 	int i, err;
-	char devpath[11];
+	char devpath[sizeof("gpioX/port")];
 	platformctl_t pctl;
 	oid_t dev;
 
@@ -72,7 +70,7 @@ int init(void)
 
 	for (i = 0; i < sizeof(common.gpio) / sizeof(common.gpio[0]); ++i) {
 
-		sprintf(devpath, "gpio%d/port", i + 1);
+		snprintf(devpath, sizeof(devpath), "gpio%u/port", (i + 1) % 10);
 
 		dev.port = common.port;
 		dev.id = gpio1 + i;
@@ -82,9 +80,7 @@ int init(void)
 			return - 1;
 		}
 
-		common.gpio[i].port = dev.id;
-
-		sprintf(devpath, "gpio%d/dir", i + 1);
+		snprintf(devpath, sizeof(devpath), "gpio%u/dir", (i + 1) % 10);
 
 		dev.port = common.port;
 		dev.id = dir1 + i;
@@ -93,8 +89,6 @@ int init(void)
 			printf("gpiodrv: Could not create direction file #%d (err %d)\n", i + 1, err);
 			return - 1;
 		}
-
-		common.gpio[i].dir = dev.id;
 
 		pctl.action = pctl_set;
 		pctl.type = pctl_devclock;
@@ -196,6 +190,7 @@ void thread(void *arg)
 			case mtRead:
 				if (msg.o.data != NULL && msg.o.size >= sizeof(uint32_t)) {
 					d = msg.i.io.oid.id;
+					/* NOTE: ignoring msg.i.io.offs to allow repetetive reads without seek / reopen */
 
 					if (d >= gpio1 && d <= gpio5) {
 						msg.o.io.err = gpioread(d, &val);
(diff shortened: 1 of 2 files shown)


gpio/imx6ull-gpio/imx6ull-gpio.c
```diff
@@ -54,7 +53,7 @@ static const int clocks[] = { pctl_clk_gpio1, pctl_clk_gpio2, pctl_clk_gpio3, pc
 int init(void)
 {
 	int i, err;
-	char devpath[11];
+	char devpath[sizeof("gpioX/port") + 1];
```
kemonats, line 56: Why `+1`?
`sizeof("gpioX/port") == 11` also includes `null` character.

gpio/imx6ull-gpio/imx6ull-gpio.c
```diff
@@ -212,13 +208,35 @@ void thread(void *arg)
...
+
+						if (pin > 31)
+							break; /* -EINVAL */
 
-					if (msg.i.size >= sizeof(uint32_t) << 1)
-						memcpy(&mask, msg.i.data + sizeof(uint32_t), sizeof(uint32_t));
-					else
-						mask = (uint32_t)-1;
+						mask = 1 << pin;
+						val = ((data[0] == '+') ? 1 : 0) << pin;
```
kemonats, line 227:
```suggestion
						mask = 1 << pin;
						val = (data[0] == '+') ? mask : 0;
```
Litle simpler.

gpio/imx6ull-gpio/imx6ull-gpio.c
nalajcie: Thanks.
^C^V from other driver, fixed.

gpio/imx6ull-gpio/imx6ull-gpio.c
nalajcie: I prefer my version as it's more semantic (the intent is clearer). With optimizations enabled gcc would produce almost the same code for both versions.

gpio/imx6ull-gpio/imx6ull-gpio.c
```diff
@@ -212,13 +208,35 @@ void thread(void *arg)
...
-					memcpy(&val, msg.i.data, sizeof(uint32_t));
+				if (msg.i.data != NULL && msg.i.size >= 2) {
+					uint8_t *data = (uint8_t *)msg.i.data;
+
+					/* NOTE: ignoring msg.i.io.offs to allow repetetive writes without seek / reopen */
+
+					if ((msg.i.size % sizeof(uint32_t) != 0) && ((data[0] == '+') || (data[0] == '-'))) {
+						printf("GPIO: text mode!\n");
+						/* text mode [+/-][pin] -> 1/0 as out/in or value */
+						uint32_t pin = 0;
```
anglov, line 219: variable declaration not at the beginning of the scope

anglov (commented):
My concern on this approach is about force to text mode be divided by sizeof(uint32_t). It can be potentially error prone.

Let's imagine that someone forget to use echo without -n
Then in case of 1 digit gpio we will have message of size 3 (ok, text mode)
But in case of 2 digit gpio we will have message of size 4 (not ok, binary mode).

It could be source of weird errors. Maybe discard ill-formated (with wrong characters as \n) messages at all would minimalize issue?

In my opinion it's ok as a quick addition, but handling text mode by device drivers should be discussed. 

Don't want to approve, could be great if @agkaminski would also check.

gpio/imx6ull-gpio/imx6ull-gpio.c
nalajcie: shit, forgot to remove debug printf, will fix once @agkaminski would review.

nalajcie:
> My concern on this approach is about force to text mode be divided by sizeof(uint32_t). It can be potentially error prone.

Yep,
actually the `README` says binary mode accepts only len == 8, but from code I see len == 4 is also accepted (and mask is assumed to be `0xffffffff` in this case) - which is also error-prone (You forget to add the mask or split it into 2 writes and You change state of all other GPIOs).

Maybe we should remove the `len == 4` binary case?

As I said previously - I'm not against introducing new separate files for text mode if You feel the interface would be more straightforward.

Btw, the driver uses mutexes to guard against concurrent access to the GPIO registers... but it's single-threaded for now, so they can be safely deleted :).

agkaminski:
I made this interface. I was _forced_ to use read/write, but I _wanted_ to have binary interface.

My proposition - move binary interface to devctl (eg. we might write to any of the gpio files and specify port we want to write to/read from) and use read/write for human readable interface (eg. simple write/read of decimal number from 0 to 2^32 - 1)

nalajcie:
I propose to go with the current version of this PR for now and switch to DevCtl binary interface when necessary improvements in `msg_t` would be done (introducing generic `oid_t` and generic `err` fields). Specifying new devCtl interface right now seems like a waste of time IMHO.

agkaminski (approved):

PR #139, opened 2021-09-22
`i2c` common API: add i2c imx6ull BUS driver and some simple i2c slave device drivers

nalajcie:
## Description

Add `i2c` master common API, add `imx6ull` implementation, and implement drivers for 2 i2c devices. Current setup supports using `i2c` bus either as a library (embedded directly int i2c device driver) or as a separate program with message communication (needed if more than one slave device is placed on the same i2c bus). Both approaches were tested on `imx6ull` target with success.

Commit summary:
 - i2c: add generic interface and message implementation …
 - i2c: add imx6ull bus driver
 - rtc: add pcf85363a i2c rtc support
 - sensor: add nct75 i2c temperature sensor support …

## Motivation and Context
Needed for DTR project, relevant JIRA tickets in respective commit messages.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/i2c/common/libi2c-msg.c
+++ b/i2c/common/libi2c-msg.c
@@ -1,0 +1,130 @@
+/*
+ * Phoenix-RTOS
+ *
+ * i2c msg client library
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Marek Bialowas
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <sys/msg.h>
+#include <sys/types.h>
+
+#include <stdio.h>
+#include <errno.h>
+#include <unistd.h>
+
+#include <i2c.h>
+#include <i2c-msg.h>
+
+static struct {
+	oid_t i2c_oid;
+	int initialized;
+} common;
+
+/* opens the device */
+int i2c_init(unsigned int dev_no)
+{
+	char devname[] = "/dev/i2cX";
+	int timeout = 10; /* x 10 ms = 100 ms */
+
+	devname[sizeof(devname) - 2] = '0' + (dev_no % 10);
+	while (lookup(devname, NULL, &common.i2c_oid) < 0) {
+		if (timeout-- <= 0)
+			return -ETIMEDOUT;
+
+		usleep(10 * 1000);
+	}
+
+	common.initialized = 1;
+	return 0;
+}
+
+
+/* Performs i2c generic write operation to the given slave device. */
+int i2c_busWrite(uint8_t dev_addr, const uint8_t *data, uint32_t len)
+{
+	msg_t msg;
+	int res;
+	i2c_devctl_t *in = (i2c_devctl_t *)msg.i.raw;
+	i2c_devctl_t *out = (i2c_devctl_t *)msg.o.raw;
+
+	if (!common.initialized)
+		return -EIO;
+
+	msg.type = mtDevCtl;
+
+	msg.i.data = (uint8_t *)data; /* FIXME: dropping const because of broken msg_t declaration */
+	msg.i.size = len;
+	msg.o.data = NULL;
+	msg.o.size = 0;
+
+	in->i.type = i2c_devctl_bus_write;
+	in->i.dev_addr = dev_addr;
+
+	if ((res = msgSend(common.i2c_oid.port, &msg)) < 0)
+		return res;
+
+	return out->o.err;
+}
+
+
+/* Performs i2c generic read operation from the given slave device. */
+int i2c_busRead(uint8_t dev_addr, uint8_t *data_out, uint32_t len)
+{
+	msg_t msg;
+	int res;
+	i2c_devctl_t *in = (i2c_devctl_t *)msg.i.raw;
+	i2c_devctl_t *out = (i2c_devctl_t *)msg.o.raw;
+
+	if (!common.initialized)
+		return -EIO;
+
+	msg.type = mtDevCtl;
+
+	msg.i.data = NULL;
+	msg.i.size = 0;
+	msg.o.data = data_out;
+	msg.o.size = len;
+
+	in->i.type = i2c_devctl_bus_read;
+	in->i.dev_addr = dev_addr;
+
+	if ((res = msgSend(common.i2c_oid.port, &msg)) < 0)
+		return res;
+
+	return out->o.err;
+}
+
+
+/* Performs i2c regiester read operation from the given slave device */
+int i2c_regRead(uint8_t dev_addr, uint8_t reg_addr, uint8_t *data_out, uint32_t len)
+{
+	msg_t msg;
+	int res;
+	i2c_devctl_t *in = (i2c_devctl_t *)msg.i.raw;
+	i2c_devctl_t *out = (i2c_devctl_t *)msg.o.raw;
+
+	if (!common.initialized)
+		return -EIO;
+
+	msg.type = mtDevCtl;
+
+	msg.i.data = NULL;
+	msg.i.size = 0;
+	msg.o.data = data_out;
+	msg.o.size = len;
+
+	in->i.type = i2c_devctl_reg_read;
+	in->i.dev_addr = dev_addr;
+	in->i.reg_addr = reg_addr;
+
+	if ((res = msgSend(common.i2c_oid.port, &msg)) < 0)
+		return res;
+
+	return out->o.err;
+}
--- a/i2c/imx6ull/imx6ull-i2c.c
+++ b/i2c/imx6ull/imx6ull-i2c.c
@@ -1,0 +1,129 @@
+/*
+ * Phoenix-RTOS
+ *
+ * i.MX 6ULL I2C BUS driver
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Marek Bialowas
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <sys/msg.h>
+#include <posix/utils.h>
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <errno.h>
+
+#include <i2c.h>
+#include <i2c-msg.h>
+
+
+static int dev_ctl(msg_t *msg)
+{
+	i2c_devctl_t *in = (i2c_devctl_t *)msg->i.raw;
+
+	switch (in->i.type) {
+		case i2c_devctl_bus_write:
+			return i2c_busWrite(in->i.dev_addr, msg->i.data, msg->i.size);
+
+		case i2c_devctl_bus_read:
+			return i2c_busRead(in->i.dev_addr, msg->o.data, msg->o.size);
+
+		case i2c_devctl_reg_read:
+			return i2c_regRead(in->i.dev_addr, in->i.reg_addr, msg->o.data, msg->o.size);
+
+		default:
+			return -EINVAL;
+	}
+}
+
+
+static void thread(void *arg)
+{
+	uint32_t port = (uint32_t)arg;
+	msg_t msg;
+	unsigned long int rid;
+	int err;
+
+	for (;;) {
+		if ((err = msgRecv(port, &msg, &rid)) < 0) {
+			printf("i2c: msgRecv returned error: %s\n", strerror(-err));
+			if (err == -EINTR)
+				continue;
+			else
+				break; /* EINVAL (invalid/closed port) or ENOMEM - fatal error */
+		}
+
+		switch (msg.type) {
+			case mtOpen:
+			case mtClose:
+				msg.o.io.err = EOK;
+				break;
+			case mtRead:
+			case mtWrite:
+				msg.o.io.err = -EINVAL;
+				break;
+			case mtDevCtl: {
+				i2c_devctl_t *out = (i2c_devctl_t *)msg.o.raw;
+				out->o.err = dev_ctl(&msg);
+				break;
+			}
+			default:
+				msg.o.io.err = -EINVAL;
+				break;
+		}
+
+		msgRespond(port, &msg, rid);
+	}
+}
+
+
+static void print_usage(const char *progname)
+{
+	printf("Usage: %s [i2c_bus_no <1,4>]\n", progname);
+}
+
+
+int main(int argc, char **argv)
+{
+	oid_t dev;
+	unsigned int dev_no;
+	uint32_t port;
+	char devname[sizeof("i2cX")];
+
+	if (argc != 2) {
+		print_usage(argv[0]);
+		return 1;
+	}
+
+	dev_no = atoi(argv[1]);
+	if (i2c_init(dev_no) < 0) {
+		printf("i2c: initialization failed\n");
+		return 1;
+	}
+
+	if (portCreate(&port) != EOK) {
+		printf("i2c: port allocation failed\n");
+		return 2;
+	}
+
+	dev.port = port;
+	dev.id = 0;
+
+	snprintf(devname, sizeof(devname), "i2c%u", dev_no % 10);
+	if (create_dev(&dev, devname) < 0) {
+		printf("i2c: could not create device\n");
+		return 3;
+	}
+
+	thread((void *)port);
+
+	printf("i2c: exiting\n");
+	return 0;
+}
--- a/i2c/imx6ull/libimx6ull-i2c.c
+++ b/i2c/imx6ull/libimx6ull-i2c.c
@@ -1,0 +1,377 @@
+/*
+ * Phoenix-RTOS
+ *
+ * i.MX 6ULL I2C driver
+ *
+ * Copyright 2021 Phoenix Systems
+ * Author: Marek Bialowas
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <sys/mman.h>
+#include <sys/msg.h>
+#include <sys/platform.h>
+#include <sys/threads.h>
+#include <sys/interrupt.h>
+
+#include <stdint.h>
+#include <unistd.h>
+#include <errno.h>
+#include <i2c.h>
+
+#include <phoenix/arch/imx6ull.h>
+
+
+/* clang-format off */
+/* NOTE: i2c registers are 16-bit and every second offset is reserved for future use */
+enum { iadr = 0, ifdr = 2, i2cr = 4, i2sr = 6, i2dr = 8 };
+enum { i2c_cmd_write = 0x0, i2c_cmd_read = 0x1 };
+/* clang-format on */
+
+static const uint32_t i2c_addr[] = { 0x021a0000, 0x021a4000, 0x021a8000, 0x021f8000 };
+static const int clocks[] = { pctl_clk_i2c1_serial, pctl_clk_i2c2_serial, pctl_clk_i2c3_serial, pctl_clk_i2c4_serial };
+
+
+typedef struct {
+	int pctl;
+	char val;
+} i2c_pctl_t;
+
+static const i2c_pctl_t i2c_pctl_mux[4][2] = {
+	{ { pctl_mux_gpio1_02, 0 }, { pctl_mux_gpio1_03, 0 } },
+	{ { pctl_mux_gpio1_00, 0 }, { pctl_mux_gpio1_01, 0 } },
+	{ { pctl_mux_lcd_d1, 4 }, { pctl_mux_lcd_d0, 4 } },
+	{ { pctl_mux_lcd_d3, 4 }, { pctl_mux_lcd_d2, 4 } },
+};
+
+static const i2c_pctl_t i2c_pctl_isel[4][2] = {
+	{ { pctl_isel_i2c1_scl, 0 }, { pctl_isel_i2c1_sda, 1 } },
+	{ { pctl_isel_i2c2_scl, 1 }, { pctl_isel_i2c2_sda, 1 } },
+	{ { pctl_isel_i2c3_scl, 2 }, { pctl_isel_i2c3_sda, 2 } },
+	{ { pctl_isel_i2c4_scl, 2 }, { pctl_isel_i2c4_sda, 2 } },
+};
+
+
+static const unsigned i2c_int_no[] = { 32 + 36, 32 + 37, 32 + 38, 32 + 35 };
+
+
+static struct {
+	volatile uint16_t *base;
+	uint16_t dev_no;
+	int initialized;
+
+	handle_t cond;
+	handle_t inth;
+	handle_t lock;
+} i2c = { 0 };
+
+/* Not implemented:
+ *  - slave mode
+ *  - multi-master mode (arbitration lost status flag inspection)
+ *  - register readout could be written by generating RESTART event instead of STOP / START (beneficial in multi-master mode)
+ */
+
+
+static void platform_init(void)
+{
+	platformctl_t pctl;
+	unsigned int i;
+
+	/* enable clock */
+	pctl.action = pctl_set;
+	pctl.type = pctl_devclock;
+	pctl.devclock.dev = clocks[i2c.dev_no - 1];
+	pctl.devclock.state = 0x3;
+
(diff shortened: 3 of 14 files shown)


nalajcie:
The build currently fails due to:
- https://github.com/phoenix-rtos/phoenix-rtos-build/pull/85

i2c/common/libi2c-msg.c
```diff
@@ -0,0 +1,123 @@
...
+	int initialized;
+} common;
+
+/* opens the device */
+int i2c_init(unsigned int dev_no)
+{
+	char devname[sizeof("/dev/i2cX")];
+	int res = 0;
+
+	sprintf(devname, "/dev/i2c%u", dev_no % 10);
```
kemonats, line 34: Maybe instead `sprintf()`:
```suggestion
	char devname[] = "/dev/i2cX";
	int res = 0;

	devname[8] = '0' + (dev_no % 10);
```
it will be faster.

i2c/imx6ull/imx6ull-i2c.c
```diff
@@ -0,0 +1,123 @@
...
+	char devname[sizeof("i2cX")];
+
+	if (argc != 2) {
+		print_usage(argv[0]);
+		return 1;
+	}
+
+	dev_no = atoi(argv[1]);
+	if (i2c_init(dev_no) < 0) {
+		printf("i2c: initialization failed\n");
```
kemonats, line 102: It is better to send error messages to `stderr`.

rtc/pcf85363a.c
```diff
@@ -0,0 +1,371 @@
...
+		/* no weekdays in alarm / event time registers */
+		rtc_time->tm_mday = from_bcd(timestamp[3]);       /* 1-31 */
+		rtc_time->tm_mon = from_bcd(timestamp[4]) - 1;    /* 0-11; RTC holds 1-12 */
+		rtc_time->tm_year = 100 + from_bcd(timestamp[5]); /* year since 1900, RTC holds only last 2 digids - assume 20XX */
+	}
+	else {
+		rtc_time->tm_mday = from_bcd(timestamp[3]);       /* 1-31 */
+		rtc_time->tm_wday = from_bcd(timestamp[4]);       /* weekday 0-6; 0: sunday (RTC has the same semantics) */
+		rtc_time->tm_mon = from_bcd(timestamp[5]) - 1;    /* 0-11; RTC holds 1-12 */
+		rtc_time->tm_year = 100 + from_bcd(timestamp[6]); /* year since 1900, RTC holds only last 2 digids - assume 20XX */
```
kemonats, line 89: Maybe `tm_mon` and `tm_year` correction after the `if` statement.

rtc/pcf85363a.c
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+
+	rtc_time->tm_yday = 0;  /* FIXME?: yday not stored in RTC */
+	rtc_time->tm_isdst = 0; /* RTC holds UTC time, no DST */
```
kemonats, line 94: Earlier it is zeroing the entire structure.

i2c/common/libi2c-msg.c
agkaminski: That's what I was going to suggest ;)

agkaminski: This is also less error prone - string is in one place.

i2c/common/libi2c-msg.c
```diff
@@ -0,0 +1,123 @@
...
+	msg.o.data = NULL;
+	msg.o.size = 0;
+
+	i->type = i2c_devctl_bus_write;
+	i->dev_addr = dev_addr;
+
+	if ((res = msgSend(common.i2c_oid.port, &msg)) < 0)
+		return res;
+
+	return msg.o.io.err;
```
agkaminski, line 67: If you use devctl, don't use fields for IO operations. For `mtDevCtl` there are only `raw` fields available. You should pass error code in a custom struct (so maybe just add `err` field to `i2c_devctl_t` and pass it in both `msg.i.raw `and `msg.o.raw`?)

i2c/common/libi2c-msg.c
```diff
@@ -0,0 +1,123 @@
...
+	msg.o.data = data_out;
+	msg.o.size = len;
+
+	i->type = i2c_devctl_bus_read;
+	i->dev_addr = dev_addr;
+
+	if ((res = msgSend(common.i2c_oid.port, &msg)) < 0)
+		return res;
+
+	return msg.o.io.err;
```
agkaminski, line 94: As above, this field is for mtOpen, mtRead etc, but not for mtDevCtl.

i2c/common/libi2c-msg.c
```diff
@@ -0,0 +1,123 @@
...
+	msg.o.size = len;
+
+	i->type = i2c_devctl_reg_read;
+	i->dev_addr = dev_addr;
+	i->reg_addr = reg_addr;
+
+	if ((res = msgSend(common.i2c_oid.port, &msg)) < 0)
+		return res;
+
+	return msg.o.io.err;
```
agkaminski, line 122: As above, this field is for mtOpen, mtRead etc, but not for mtDevCtl.

i2c/imx6ull/imx6ull-i2c.c
```diff
@@ -0,0 +1,123 @@
...
+static void thread(void *arg)
+{
+	uint32_t port = (uint32_t)arg;
+	msg_t msg;
+	unsigned long int rid;
+
+	for (;;) {
+		if (msgRecv(port, &msg, &rid) < 0) {
+			memset(&msg, 0, sizeof(msg));
+			msgRespond(port, &msg, rid);
```
agkaminski, line 56: I think we shouldn't reply if `msgRecv` returns error

i2c/imx6ull/imx6ull-i2c.c
```diff
@@ -0,0 +1,123 @@
...
+	printf("Usage: %s [i2c_bus_no <1,4>]\n", progname);
+}
+
+
+int main(int argc, char **argv)
+{
+	oid_t dev;
+	unsigned int dev_no;
+	uint32_t port;
+	char devname[sizeof("i2cX")];
```
agkaminski, line 93: As above, this is error prone and sprintf is slower than just replacing one char in string

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+static void platform_init(void)
+{
+	platformctl_t pctl;
+	unsigned int i;
+
+	/* enable clock */
+	pctl.action = pctl_set;
+	pctl.type = pctl_devclock;
+	pctl.devclock.dev = clocks[i2c.dev_no - 1];
+	pctl.devclock.state = 0b11;
```
agkaminski, line 85: Binaries are GNU extension. Use either hex or decimal

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+
+	platformctl(&pctl);
+
+	pctl.iopad.pad = pctl_pad_gpio1_03;
+	platformctl(&pctl);
+#endif
+}
+
+
+static int is_bus_busy(void)
```
agkaminski, line 130: Use camelCase

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+{
+	return *(i2c.base + i2sr) & (1 << 5);
+}
+
+
+static int wait_bus_busy(int for_busy)
```
agkaminski, line 136: Use camelCase

rtc/pcf85363a.c
nalajcie: In here the comments are important, not the code. Leaving as-is.

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+	*(i2c.base + i2sr) = 0; /* clear previous interrupts */
+	*(i2c.base + i2dr) = byte;
+
+	/* wait for byte to be sent */
+	ret = i2c_trx_complete();
+	if (ret < 0)
+		return ret;
+
+	ret = i2c_acked();
+	return ret;
```
agkaminski, line 205:
```suggestion
	return  i2c_acked();
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+	ret = wait_bus_busy(1);
+
+	/* send first byte - DEVICE WRITE request */
+	if (ret == 0)
+		ret = write_byte((dev_addr << 1) | 0x0);
+
+	/* send data */
+	for (i = 0; (ret == 0) && (i < len); ++i) {
+		ret = write_byte(data[i]);
+	}
```
agkaminski, line 232:
```suggestion
	for (i = 0; (ret == 0) && (i < len); ++i)
		ret = write_byte(data[i]);
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+
+	/* generate START: enable master mode, enable TX, enable interrupt */
+	*(i2c.base + i2cr) |= (1 << 6) | (1 << 5) | (1 << 4);
+
+	/* wait for the bus to be busy */
+	ret = wait_bus_busy(1);
+
+	/* send first byte - DEVICE WRITE request */
+	if (ret == 0)
+		ret = write_byte((dev_addr << 1) | 0x0);
```
agkaminski, line 227: `| 0x0` does nothing
```suggestion
		ret = write_byte(dev_addr << 1);
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+	if (is_bus_busy())
+		return -EBUSY;
+
+	/* generate START: enable master mode, enable TX for sending register address */
+	*(i2c.base + i2cr) |= (1 << 6) | (1 << 5) | (1 << 4);
+
+	/* wait for the bus to be busy */
+	ret = wait_bus_busy(1);
+	if (ret < 0)
+		goto generate_stop;
```
agkaminski, line 261: `goto`, for better or for worse, is forbidden by our coding guidelines.

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+
+	/* setup bus for data readout - disable TX, enable ACKing for every byte except the last one */
+	temp = *(i2c.base + i2cr);
+	temp &= ~(1 << 4);
+	if (len > 1)
+		temp &= ~(1 << 3);
+	else
+		temp |= (1 << 3);
```
agkaminski, line 274:
```suggestion
	temp = *(i2c.base + i2cr) & ~((1 << 4) | (1 << 3));
	if (len <= 1)
		temp |= (1 << 3);
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+int i2c_regRead(uint8_t dev_addr, uint8_t reg_addr, uint8_t *data_out, uint32_t len)
+{
+	int ret;
+
+	ret = i2c_busWrite(dev_addr, &reg_addr, 1);
+	if (ret < 0)
+		return ret;
+
+	ret = i2c_busRead(dev_addr, data_out, len);
+	return ret;
```
agkaminski, line 319:
```suggestion
	return i2c_busRead(dev_addr, data_out, len);
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+
+int i2c_init(unsigned int dev_no)
+{
+	i2c.dev_no = dev_no;
+
+	if ((i2c.dev_no < 1) || (i2c.dev_no > (sizeof(i2c_addr) / sizeof(i2c_addr[0]))))
+		return -1;
+
+	/* initialize */
+	i2c.base = mmap(NULL, 0x1000, PROT_WRITE | PROT_READ, MAP_DEVICE, OID_PHYSMEM, i2c_addr[i2c.dev_no - 1]);
```
agkaminski, line 331:
```suggestion
	i2c.base = mmap(NULL, _PAGE_SIZE, PROT_WRITE | PROT_READ, MAP_DEVICE, OID_PHYSMEM, i2c_addr[i2c.dev_no - 1]);
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+
+	if (i2c.base == MAP_FAILED)
+		return -2;
+
+	platform_init();
+
+	/* prepare interrupt routine */
+	if (mutexCreate(&i2c.lock) != EOK)
+		return -3;
```
agkaminski, line 340: We might want to unmap `i2c.base` here

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,361 @@
...
+
+	if (condCreate(&i2c.cond) != EOK)
+		return -3;
```
agkaminski, line 343: We might want to unmap `i2c.base` and destroy mutex here

i2c/imx6ull/imx6ull-i2c.c
nalajcie: For userspace programs - yes, for drivers - not really. Right now it's messy (stdout/stderr pinned to console so no possibility to read the output from drivers remotely).

The obvious choice for now would be to use syslog() to log all of the driver messages, but we decided to implement dmesg-alike circular buffer (with `/dev/kmsg` interface) and pin it to stdout of all of the drivers and than use syslogd to automatically pump the logs from `/dev/kmsg` to syslog (on bigger systems) or be able do inspect dmesg buffer directly.  **Please note that using line-buffered stream (`stdout` as opposed to `stderr`) would be much more useful then to prevent mixing log messages from multiple processes**.

I don't see any real benefit for using `stderr` now, I would much more welcome dedicated `log` wrappers which will add message priority to the dmesg messages (along with the introduction of `dmesg`).

Leaving as-is on purpose.

rtc/pcf85363a.c
```diff
@@ -0,0 +1,371 @@
...
+}
+
+
+/* assume errors during setup are not fatal as they control periperial RTC functions, not the main one */
+static void setup_rtc(void)
+{
+	const uint8_t battery_check_magic = 0x5a;
+	const uint8_t event_config[] = {
+		RTC_TSR_MODE_REG_ADDR,
+		(0b10 << 6) | (0b001 << 2), /* TSR3 = LB, TSR2 = FB */
```
agkaminski, line 300: Binaries are GNU extention

rtc/pcf85363a.c
```diff
@@ -0,0 +1,371 @@
...
+	const uint8_t reset_flags[] = {
+		RTC_FLAGS_REG_ADDR,
+		0,
+	};
+
+	uint8_t buf[1] = { 0 };
+	int res;
+
+	if ((res = i2c_busWrite(DEV_ADDR, event_config, sizeof(event_config))) < 0)
+		printf("rtc: failed to setup the events configuration: %s\n", strerror(res));
```
agkaminski, line 315: Maybe use `stderr` to output errors?

sensor/temp-nct75.c
```diff
@@ -0,0 +1,126 @@
...
+#include <string.h>
+#include <errno.h>
+
+#include <sys/msg.h>
+#include <posix/utils.h>
+
+#include <i2c.h>
+
+
+static const uint32_t INVALID_TEMP = -1000 * 1000u;
```
agkaminski, line 26:
```suggestion
static const uint32_t INVALID_TEMP = (unsigned)(-1000 * 1000);
```

sensor/temp-nct75.c
```diff
@@ -0,0 +1,126 @@
...
+{
+	uint8_t temp_raw[2];
+	int16_t temp_enc;
+
+	if (i2c_regRead(DEV_ADDR, TEMP_REG_ADDR, temp_raw, sizeof(temp_raw)) < 0)
+		return INVALID_TEMP;
+
+	/* 12-bit left-adjusted, 2s complement, every bit is 0.0625 C */
+	temp_enc = (temp_raw[0] << 8) | temp_raw[1];
+	return (temp_enc >> 5) * 125;
```
agkaminski, line 43: Hmm, if it is 12 bit left-adjusted, shouldn't it be here `(temp_enc >> 4)`?

sensor/temp-nct75.c
```diff
@@ -0,0 +1,126 @@
...
+	printf("Usage: %s [i2c_bus_no <1,4>] [temp_device_no]\n", progname);
+}
+
+
+int main(int argc, char **argv)
+{
+	unsigned int i2c_bus_no, dev_no;
+	uint32_t port;
+	oid_t dev;
+	char devname[sizeof("tempX")];
```
agkaminski, line 98: As above, `char devname[] = "tempX";` is less error prone and faster.

i2c/common/libi2c-msg.c
nalajcie: no docs and does it really matter? As we're using `msg` fields directly everywhere, I don't consider their placement an implementation detail.

Will change to some other solution but I believe we were talking about providing one common `err` field in `msg.o` also instead of repeating it in every union member...

i2c/common/libi2c-msg.c
nalajcie: hehh, seems for mtCreate I need to use `msg.o.create.err` (the and other errors for `mtLookup`, and `mtGetAttr`, `mtSetAttr` ...) :P

i2c/imx6ull/imx6ull-i2c.c
nalajcie: no docs, copied from other available sources. What should be done?

i2c/imx6ull/imx6ull-i2c.c
nalajcie: will change to `snprintf` because of  C89 and readability

i2c/imx6ull/libimx6ull-i2c.c
nalajcie: Ok, I assumed it is a requirement for external functions only

i2c/imx6ull/libimx6ull-i2c.c
nalajcie: will do, remnant of it being a part of other function

i2c/imx6ull/libimx6ull-i2c.c
nalajcie: I know the code is a no-op and will be optimized out but it highlights the intent (it's symmetric to `i2c_busRead`). 

I can define `I2C_OPERATION_WRITE` and `I2C_OPERATION_READ` if that is more preferable.

sensor/temp-nct75.c
nalajcie: yeah, the code is sneaky here:
```c
(temp_enc >> 4) * 62.5 ~= (temp_enc >> 4) / 2 * 125 = (temp_enc >> 5) * 125
```
loosing one bit of information for the sake of not dealing with floats :)

sensor/temp-nct75.c
nalajcie: hmm, actually will change the define to `int32_t` to be consistent with return value of `getTemp`

sensor/temp-nct75.c
nalajcie: will go for `snprintf`

i2c/imx6ull/libimx6ull-i2c.c
nalajcie: Yeah, that's just because somebody didn't read the Dijkstra article thoughtfully, but I won't argue with You on that here.

I considered other possibilities - I'm against empty `do-while(0)` as it's only a syntactic sugar over goto on top of increasing cyclomatic complexity and also being advisory rule in MISRA:2012 (_15.4 There should be no more than one break statement used to terminate any iteration statement_, so in the same basket as _15.1 Goto statement should not be used_).

I tried moving body of the function to the separate sub-function but it seemed less readable so I settled on using `goto` for "error handling" causes - which actually is consistent with the allowed usage in linux kernel **and implicitly allowed / proposed by Dijkstra himself** :)
Screenshot_20210923_112959


I'll go for putting body of the function into separate sub-function but it would be less readable that way,

i2c/common/libi2c-msg.c
nalajcie: introduced i/o union of structs in `i2c_devctl_t`

i2c/imx6ull/imx6ull-i2c.c
nalajcie: ok, I assume it's a fatal error so it will result in device driver exiting.

rtc/pcf85363a.c
nalajcie: written in detail above why I think it's better to use `stdout` (although I would prefer to use `syslog` directly because of multiple logging levels available).

i2c/imx6ull/imx6ull-i2c.c
agkaminski: Such case was never seen in nature (only if e.g. port hasn't been created). Commonly it we do just `continue`

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,368 @@
...
+	return 0;
+}
+
+
+static int i2c_intr(unsigned int intr, void *data)
+{
+	/* clear interrupt */
+	*(i2c.base + i2sr) &= ~(1 << 1);
+
+	return i2c.cond;
```
agkaminski, line 156: Returning cond is an old style API. Now cond is registered within irq handler

```suggestion
	return 1;
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -0,0 +1,368 @@
...
+	platform_init();
+
+	/* prepare interrupt routine */
+	if (mutexCreate(&i2c.lock) != EOK) {
+		munmap((void *)i2c.base, _PAGE_SIZE);
+		return -3;
+	}
+
+	if (condCreate(&i2c.cond) != EOK) {
+		munmap((void *)i2c.base, _PAGE_SIZE);
```
agkaminski, line 348: Missing `resourceDestroy(i2c.lock);`

agkaminski: From what I understand this driver can be used as lib - so failure here not always cause process termination

i2c/imx6ull/imx6ull-i2c.c
nalajcie: Ok, after looking at the kernel code I see that msgRecv() can exit with (on MMU archs):
 - `EINVAL` if the port is invalid or was closed (I don't know how a port can be closed, can it be done by other process?)
 - `EINTR` if we receive a signal, so my solution needs to be fixed (non-fatal error)
Screenshot_20210923_130845
 - `ENOMEM` if message copying to the target process would fail - this is a fatal error and we don't know what to do in case of system OOM, so exiting from non-critical driver might be a way to keep floating

I'll fix the the message loop to tread `EINTR` as non-fatal and exit on other errors.

i2c/imx6ull/libimx6ull-i2c.c
nalajcie: thx, done.

i2c/imx6ull/libimx6ull-i2c.c
nalajcie: thx, done.
Yeah, you're right it might be used that way (I've focused only on my use cases - so i2c device drivers exit if they cannot initialize i2c bus). Added also `i2c.initialized` to avoid accessing `i2c.base` if it's not mapped correctly and somebody would not check `i2c_init` return value.

agkaminski (approved):

PR #150, opened 2021-11-05
imxrt-multi: Use integer values for setting uart baudrate

MaciejPurski:
## Description

Use speed_t type only, when communicating with libtty. Internally use
a plain uint32_t.

JIRA: RTOS-122

## Motivation and Context

This change allows setting uart baudrate through BOARD_CONFIG using integer values instead of speed_t. Thanks to this change we can set uart baudrate in plo, kernel and in imxrt-multi using the same value without the need for any conversion. It makes setting uart baudrate less error prone.

## Types of changes

In all projects using imxrt-multi, which set the baudrate value through BOARD_CONFIG, we shall change the values from speed_t to integer.

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt1064).

## Checklist:

## Special treatment

### DIFF
--- a/multi/imxrt-multi/uart.c
+++ b/multi/imxrt-multi/uart.c
@@ -194,28 +194,27 @@ static void set_cflag(void *_uart, tcflag_t* cflag)
 }
 
 
-static uint32_t calculate_baudrate(speed_t baud)
+static uint32_t calculate_baudrate(uint32_t baud)
 {
-	uint32_t osr, sbr, bestSbr = 0, bestOsr = 0, bestDiff, t, tDiff, baud_rate;
+	uint32_t osr, sbr, bestSbr = 0, bestOsr = 0, bestDiff, t, tDiff;
 
-	bestDiff = baud_rate = (uint32_t)libtty_baudrate_to_int(baud);
-
-	if ((int)baud_rate <= 0) {
+	if (baud == 0)
 		return 0;
-	}
+
+	bestDiff = baud;
 
 	for (osr = 4; osr <= 32; ++osr) {
 		/* find sbr value in range between 1 and 8191 */
-		sbr = (UART_CLK / (baud_rate * osr)) & 0x1fff;
+		sbr = (UART_CLK / (baud * osr)) & 0x1fff;
 		sbr = (sbr == 0) ? 1 : sbr;
 
 		/* baud rate difference based on temporary osr and sbr */
-		tDiff = UART_CLK / (osr * sbr) - baud_rate;
+		tDiff = UART_CLK / (osr * sbr) - baud;
 		t = UART_CLK / (osr * (sbr + 1));
 
 		/* select best values between sbr and sbr+1 */
-		if (tDiff > baud_rate - t) {
-			tDiff = baud_rate - t;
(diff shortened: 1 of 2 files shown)


multi/imxrt-multi/uart.c
```diff
@@ -194,28 +194,24 @@ static void set_cflag(void *_uart, tcflag_t* cflag)
...
-
-	if ((int)baud_rate <= 0) {
-		return 0;
-	}
+	bestDiff = baud;
 
 	for (osr = 4; osr <= 32; ++osr) {
 		/* find sbr value in range between 1 and 8191 */
-		sbr = (UART_CLK / (baud_rate * osr)) & 0x1fff;
+		sbr = (UART_CLK / (baud * osr)) & 0x1fff;
```
gerard5, line 205: If `baud=0` division by zero. The removed `if` check shall be restored/updated for that case.

multi/imxrt-multi/uart.c
```diff
@@ -235,7 +231,7 @@ static void set_baudrate(void *_uart, speed_t baud)
 	uint32_t reg, t;
 	uart_t *uartptr = (uart_t *)_uart;
 
-	reg = calculate_baudrate(baud);
+	reg = calculate_baudrate((uint32_t)libtty_baudrate_to_int(baud));
```
gerard5, line 234: What if `libtty_baudrate_to_int(baud)` return `-1`, or `0` ?

multi/imxrt-multi/uart.c
MaciejPurski: I have updated this check.

agkaminski (approved):

multi/imxrt-multi/uart.c
gerard5: An update: 
I see that `libtty` does not check the case for `B0` and any `unknown` baud value that is passed to `set_baudrate` callback:
https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/05e9b3785492e1e361a9e8b9e6714b4ff18e2a24/tty/libtty/libtty.c#L443-L448
@agkaminski maybe either  the `set_baudrate` here in `imxrt-multi` for the case `baud <= 0` shall be kind of `no-op` (return;), because of undefined behaviour or the `libtty` should return error for the `ioctl` (see [`man termios`](https://man.openbsd.org/tcsetattr.3#GETTING_AND_SETTING_THE_BAUD_RATE) ).

agkaminski: Hmm: The zero baud rate, B0, is used to terminate the connection.  If
       B0 is specified, the modem control lines shall no longer be
       asserted.  Normally, this will disconnect the line.  CBAUDEX is a
       mask for the speeds beyond those defined in POSIX.1 (57600 and
       above).  Thus, B57600 & CBAUDEX is nonzero.

So it seems it should be no-op, but I don't know how to handle this

gerard5: @MaciejPurski, for me, the PR works ok on `imxrt117x`, so it can be mergerd now. Marking as resolved.

gerard5 (approved):

PR #154, opened 2021-11-16
stm32l4-multi: add libspi_readwrite fix

damianloew:
## Description

- Add waiting for transmit buffer empty after sending transmit data
- add first init read before save data from receive buffer

## Motivation and Context

Sometimes, when reading for example sensor address using SPI, unwanted received data (from first cycle) was presented in data register. To clear data register and prepare to read next byte from rx buffer, the register is read double times.

Different read values (before changes from this pr) using same method:
Screenshot from 2021-11-16 22-57-32

Signal from both 0xff and 0x69 readings:

Screenshot from 2021-11-16 23-24-14

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4x6 - nucleo, lsm6ds3 sensor).

## Checklist:

## Special treatment

### DIFF
--- a/multi/stm32l4-multi/libmulti/libspi.c
+++ b/multi/stm32l4-multi/libmulti/libspi.c
@@ -71,6 +71,12 @@ static unsigned char libspi_readwrite(libspi_ctx_t *ctx, unsigned char txd)
 	/* Initiate transmission */
 	*((volatile uint8_t *)(ctx->base + dr)) = txd;
 
+	/* Wait until TXNE==1 */
+	while (!(*(ctx->base + sr) & 0x2))
+		;
+	/* Clear data register */
+	rxd = *((volatile uint8_t *)(ctx->base + dr));
+
 	/* Wait until RXNE==1 */
 	while (!(*(ctx->base + sr) & 0x1))
 		;


agkaminski:
This introduces race condition IMO. We might clear data register either before or after `*(ctx->base + sr) & 0x1` becomes true

nalajcie:
I'm not sure what really happens here. SPI is a full-duplex protocol, You won't receive response for the request You're sending at the same time - probably You need to send 2x uint8_t and should treat the second RX byte as a response to first TX byte reqest (this is how it's usually done on SPI with 8-bit register addressing). For example - see SPI usage in ad7779 driver:
https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/09af4d2e69510f7094de3e86f5c7d0cc4391918f/adc/ad7779/ad7779.c#L79-L97

```
TX: [address] [0x0] [0x0] [0x0] [...]
RX: [nothing] [rx0] [rx1] [rx2] [...]
```

As for the screenshot - if second program execution results in always-invalid reads it seems that there is a problem with SPI initialization / deinitialization - not the exchange operation (I'm just guessing as I don't know the source code of `accread`)

agkaminski:
So it seems that selecting SPI mode 0 fixes the problem. Can I close this PR?

damianloew:
> So it seems that selecting SPI mode 0 fixes the problem. Can I close this PR?

Yes, that's right. I still don't know why there is such a problem when CPOL and CPHA are set to 1 (the same transmission, but the idle clock state is 1). This configuration better corresponds to the sensor documentation (where the clock's idle state is 1), but with `00` works fine. Thanks for all your responses.

PR #158, opened 2021-12-06
imx6ull-flash: rewrite initialization 

nalajcie:
## Description
- rewrite options parsing
- parse/validate options first, then create necessary partitions
- common logging
- introduce partition names with symlink creation
- select rootfs to be mounted based on SRC secondary boot 

## Motivation and Context
JIRA: DTR-181

Needed for FW upgrade (the same kernel image for primary/secondary boot, current rootfs detection, etc.)

## Types of changes

- [x] Breaking change: **parameter syntax for `imx6ull-flash` changed, PRs for `psu`/`project` would follow**

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/storage/imx6ull-flash/flashsrv.c
+++ b/storage/imx6ull-flash/flashsrv.c
@@ -5,6 +5,7 @@
 #include <sys/threads.h>
 #include <sys/stat.h>
 #include <sys/mman.h>
+#include <sys/reboot.h>
 #include <sys/file.h>
 #include <string.h>
 #include <errno.h>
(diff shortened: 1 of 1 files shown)


storage/imx6ull-flash/flashsrv.c
```diff
@@ -767,20 +782,138 @@ static void daemonize(void)
...
+				}
+
+				break;
+
+			case 'p': /* start:size[:name] */
+				/* TODO: switch to positional arguments to highlight partition order matter */
+
+				errno = 0;
+				part_start = strtol(optarg, &p, 10);
+				if (*p++ != ':') {
```
agkaminski, line 817: `char *p` might be uninitialized in this case

storage/imx6ull-flash/flashsrv.c
```diff
@@ -743,21 +753,26 @@ static void flashsrv_devThread(void *arg)
...
+	int ret;
 
 	p = malloc(sizeof(*p));
 
+	if (p == NULL)
+		return -1;
+
 	p->start = start;
 	p->size = size;
+	p->name = name;
```
agkaminski, line 768: `strdup()`? You pass to this function `name` placed in non static storage

nalajcie: I believe `**argv` is mine to rule inside the program? I can strdup if that would be more readable.

Second thought: if we would ever add partition table support, `strdup` hete might be more straightforward, will fix in a sec

storage/imx6ull-flash/flashsrv.c
nalajcie: how? if the entire input string is valid, `strtol` will set `*endptr` as a pointer to `\0`

agkaminski: Yes, didn't notice p in strtol

agkaminski (approved):

PR #170, opened 2022-01-10
stm32l4-multi/flash: Add bank switching (dual boot)

Darchiv:
JIRA: SKAL-258

## Description

This change introduces dual booting to STM32L4 targets. The new message (flash_switch) toggles a dual boot flag and causes a write to flash option bytes. The device is then reset by flash option loader (OBL_LAUNCH) - a normal reset does not apply the changes during boot.

The reset itself could be moved to the kernel (pctl_reboot) but it still needs flash-related code (unlocking the flash registers).

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: stm32l4.

## Checklist:

## Special treatment

### DIFF
--- a/multi/stm32l4-multi/flash.c
+++ b/multi/stm32l4-multi/flash.c
@@ -306,6 +306,15 @@ int flash_switchBanks(void)
 }
 
 
+void flash_getInfo(flashinfo_t *info)
+{
+	info->dualbank = !!(_flash_getOptions() & (1 << opt_dualbank));
+	info->dualboot = !!(_flash_getOptions() & (1 << opt_bfb2));
+	info->remap = flash_common.bankflip;
+	info->activebank = flash_common.activebank;
+}
+
+
 int flash_init(void)
 {
 	volatile unsigned int *syscfg = (void *)0x40010000;
--- a/multi/stm32l4-multi/flash.c
+++ b/multi/stm32l4-multi/flash.c
@@ -28,11 +28,14 @@ enum { flash_acr = 0, flash_pdkeyr, flash_keyr, flash_optkeyr, flash_sr, flash_c
 	flash_optr = flash_eccr + 2, flash_pcrop1sr, flash_pcrop1er, flash_wrp1ar, flash_wrp1br,
 	flash_pcrop2sr = flash_wrp1br + 5, flash_pcrop2er, flash_wrp2ar, flash_wrp2br };
 
+enum { opt_bfb2 = 20, opt_dualbank = 21 };
+
 
 struct {
 	volatile unsigned int *flash;
 	volatile int operr;
 	unsigned int bankflip;
+	unsigned int activebank;
 	handle_t lock;
 	handle_t irqcond;
 	handle_t irqlock;
@@ -92,6 +95,15 @@ static void _program_unlock(void)
 	*(flash_common.flash + flash_keyr) = 0x45670123;
 	dataBarier();
 	*(flash_common.flash + flash_keyr) = 0xcdef89ab;
+	dataBarier();
+}
+
+
+static void _flash_optunlock(void)
+{
+	*(flash_common.flash + flash_optkeyr) = 0x08192a3b;
+	dataBarier();
+	*(flash_common.flash + flash_optkeyr) = 0x4c5d6e7f;
 	dataBarier();
 }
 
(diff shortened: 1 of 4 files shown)


multi/stm32l4-multi/flash.c
```diff
@@ -96,6 +99,15 @@ static void _program_unlock(void)
 }
 
 
+static void _program_optunlock(void)
```
agkaminski, line 102: Use module name prefix even for static functions

multi/stm32l4-multi/flash.c
```diff
@@ -232,6 +244,68 @@ size_t flash_writeData(uint32_t offset, const char *buff, size_t size)
...
+	*(flash_common.flash + flash_cr) |= 1 << 17;
+
+	err = _flash_wait();
+	_flash_clearFlags();
+
+	return err;
+}
+
+
+extern int flash_switchBanks(void)
```
agkaminski, line 275: Why is this declared as extern?

multi/stm32l4-multi/flash.c
```diff
@@ -232,6 +244,68 @@ size_t flash_writeData(uint32_t offset, const char *buff, size_t size)
...
+		err = 0;
+
+	if (!err) {
+		/* Reset by flash option loader */
+		*(flash_common.flash + flash_cr) |= 1 << 27;
+		dataBarier();
+
+		/* Reset failed */
+		err = -1;
+	}
```
agkaminski, line 299:
```suggestion
	if (!bfb2 || flash_common.activebank != 0) {
		err = _setOptions(opt_bfb2, !bfb2);
	}
	else {
		/* Reset by flash option loader */
		*(flash_common.flash + flash_cr) |= 1 << 27;
		dataBarier();

		/* Reset failed */
		err = -1;
	}
```

multi/stm32l4-multi/flash.c
```diff
@@ -232,6 +244,68 @@ size_t flash_writeData(uint32_t offset, const char *buff, size_t size)
...
+	err = _flash_wait();
+	_flash_clearFlags();
+
+	return err;
+}
+
+
+extern int flash_switchBanks(void)
+{
+	int err;
```
agkaminski, line 277:
```suggestion
	int err = 0;
```

multi/stm32l4-multi/flash.c
```diff
@@ -232,6 +244,68 @@ size_t flash_writeData(uint32_t offset, const char *buff, size_t size)
...
+	unsigned int bfb2;
+
+	mutexLock(flash_common.lock);
+	bfb2 = !!(_getOptions() & opt_bfb2);
+	_program_unlock();
+	_program_optunlock();
+
+	/* If BFB2==1 and Bank 1 is the active bank, then booting from Bank 2
+	must have failed so leave BFB2 as is and just retry booting */
```
agkaminski, line 286: Hmmm, I think it's ok, but we need to discuss this - what is the best approach in such case.

multi/stm32l4-multi/stm32l4-multi.c
```diff
@@ -87,6 +87,10 @@ static void handleMsg(msg_t *msg)
 			err = flash_writeData(imsg->flash_addr, msg->i.data, msg->i.size);
 			break;
 
+		case flash_switch:
+			err = flash_switchBanks();
```
agkaminski, line 91: Maybe some magic value? Should this msg come by accident (for any reason), it would be bad.

multi/stm32l4-multi/flash.c
```diff
@@ -232,6 +244,68 @@ size_t flash_writeData(uint32_t offset, const char *buff, size_t size)
...
+{
+	int err;
+	unsigned int t;
+
+	if ((err = _flash_wait()) < 0)
+		return err;
+
+	_flash_clearFlags();
+
+	t = *(flash_common.flash + flash_optr) & ~mask;
```
anglov, line 263: You could use _getOptions() & ~mask here

multi/stm32l4-multi/flash.c
```diff
@@ -232,6 +244,68 @@ size_t flash_writeData(uint32_t offset, const char *buff, size_t size)
...
+{
+	int err;
+	unsigned int bfb2;
+
+	mutexLock(flash_common.lock);
+	bfb2 = !!(_getOptions() & opt_bfb2);
+	_program_unlock();
+	_program_optunlock();
+
+	/* If BFB2==1 and Bank 1 is the active bank, then booting from Bank 2
```
anglov, line 285: Wonder if it's possible that bfb2 == 0 and Bank2 is active bank (fail to boot from Bank 1) or it can never happen.

anglov (commented):
Generally LGTM just point mimor thing and want to be sure about booting behavior

multi/stm32l4-multi/flash.c
agkaminski: It's ok

multi/stm32l4-multi/flash.c
Darchiv: Done, although I did not change prefixes for not my functions

multi/stm32l4-multi/flash.c
Darchiv: Copy typo, fixed

multi/stm32l4-multi/flash.c
Darchiv: Although your suggestion cleans up code, it also changes the logic - the reset will be performed only when booting from Bank 2 failed. The reset shall actually be performed in all circumstances (except when _setOptions fails).

multi/stm32l4-multi/flash.c
Darchiv: Done, see the comment above

multi/stm32l4-multi/stm32l4-multi.c
Darchiv: Good point, I applied the magic

multi/stm32l4-multi/flash.c
Darchiv: I am not convinced to that because the currently used idiom (at least the dominant one in multidrv) for register read-modify-write operations is `t = reg & ~mask; reg = t | val` where `reg` is a register direct access expression. If _getOptions() changes in the future then this line here will be invalid.

multi/stm32l4-multi/flash.c
Darchiv: According to ST, when bfb2 == 0 then the device boots directly from Bank 1 and does not perform any checks. Only when bfb2 == 1 the bootloader checks Bank 2's initial stack pointer and decides whether to boot from Bank 1 or Bank 2. So with bfb2 == 0 there is no fail-safe.

Darchiv:
I have also checked the effect of different reset types on bank switching:

- power/JTAG reset, IWDG reset, NVIC reset - all of these do not reload flash options and no switching occurs, tested on both banks;
- OBL_LAUNCH - reloads flash options and switches banks properly.

multi/stm32l4-multi/flash.c
agkaminski: Right

agkaminski (approved):
LGTM

anglov (approved):
LGTM

agkaminski:
Shall I merge this?

Darchiv:
> Shall I merge this?

Yes, please

PR #175, opened 2022-01-17
stm32l4-multi: Use create_dev for tty devices

anglov:
tty devices was not available on /dev
This commit fixes that

## Description

It's quite complicated because:
- Port for main console have to be first registered port (depending on configuration it's tty or multidrv port)
- Rootfs have to be initialized before tty want's to create dev on it
- tty will not work if rootfs is not created (it's not possible to create entries in /dev then)

## Motivation and Context

Required for use open("/dev/uartX') in application code again, after posixifyizing open on stm32l4. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (armv7m4-stm32l4x6).

psh run on target. Directory /dev appeared with tty, uart1 and uart3 files.

## Checklist:

## Special treatment

### DIFF
--- a/multi/stm32l4-multi/stm32l4-multi.c
+++ b/multi/stm32l4-multi/stm32l4-multi.c
@@ -237,17 +237,32 @@ int main(void)
 	int i;
 	oid_t oid;
 	static const char welcome[] = "multidrv: Started\n";
+#if CONSOLE_IS_TTY
+	unsigned int ttyConsolePort;
+#endif
 
 	priority(THREADS_PRIORITY);
 
-#if !CONSOLE_IS_TTY
-	/* called before init functions to redirect logs here */
+#if CONSOLE_IS_TTY
+	portCreate(&ttyConsolePort);
+#endif
 	portCreate(&common.port);
+
+#if BUILTIN_DUMMYFS
+	fs_init();
+#else
+	/* Wait for the filesystem */
+	while (lookup("/", NULL, &oid) < 0)
+		usleep(10000);
 #endif
 
 	rcc_init();
 	exti_init();
-	tty_init();
+#if CONSOLE_IS_TTY
+	tty_init(&ttyConsolePort);
+#else
+	tty_init(NULL);
+#endif
 	gpio_init();
 	spi_init();
 	adc_init();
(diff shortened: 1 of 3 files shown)


multi/stm32l4-multi/tty.c
```diff
@@ -42,6 +43,10 @@
 #define THREAD_STACKSZ 768
 #define THREAD_PRIO 1
 
+#if TTY_CNT != 0 && !BUILTIN_DUMMYFS
+#error "rootfs is needed when tty is enabled"
+#endif
```
agkaminski, line 48: rootfs is not optional - being glued to the multidriver is. So if !BUILTIN_DUMMYFS we should rather wait for rootfs to register

multi/stm32l4-multi/stm32l4-multi.c
```diff
@@ -44,6 +44,9 @@ struct {
 	char stack[THREADS_NO - 1][STACKSZ] __attribute__ ((aligned(8)));
 
 	unsigned int port;
+#if CONSOLE_IS_TTY
+	unsigned int ttyConsolePort;
+#endif
```
agkaminski, line 49: We don't need static storage for it in this compilation unit.

multi/stm32l4-multi/stm32l4-multi.c
```diff
@@ -240,14 +243,22 @@ int main(void)
...
 
-#if !CONSOLE_IS_TTY
-	/* called before init functions to redirect logs here */
+#if CONSOLE_IS_TTY
+	portCreate(&common.ttyConsolePort);
+#endif
 	portCreate(&common.port);
+
+#if BUILTIN_DUMMYFS
+	fs_init();
```
agkaminski, line 252: #else wait for rootfs to register?

multi/stm32l4-multi/tty.c
anglov: Ok, so I misunderstood the idea. If BUILTIN_DUMMYFS is not enabled, then we should assume that someone else (before multidrv can be operational) will create it?

agkaminski: We need to wait for rootfs to register

anglov: Done

multi/stm32l4-multi/stm32l4-multi.c
anglov: I thinking if it's valid. When you want to have rootfs on some real memory, then you can need multi to operate with it, but multi will not start until rootfs is not established… 

But as it's theoretical, I change it as you said.

agkaminski (approved):

PR #180, opened 2022-02-21
stm32l4-multi: makes: move libmulti external interface to separate dir

nalajcie:
## Description
Straighten up external interface.

Coincidentally - test for https://github.com/phoenix-rtos/phoenix-rtos-build/pull/102

## Motivation and Context
JIRA: CI-59

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work: https://github.com/phoenix-rtos/phoenix-rtos-build/pull/102


agkaminski:
Couldn't it be "stm32l4-multi/libmulti/include/header.h" and installed into "<module name>/header.h"? This additional directory inside of include directory look imho bad - most components will need to have include directory containing only one directory named after the module. This would be best to be solved in the makefile itself

nalajcie:
I don't like the deep dir structure, so I'm not enforcing this change. Maybe the lib component should be moved out of the folder with binary (or they should be side-by-side)?

Not done in this PR, but some ideas for the future (i don't know the full scope of needed changes):
- maybe we could rename files to `libmulti/uart.h` instead of `libmulti/libuart.h`?
- if `libmulti` interface is common across various targets, it should be moved from here, if it's target-specific - the target name needs to be in the path (eg. `libmulti-stm32l4/uart.h`) to avoid name clashing with future components
- include'ing "../common.h" should probably be resolved in better way

agkaminski:
- If we put libmulti's headers into one directory it would be best to rename includes, yes, no need for lib prefix this way. 
- Interface is not really made to be common, so libmulti-stm32l4x6 is a good idea.
- imho nothing wrong with include "../common.h"

nalajcie:
> Couldn't it be "stm32l4-multi/libmulti/include/header.h" and installed into "{module name]/header.h"? This additional directory inside of include directory look imho bad - most components will need to have include directory containing only one directory named after the module. This would be best to be solved in the makefile itself

I don't know if adding "module name" in Makefile would be user friendly, but we can also try that. There is one case when this would fail - if external API consists of cross-refeneced include files:

in `include/f1.h` -> `#include <module_name/f2.h>` would include `PREFIX_H/module_name/f2.h` instead of local `include/f2.h` file. The current solution would include `include/module_name/f2.h` when compiling the component.

Personally I would prefer to have option to do either:
1.  `include/libmulti.h` -> one-file interface to be used as `#include <libmulti.h>`
2. `include/libmulti/uart.h` ... -> multi-file interface to be used as `#include <libmulti/uart.h>`

I think most of our components would be of type (1) so this currently is the easiest to achieve :).

nalajcie:
> * If we put libmulti's headers into one directory it would be best to rename includes, yes, no need for lib prefix this way.
> * Interface is not really made to be common, so libmulti-stm32l4x6 is a good idea.

Just FYI, currently `libmulti` headers are also installed in `libmulti` subdir, this PR changes only how it's described in Makefiles. I didn't want to introduce filename changes as I don't know all of the use cases (external to phoenix-rtos-project) to do it consistently.

agkaminski:
Ok, so your approach seems to be most straightforward - what you put in "include" you'll see installed. Point good enough for me to accept ugly path.

agkaminski (approved):
@anglov what're your thoughts on this matter?

anglov (approved):
Generally:
- I think that putting files in exact include dir (<PATH>/module/header.h) when you include it using it with `#include <module/header.h>` is required for most IDE to properly resolve headers. - I also think that straightforward rule that everything in include goes into include_dir is good and logical. It avoids additional confusion.
- Maybe repository structure should be revised to have more global include dir (per module maybe). Or maybe just move libmulti to upper dir as libstm32l4-multi?
- I think this solution is good enough.

PR #190, opened 2022-04-07
usb/ehci: Fix qtds management

MaciejPurski:
Save qtd physical address to avoid calling va2pa() every time.
Move the queue forward, when it gets stuck due to hardware racing
with ehci_enqueue() function.

JIRA: DTR-227

## Motivation and Context

It fixes an issue, when new transfers are being enqueued constantly and they tend to race with the hardware.
When executing transfers, ehci fetches first qtd from the queue by copying it to an "overlay area" in a Queue Head structure. It then performs a transfers and afterwards saves the modifications back in the original qtd and follows the `next` pointer from the overlay area. If we modify a qtd that is now being processed in the overlay area, the controller won't see the modifications. If the modification is based on updating the `next` pointer it causes the queue to get stucked. This commit fixes this situation.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imx6ull).

## Checklist:

## Special treatment

### DIFF
--- a/usb/ehci/ehci.c
+++ b/usb/ehci/ehci.c
@@ -66,24 +66,28 @@ static void ehci_stopAsync(hcd_t *hcd)
 
 static void ehci_qtdLink(ehci_qtd_t *prev, ehci_qtd_t *next)
 {
-	prev->hw->next = QTD_PTR(next);
+	prev->hw->next = next->paddr;
 	ehci_memDmb();
 }
 
 
-static void ehci_enqueue(ehci_qh_t *qh, ehci_qtd_t *first, ehci_qtd_t *last)
+static void ehci_enqueue(hcd_t *hcd, ehci_qh_t *qh, ehci_qtd_t *first, ehci_qtd_t *last)
 {
+	ehci_t *ehci = (ehci_t *)hcd->priv;
+
+	mutexLock(ehci->asyncLock);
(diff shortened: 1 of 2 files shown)


usb/ehci/ehci.c
```diff
@@ -66,24 +66,28 @@ static void ehci_stopAsync(hcd_t *hcd)
...
-	prev->hw->next = QTD_PTR(next);
+	prev->hw->next = next->paddr;
 	ehci_memDmb();
 }
 
 
-static void ehci_enqueue(ehci_qh_t *qh, ehci_qtd_t *first, ehci_qtd_t *last)
+static void ehci_enqueue(hcd_t *hcd, ehci_qh_t *qh, ehci_qtd_t *first, ehci_qtd_t *last)
 {
+	ehci_t *ehci = (ehci_t *)hcd->priv;
```
agkaminski, line 76: This cast shouldn't really be necessary - you can use forward declaration to be able to hide this struct implementation in public header, if that's the reason

agkaminski (approved):
One small remark, not a merge blocker imho

usb/ehci/ehci.c
MaciejPurski: The reason for it is not hiding implementation details, but to allow a kind of "polymorphism", so that there are different host controllers, which keep their context in this pointer. At the moment we only have ehci.

PR #191, opened 2022-04-08
imu: add i2c imu server

niewim19:
JIRA: PP-6

## Description

 - i2c server for imu and magnetometer
 - using i2c library for imx6ull
 - adding regWrite() function to that i2c library
 - i2c addresses as defines in family specific .c files

## Motivation and Context

 - imu server shall be capable of quickly adding simmilar devices (of the same family) 
 - imu server shall be capable of quickly adding new family devices 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imx6ull

## Checklist:

## Special treatment


imu/main.c
```diff
@@ -0,0 +1,155 @@
...
+#include <unistd.h>
+#include <sys/time.h>
+
+#include <sys/msg.h>
+#include <posix/utils.h>
+#include <sys/threads.h>
+
+#include <communication.h>
+
+typedef struct {
```
nalajcie, line 27: no need for typedef

imu/communication.h
```diff
@@ -0,0 +1,52 @@
...
+
+typedef struct {
+	enum dev_types type;
+	enum dev_families family;
+	int model;
+	char name[16];
+	uint8_t addr;
+	uint8_t whoami_addr;
+	uint8_t whoami_val;
+} i2c_dev_t;
```
nalajcie, line 35: please leve the i2c_* namespace to i2c libs

imu/communication.h
```diff
@@ -0,0 +1,52 @@
...
+enum dev_types { type_imu, type_magmeter };
+
+enum dev_families { icm20xxx, ak099xx, unknown };
+
+typedef struct {
+	enum dev_types type;
+	enum dev_families family;
+	int model;
+	char name[16];
+	uint8_t addr;
```
nalajcie, line 32: dev_addr?

imu/icm20xxx.h
```diff
@@ -0,0 +1,45 @@
...
+#define _PHOENIX_I2C_ICM20XXX_H
+
+#include <stdint.h>
+#include <communication.h>
+
+/* size of all data accessible from icm20xxx */
+#define ICM20XXX_DATA_ALL_SIZE 14
+
+/* all supported members of icm20xxx family */
+enum icm_family { icm20948 };
```
nalajcie, line 23: all of the below seems like internal stuff (apart from init/get all data) and should be moved to .c file (applies everywhere else also)

imu/communication.c
```diff
@@ -0,0 +1,112 @@
...
+const i2c_dev_t *checkDevices(enum dev_types devtype)
+{
+	int i;
+	uint8_t whoami = 0;
+
+	for (i = 0; i < sizeof(devicelist) / sizeof(i2c_dev_t); i++) {
+		if (devicelist[i].type != devtype) {
+			continue;
+		}
+		i2c_regRead(devicelist[i].addr, devicelist[i].whoami_addr, &whoami, 1);
```
nalajcie, line 59: check for result code, You don't know if timeout has happened or not, You're not resetting the whoami variable (which will be left unchanged in case of error)

imu/communication.c
```diff
@@ -0,0 +1,112 @@
...
+}
+
+
+int initialize_i2cbus(void)
+{
+	return i2c_init(2);
+}
+
+
+const i2c_dev_t *checkDevices(enum dev_types devtype)
```
nalajcie, line 50: the usual term for searching for available devices is `probe` (eg. in linux every driver has `probe` function which tries to find out if a device exists or not).

Do we need device probing in runtime? Or should it be done by specifying the expected devices by program params (like drivers in lwip) to be changed in the future to devicetree?

imu/icm20xxx.c
```diff
@@ -0,0 +1,216 @@
...
+int icm20xxx_getAllData(const i2c_dev_t *dev, float *buffer, uint8_t buflen)
+{
+	uint8_t databuf[ICM20XXX_DATA_ALL_SIZE];
+
+	if (buflen < (ICM20XXX_DATA_ALL_SIZE / sizeof(uint16_t))) {
+		return -ENOMEM;
+	}
+
+	/* read data from device */
+	if (i2c_regRead(dev->addr, REG_DATA_ALL, databuf, ICM20XXX_DATA_ALL_SIZE)) {
```
nalajcie, line 202: misra: add explicit comparison (don't use ints as bools)

imu/communication.h
```diff
@@ -0,0 +1,52 @@
...
+int sendbyte(uint8_t dev_addr, uint8_t reg_addr, uint8_t val);
+
+/* checks if given devname from icm20xxx family is connected */
+const i2c_dev_t *checkDevices(enum dev_types devtype);
+
+/* setups given device */
+int initDevice(const i2c_dev_t *dev);
+
+/* gets all available data from accel and gyro */
+int getAllData(const i2c_dev_t *dev, float *buffer, uint8_t buflen);
```
nalajcie, line 50: I'm not fond of messing low-level internal API (sendbyte) to be used by individual drivers with high-level API (to be used by server to access device instead of device-specific calls?). If they are to be in the single file - they need to be clearly separated.

Regarding the lower-layer API - I don't personally like writing all data single buffer (regarding of type, origin, etc.), because we're loosing the meaning of data and make it hard to use (but I don't know the use cases so whatever). Personally I would go for `struct with callbacks` as an API to the drivers (with specific callbacks being non-null if driver supports them (eg. `int (*getAccel) (float *x, float *y, float *z)` - or define struct with these fields?)

Not sure we need the comm manager - maybe imu and magnetometer should be separate processes? (right now we're limiting ourselfs to having IMU and magnetometer on the same i2c bus, would it always be the case?)

imu/communication.c
```diff
@@ -0,0 +1,112 @@
...
+
+int sendbyte(uint8_t dev_addr, uint8_t reg_addr, uint8_t val)
+{
+	return (i2c_regWrite(dev_addr, reg_addr, &val, 1) >= 0) ? 0 : -1;
+}
+
+
+int initialize_i2cbus(void)
+{
+	return i2c_init(2);
```
nalajcie, line 46: ugly hardcode, should probably driver cmdline parameter

imu/icm20xxx.c
```diff
@@ -0,0 +1,216 @@
...
+		fprintf(stderr, "Cannot reset and put in run\n");
+		return -EIO;
+	}
+	ret = 0;
+
+	/* Accellerometer and Gyro configs are in USER_BANK_2 */
+	ret += sendbyte(dev->addr, REG_BANK, VAL_BANK_2);
+	{
+		/* set accel sample rate divider */
+		ret += sendbyte(dev->addr, REG_ACCEL_SMPLRT_DIV_2, VAL_ACCEL_SMPLRT_DIV_2);
```
nalajcie, line 172: if the bank switching failed, we should probably not issue these commands (we might be changing other configs?)

imu/communication.c
```diff
@@ -0,0 +1,112 @@
...
+
+
+const i2c_dev_t devicelist[] = {
+	{ .type = type_imu, .family = icm20xxx, .model = icm20948, .name = "ICM20948", .addr = 0x68, .whoami_addr = 0x00, .whoami_val = 0xEA },
+	{ .type = type_imu, .family = icm20xxx, .model = icm20948, .name = "ICM20948_1", .addr = 0x69, .whoami_addr = 0x00, .whoami_val = 0xEA },
+	{ .type = type_magmeter, .family = ak099xx, .model = ak09916, .name = "AK09916", .addr = 0x0C, .whoami_addr = 0x01, .whoami_val = 0x09 }
+};
+
+
+int sendbyte(uint8_t dev_addr, uint8_t reg_addr, uint8_t val)
```
nalajcie, line 38: unfortunate name as You're actually sending 2 bytes at i2c bus. maybe write_reg8?
also module prefixes for non-static functions should probably be added here

nalajcie (changes requested):
pointed out some issues, not doing a really careful review as I don't know the full use-cases and the expected external API (but the current needs to change ;))

imu/main.c
```diff
@@ -0,0 +1,155 @@
...
+	dev.id = 0;
+	printf("%s: running on port %u\n", devname, port);
+	if (create_dev(&dev, devname) < 0) {
+		printf("%s: could not create device\n", devname);
+		return 4;
+	}
+
+	/* prepare and start data acquisition thread thread */
+	mutexCreate(&data_context.imu_data_lock);
+	mutexCreate(&data_context.mag_data_lock);
```
agkaminski, line 144: I don't see any pros of having two mutexes - you are really locking the communication channel, now device - and there's one communication channel

imu/main.c
```diff
@@ -0,0 +1,155 @@
...
+				break;
+		}
+
+		msgRespond(port, &msg, rid);
+	}
+}
+
+int main(int argc, char **argv)
+{
+	int stacksz = 1024;
```
kemonats, line 102:
```suggestion
	const size_t stacksz = 1024;
```

imu/main.c
```diff
@@ -0,0 +1,155 @@
...
+	printf("%s: running on port %u\n", devname, port);
+	if (create_dev(&dev, devname) < 0) {
+		printf("%s: could not create device\n", devname);
+		return 4;
+	}
+
+	/* prepare and start data acquisition thread thread */
+	mutexCreate(&data_context.imu_data_lock);
+	mutexCreate(&data_context.mag_data_lock);
+	if ((stack = (char *)malloc(stacksz)) == NULL) {
```
kemonats, line 145: casting `malloc()` is not recommended.
```suggestion
	if ((stack = malloc(stacksz)) == NULL) {
```

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -243,6 +243,22 @@ int i2c_busWrite(uint8_t dev_addr, const uint8_t *data, uint32_t len)
...
+	uint8_t d[2];
+
+	/* TODO: support more than 1 byte long writes */
+	if (len == 1) {
+		d[0] = reg_addr;
+		d[1] = data[0];
+		return i2c_busWrite(dev_addr, d, 2);
+	}
+	else {
+		return -E2BIG;
```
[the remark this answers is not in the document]
niewim19: Removed sendbyte() wrapper and embeded simple one byte write to specified register as `i2c_regWrite()` into i2c driver

imu/Makefile
```diff
@@ -0,0 +1,15 @@
...
+# Copyright 2021 Phoenix Systems
+#
+#
+
+# by default use i2c message implementation (can be overridden to use custom target i2c lib directly, eg. `make I2C_LIB=libimx6ull-i2c temp-nct75`)
+I2C_LIB ?= libimx6ull-i2c
+
+NAME := imu
+LOCAL_SRCS := main.c communication.c icm20xxx.c ak099xx.c
+LOCAL_HEADERS := communication.h icm20xxx.h ak099xx.h
```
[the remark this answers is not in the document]
niewim19: As you mentioned there are no good candidates to be installed so currently `include/imu` i'll skip `include/imu` addition. Changed all header usage to be local `<...>` -> `"..."`

imu/main.c
```diff
@@ -0,0 +1,155 @@
...
+
+data_context_t data_context;
+
+void daq(void *arg)
+{
+	int i = 0;
+	for (;;) {
+		for (i = 0; i < 10; i++) {
+			mutexLock(data_context.imu_data_lock);
+			getAllData(data_context.imu, data_context.imu_data, 7);
```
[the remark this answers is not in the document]
niewim19: Used sizeof.

Currently I`ll leave it without an error check. If it is an occasional error in communication we want to skip wrong data with last good data.

imu/main.c
```diff
@@ -0,0 +1,155 @@
...
+	handle_t mag_data_lock;
+} data_context_t;
+
+data_context_t data_context;
+
+void daq(void *arg)
+{
+	int i = 0;
+	for (;;) {
+		for (i = 0; i < 10; i++) {
```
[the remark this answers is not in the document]
niewim19: Yes. This is a quick solution as usually magnetometers have longer relaxation time between possible measurements (lower sampling frequency). Current implementation is temporary.

Sampling rate call/message will be added when power consumption will come to the foreground.

imu/icm20xxx.h
niewim19: Yes, thats true. Moved to .c file. Thanks!

imu/communication.c
niewim19: Thank you for the hint about `probe`. Function name changed.

About runtime probing - this is desired feature of the driver. Final use case may be used with different imu sensors interchangeably. Thats why runtime probe is convenient.

imu/communication.c
niewim19: Added commandline setup + default value.

imu/communication.c
niewim19: Removed after `regWrite()` changes.

imu/ak099xx.c
```diff
@@ -0,0 +1,96 @@
...
+		return -EAGAIN;
+	}
+	if (ret < 0) {
+		return -EIO;
+	}
+
+	/* read data */
+	ret += i2c_regRead(dev->addr, REG_DATA_ALL, buf, AK0990XX_DATA_ALL_SIZE);
+	/* read hall overflow */
+	ret += i2c_regRead(dev->addr, REG_ST2, &status, 1);
```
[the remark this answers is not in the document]
niewim19: Went for logic `or` solution. Thanks!

imu/main.c
niewim19: At first I was skeptical of two mutexes, but as long as there is no value specific API (only generic getAllData) then you are right - two mutexes are redundant. 

Substituted both with one called `data_lock`. Thanks!

imu/ak099xx.c
```diff
@@ -0,0 +1,96 @@
...
+		return -ENOMEM;
+	}
+
+	if (ak099xx_readAllData(dev, databuf, AK0990XX_DATA_ALL_SIZE) < 0) {
+		return -EIO;
+	}
+
+	buffer[0] = translateMag((databuf[0] << 8) | (databuf[1] & 0xFF));
+	buffer[1] = translateMag((databuf[2] << 8) | (databuf[3] & 0xFF));
+	buffer[2] = translateMag((databuf[4] << 8) | (databuf[5] & 0xFF));
```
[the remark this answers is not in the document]
niewim19: Neat simplification. Added, thanks!

imu/communication.c
```diff
@@ -0,0 +1,112 @@
...
+	return i2c_init(2);
+}
+
+
+const i2c_dev_t *checkDevices(enum dev_types devtype)
+{
+	int i;
+	uint8_t whoami = 0;
+
+	for (i = 0; i < sizeof(devicelist) / sizeof(i2c_dev_t); i++) {
```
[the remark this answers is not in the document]
niewim19: In a situation where devicelist is empty wouldn't such `sizeof()` be a memory read out of bound?

imu/icm20xxx.c
```diff
@@ -0,0 +1,216 @@
...
+	switch (imu_common.scale_acc) {
+		case g2:
+			return fval * 2.F;
+		case g4:
+			return fval * 4.F;
+		case g16:
+			return fval * 16.F;
+		default:
+			return fval * 8.F;
+	}
```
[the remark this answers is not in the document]
niewim19: Yes. This is better. Implemented.

imu/icm20xxx.c
```diff
@@ -0,0 +1,216 @@
...
+
+	/* change accell values to floats */
+	buffer[0] = translateAccel((databuf[0] << 8) | (databuf[1] & 0xFF));  /* acc_x */
+	buffer[1] = translateAccel((databuf[2] << 8) | (databuf[3] & 0xFF));  /* acc_y */
+	buffer[2] = translateAccel((databuf[4] << 8) | (databuf[5] & 0xFF));  /* acc_z */
+	buffer[3] = translateGyr((databuf[6] << 8) | (databuf[7] & 0xFF));    /* gyr_x */
+	buffer[4] = translateGyr((databuf[8] << 8) | (databuf[9] & 0xFF));    /* gyr_y */
+	buffer[5] = translateGyr((databuf[10] << 8) | (databuf[11] & 0xFF));  /* gyr_z */
+	buffer[6] = translateTemp((databuf[12] << 8) | (databuf[13] & 0xFF)); /* temp */
```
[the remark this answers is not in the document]
niewim19: Neat simplification. Added, thanks!

imu/main.c
niewim19: Corrected, thanks!

i2c/imx6ull/libimx6ull-i2c.c
```diff
@@ -333,7 +349,6 @@ int i2c_regRead(uint8_t dev_addr, uint8_t reg_addr, uint8_t *data_out, uint32_t
 int i2c_init(unsigned int dev_no)
 {
 	i2c.dev_no = dev_no;
```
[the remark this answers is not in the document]
niewim19: Reverted, thanks!

niewim19:
Implemented most of the notes from kemonats, agkaminski and nalajcie reviews. Things that still need to be addressed are:
 - api cleanup and/or expansion
 - check return code in device probing in communications.h

imu/communication.c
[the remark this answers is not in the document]
kemonats: If is empty `devlist` e.g.
```
const i2c_dev_t devlist[] = {};
```
then
```
sizeof(devlist) / sizeof(devlist[0]) 
```
the compiler will calculate as `0`.
The advantage of `sizeof(devlist) / sizeof(devlist[0])` over `sizeof(devlist) / sizeof(i2c_dev_t)` is that when you change the type of an array, you don't have to change the expression that calculates the size of that array.

niewim19:
Closing this PR as introduced changes are stale.

PR #194, opened 2022-04-13
imx6ull-flash: reduce number of va2pa() calls while creating DMA descriptor chains

Maxez:
## Description

Performance optimization. DMA descriptors and flash commands buffer physical address can be calculated using an offset from the beginning of the DMA buffer page (`flashdrv_dmaBuff_t`) which reduces number of `va2pa()` calls while creating a DMA descriptor chain.

## Motivation and Context

[JIRA: DTR-169](https://jira.phoenix-rtos.com/browse/DTR-169)

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a7-imx6ull.

## Checklist:

## Special treatment

### DIFF
--- a/storage/imx6ull-flash/flashdrv.c
+++ b/storage/imx6ull-flash/flashdrv.c
@@ -63,7 +63,7 @@ typedef struct {
 	uint16_t bufsz;
 	uint32_t buffer;
 	uint32_t pio[];
-} dma_t;
+} flashdrv_dmaDesc_t;
 
 
 enum { bch_ctrl = 0, bch_ctrl_set, bch_ctrl_clr, bch_ctrl_tog, bch_status0,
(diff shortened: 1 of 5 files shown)


storage/imx6ull-flash/flashdrv.c
```diff
@@ -187,11 +187,12 @@ static const flashdrv_command_t commands[flash_num_commands] = {
 };
 
 
-typedef struct _flashdrv_dma_t {
-	dma_t *last;
-	dma_t *first;
+typedef struct _flashdrv_dmaBuff_t {
```
agkaminski, line 190: Ok, do we need this struct definition? I guess not, so it should be anonymous (typedef only)

agkaminski (commented):
One small remark, otherwise LGTM

storage/imx6ull-flash/flashdrv.c
Maxez: We need the struct definition because it's used in the header (see [link](https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/b7ea02f1755c7f65656398cc351fc590ec8dd231/storage/imx6ull-flash/imx6ull-flashdrv.h#L20)).
The don't want the struct content to be visible, just the type.

PR #198, opened 2022-04-21
zynq7000-flash: apply libstorage interface to flashdrv

HBuczynski:
## Description

- change default sector size for Micron n25q128 to run jffs2,
- add MTD and block device interface for flash driver,
- update tests after changes in flash driver.

More detailed information is provided in commit message.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (zedboard, zynq7000-qemu).

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-build/pull/109).

### DIFF
--- a/storage/zynq7000-flash/flashdrv.c
+++ b/storage/zynq7000-flash/flashdrv.c
@@ -11,30 +11,42 @@
  * %LICENSE%
  */
 
-#include "qspi.h"
+#include "flashdrv.h"
 #include "flashcfg.h"
+#include "qspi.h"
 
 #include <stdlib.h>
 #include <errno.h>
 #include <string.h>
 #include <sys/time.h>
+#include <sys/threads.h>
 
-#define TIMEOUT_CMD_MS 0x05
+/* Value determined empirically, contains time for communication via qspi and system calls invocations. */
+#define TIMEOUT_CMD_MS 1000
 #define MAX_SIZE_CMD   0x20
 
+#define MTD_DEFAULT_ERASESZ 0x10000
 
-struct {
+
+typedef struct {
+	/* Data caching in sector's buffer */
+	uint8_t *buff;
+	uint32_t buffPos;
+	uint32_t sectID;
+
 	/* Buffers for command transactions */
 	uint8_t cmdRx[MAX_SIZE_CMD];
 	uint8_t cmdTx[MAX_SIZE_CMD];
 
-	/* Data caching in sector's buffer */
-	uint8_t *buff;
-	uint32_t buffPos;
-	uint32_t regID;
-	uint32_t sectID;
+	handle_t lock;
+} flash_reg_t;
 
-	flash_info_t info;
+
+struct {
+	flash_reg_t *regs;
+
+	flash_info_t info;     /* CFI structure for NOR flash memory */
+	unsigned int initRegs; /* Number of initialized regions */
 } fdrv_common;
 
 
@@ -49,47 +61,27 @@ static inline time_t flashdrv_timeMsGet(void)
 }
 
 
-static int flashdrv_regionFind(addr_t offs, uint32_t *id)
-{
-	int i;
-	size_t sz;
-	addr_t start = 0;
-	const flash_cfi_t *cfi = &fdrv_common.info.cfi;
-
-	for (i = 0; i < cfi->regsCount; ++i) {
-		sz = CFI_SIZE_REGION(cfi->regs[i].size, cfi->regs[i].count);
-		if (offs >= start && offs < (start + sz)) {
-			*id = i;
-			return EOK;
-		}
-
-		start = sz;
-	}
-
-	return -EINVAL;
-}
-
-
 static size_t flashdrv_regStart(int id)
 {
 	int i;
 	size_t regStart = 0;
 	const flash_cfi_t *cfi = &fdrv_common.info.cfi;
 
-	for (i = 0; i < id; ++i)
+	for (i = 0; i < id; ++i) {
 		regStart += CFI_SIZE_REGION(cfi->regs[i].size, cfi->regs[i].count);
+	}
 
 	return regStart;
 }
 
 
-static inline void flashdrv_initCmdBuff(const flash_cmd_t *cmd, addr_t offs)
+static inline void _flashdrv_initCmdBuff(unsigned int id, const flash_cmd_t *cmd, addr_t offs)
 {
-	memset(fdrv_common.cmdTx, 0, cmd->size);
-	fdrv_common.cmdTx[0] = cmd->opCode;
-	fdrv_common.cmdTx[1] = (offs >> 16) & 0xff;
-	fdrv_common.cmdTx[2] = (offs >> 8) & 0xff;
-	fdrv_common.cmdTx[3] = offs & 0xff;
+	memset(fdrv_common.regs[id].cmdTx, 0, cmd->size);
+	fdrv_common.regs[id].cmdTx[0] = cmd->opCode;
+	fdrv_common.regs[id].cmdTx[1] = (offs >> 16) & 0xff;
+	fdrv_common.regs[id].cmdTx[2] = (offs >> 8) & 0xff;
+	fdrv_common.regs[id].cmdTx[3] = offs & 0xff;
 }
 
 
@@ -99,6 +91,8 @@ static int flashdrv_cfiRead(flash_cfi_t *cfi)
 {
 	ssize_t res;
 	size_t cmdSz, dataSz;
+	static uint8_t cmdRx[MAX_SIZE_CMD];
+	static uint8_t cmdTx[MAX_SIZE_CMD];
 	/* RDID instruction is a generic for each flash memory */
 	flash_cmd_t cmd;
 
@@ -109,18 +103,18 @@ static int flashdrv_cfiRead(flash_cfi_t *cfi)
 	/* Size of the data which is received during cmd transfer */
 	dataSz = cmdSz - cmd.size;
 
-	memset(fdrv_common.cmdTx, 0, cmdSz);
-	fdrv_common.cmdTx[0] = cmd.opCode;
+	memset(cmdTx, 0, cmdSz);
+	cmdTx[0] = cmd.opCode;
 
 	qspi_start();
-	res = qspi_transfer(fdrv_common.cmdTx, fdrv_common.cmdRx, cmdSz, TIMEOUT_CMD_MS);
+	res = qspi_transfer(cmdTx, cmdRx, cmdSz, TIMEOUT_CMD_MS);
 	if (res < 0) {
 		qspi_stop();
 		return res;
 	}
 
 	/* Copy data received after dummy bytes */
-	memcpy(cfi, (fdrv_common.cmdRx + cmd.size), dataSz);
+	memcpy(cfi, (cmdRx + cmd.size), dataSz);
 	res = qspi_transfer(NULL, (uint8_t *)cfi + dataSz, sizeof(flash_cfi_t) - dataSz, 5 * TIMEOUT_CMD_MS);
 	qspi_stop();
 
@@ -128,7 +122,7 @@ static int flashdrv_cfiRead(flash_cfi_t *cfi)
 }
 
 
-static int flashdrv_statusRegGet(unsigned int *val)
+static int _flashdrv_statusRegGet(unsigned int id, unsigned int *val)
 {
 	ssize_t res;
 	size_t cmdSz;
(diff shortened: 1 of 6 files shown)


storage/zynq7000-flash/flashdrv.c
```diff
@@ -11,30 +11,42 @@
...
 	/* Data caching in sector's buffer */
 	uint8_t *buff;
 	uint32_t buffPos;
-	uint32_t regID;
 	uint32_t sectID;
 
-	flash_info_t info;
+	/* Buffers for command transactions */
+	uint8_t cmdRx[MAX_SIZE_CMD];
+	uint8_t cmdTx[MAX_SIZE_CMD];
```
agkaminski, line 39: You always memset this in the functions - it's not huge (32 bytes) so imho it would be easier and cleaner to do
`uint8_t cmdTx[SIZE_WE_REALLY_NEED] = { 0 };`
inside the function (local variable) instead of doing `hal_memset`

Just a thought, I might be wrong

storage/zynq7000-flash/flashdrv.c
```diff
@@ -49,47 +61,27 @@ static inline time_t flashdrv_timeMsGet(void)
...
 
-static inline void flashdrv_initCmdBuff(const flash_cmd_t *cmd, addr_t offs)
+static inline void _flashdrv_initCmdBuff(unsigned int id, const flash_cmd_t *cmd, addr_t offs)
 {
-	memset(fdrv_common.cmdTx, 0, cmd->size);
-	fdrv_common.cmdTx[0] = cmd->opCode;
-	fdrv_common.cmdTx[1] = (offs >> 16) & 0xff;
-	fdrv_common.cmdTx[2] = (offs >> 8) & 0xff;
-	fdrv_common.cmdTx[3] = offs & 0xff;
+	memset(fdrv_common.regs[id].cmdTx, 0, cmd->size);
```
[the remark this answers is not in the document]
agkaminski: If we declare size we really need, we may even drop this `= { 0 };` part - we pass `cmdSize` to `qpi_transfer` anyway, so no need for bigger array.

storage/zynq7000-flash/flashdrv.c
```diff
@@ -398,126 +357,424 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
-	uint32_t regID = 0, sectID = 0;
-	size_t sectSz, regStart;
+	unsigned int i, pgNb;
 
+	size_t pageSz, sectSz;
 	const flash_cfi_t *cfi = &fdrv_common.info.cfi;
 
-	if (len == 0)
-		return 0;
+	if (fdrv_common.regs[id].buffPos == 0) {
```
agkaminski, line 372: Hmmm, why is this called buffPos? Why position zero means that there's nothing to do?

storage/zynq7000-flash/flashdrv.c
```diff
@@ -398,126 +357,424 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
+	}
+
+	fdrv_common.regs[id].sectID = (uint32_t)-1;
+	fdrv_common.regs[id].buffPos = 0;
+
+	return EOK;
+}
 
-	if (buff == NULL || (offs + len) > CFI_SIZE_FLASH(cfi->chipSize))
+static ssize_t flashdrv_blkRead(struct _storage_t *strg, off_t start, void *data, size_t size)
```
agkaminski, line 396:
```suggestion

static ssize_t flashdrv_blkRead(struct _storage_t *strg, off_t start, void *data, size_t size)
```

storage/zynq7000-flash/flashdrv.c
```diff
@@ -398,126 +357,424 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
+	ssize_t res;
+
+	if ((start + size) > strg->size) {
 		return -EINVAL;
+	}
 
-	while (saveSz < len) {
-		offs += chunkSz;
+	mutexLock(fdrv_common.regs[strg->dev->id].lock);
+	res = _flashdrv_read(strg->dev->id, start + strg->start, data, size, TIMEOUT_CMD_MS * size);
```
agkaminski, line 405: IMHO it would be better to pass timeout per cmd (or even don't pass timeout at all), _flashdrv_read knows about TIMEOUT_CMD_MS and size and can do this operation by itself

storage/zynq7000-flash/flashdrv.c
```diff
@@ -398,126 +357,424 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
+		res = _flashdrv_pageProgram(strg->dev->id, strg->start + offs, data, chunkSz);
+		mutexUnlock(fdrv_common.regs[strg->dev->id].lock);
+		if (res < 0) {
+			return res;
+		}
+
+		tempSz = res;
+	}
+
+	mutexLock(fdrv_common.regs[strg->dev->id].lock);
```
agkaminski, line 570: Less pretty, but we don't give lock away between operations 
```suggestion
	if (chunkSz != 0) {
		chunkSz = len < (strg->dev->mtd->writeBuffsz - chunkSz) ? len : (strg->dev->mtd->writeBuffsz - chunkSz);
		mutexLock(fdrv_common.regs[strg->dev->id].lock);
		res = _flashdrv_pageProgram(strg->dev->id, strg->start + offs, data, chunkSz);
		if (res < 0) {
			mutexUnlock(fdrv_common.regs[strg->dev->id].lock);
			return res;
		}

		tempSz = res;
	}
	else {
		mutexLock(fdrv_common.regs[strg->dev->id].lock);
	}
```

agkaminski: Or simpler solution
```suggestion
	mutexLock(fdrv_common.regs[strg->dev->id].lock);
	if (chunkSz != 0) {
		chunkSz = len < (strg->dev->mtd->writeBuffsz - chunkSz) ? len : (strg->dev->mtd->writeBuffsz - chunkSz);
		res = _flashdrv_pageProgram(strg->dev->id, strg->start + offs, data, chunkSz);
		if (res < 0) {
			mutexUnlock(fdrv_common.regs[strg->dev->id].lock);
			return res;
		}

		tempSz = res;
	}
```

storage/zynq7000-flash/flashdrv.c
```diff
@@ -398,126 +357,424 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
+
+		tempSz += res;
+	}
+	mutexUnlock(fdrv_common.regs[strg->dev->id].lock);
+
+	return tempSz;
+}
+
+
+const static storage_mtdops_t mtdOps = {
```
agkaminski, line 587: Maybe put this inside `flashdrv_devInit` funtion?

storage/zynq7000-flash/flashdrv.c
```diff
@@ -398,126 +357,424 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
+int flashdrv_done(storage_t *strg)
+{
+	int i, res;
+
+	if (strg == NULL) {
+		return -EINVAL;
+	}
+
+	res = flashdrv_blkSync(strg);
+
```
agkaminski, line 624: Isn't some lock needed here?

storage/zynq7000-flash/flashdrv.c
```diff
@@ -398,126 +357,424 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
+		free(strg->dev);
+		free(strg->dev->mtd);
 		return -ENOMEM;
+	}
+	strg->dev->blk->ops = &blkOps;
 
-	return EOK;
+	fdrv_common.initRegs++;
+
+	return (info->cfi.regsCount - fdrv_common.initRegs);
```
agkaminski, line 734: Isn't some lock needed here?

storage/zynq7000-flash/flashdrv.c
HBuczynski: The `_flashdrv_blkSync` needs a lock before usage. However, the `flashdrv_blkSync` is thread safe.  I keep convention commonly used in kernel.

storage/zynq7000-flash/flashdrv.c
HBuczynski: Access to `flashdrv_init` shouldn't be done by many threads, this function has to be invoked only once.
This is the main reason why I decided to not add additional lock for `fdrv_common` fields, in runtime, there are not changed.

storage/zynq7000-flash/flashdrv.c
HBuczynski: Hmm, I don't see any change in Your suggestion.

agkaminski: Added empty line between functions

storage/zynq7000-flash/flashdrv.c
agkaminski: I don't mean flashdrv_blkSync (it has locks, it's ok), but the stuff you're doing after it. So maybe lock all and use _flashdrv_blkSync instead?

storage/zynq7000-flash/flashdrv.c
HBuczynski: Thanks for suggestion. I have removed the timeout argument from `_flashdrv_read`.

storage/zynq7000-flash/flashdrv.c
HBuczynski: Thx, added.

storage/zynq7000-flash/flashdrv.c
```diff
@@ -11,30 +11,42 @@
...
 
 #include <stdlib.h>
 #include <errno.h>
 #include <string.h>
 #include <sys/time.h>
+#include <sys/threads.h>
 
-#define TIMEOUT_CMD_MS 0x05
+/* Value determined empirically, contains time for communication via qspi and system calls invocations. */
+#define TIMEOUT_CMD_MS 0x50
```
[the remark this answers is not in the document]
HBuczynski: Ok, I have increased the value up to 1000 ms.

storage/zynq7000-flash/flashdrv.c
HBuczynski: `buffPos` define the current position in cache buffer.

It means that there is not any data in a buffer. For example, if `sync` function has been invoked two times, for the second time the buffer should not be saved to the flash memory.

storage/zynq7000-flash/flashdrv.c
HBuczynski: The command size depends on the flash memory. So the size is not always known a priori, some calculations are needed. It would be difficult to be compliant with our convention of declaring data at the beginning of the function.

storage/zynq7000-flash/flashdrv.c
HBuczynski: Ok, I have changed to `_flashdrv_blkSync`

storage/zynq7000-flash/flashdrv.c
```diff
@@ -383,141 +342,439 @@ ssize_t flashdrv_read(addr_t offs, void *buff, size_t len, time_t timeout)
...
+	resourceDestroy(fdrv_common.regs[strg->dev->id].lock);
+	free(strg->dev->mtd);
+	free(strg->dev->blk);
+	free(strg->dev);
+
+	for (i = 0; i < fdrv_common.info.cfi.regsCount; ++i) {
+		if (fdrv_common.regs[i].buff != NULL) {
+			return res;
+		}
+	}
```
agkaminski, line 635: Doesn't this need to be under lock too?

HBuczynski: `buff` is changed only in `flashdrv_done` so in this `for` there is no chance that other thread will change `buff`

agkaminski: Ok, but above you agreed to protect buff under lock, but here is not needed? It's either needed in both cases or not needed at all. Can we have two concurrent flashdrv_done operations?

agkaminski (approved):

storage/zynq7000-flash/flashdrv.c
HBuczynski: Yeah, my mistake. We can assume that there will not be concurrent `flashdrv_done` operations. Restored previous code.

PR #206, opened 2022-05-04
Ziemleszcz/uart

ziemleszcz:
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imx6ull).

## Checklist:

## Special treatment

### DIFF
--- a/tty/imx6ull-uart/imx6ull-uart.c
+++ b/tty/imx6ull-uart/imx6ull-uart.c
@@ -91,6 +91,31 @@ uart_t uart = { 0 };
 
 #define BUFSIZE 4096
 
+/* register flags */
+#define USR1_RRDY      (1 << 9)
+#define USR1_TRDY      (1 << 13)
+#define USR1_FRAMERR   (1 << 10)
+#define USR1_PARITYERR (1 << 15)
+#define USR2_RDR       (1 << 0)
+#define USR2_ORE       (1 << 1)
+#define UTS_TXFULL     (1 << 4)
+#define UTS_SOFTRST    (1 << 0)
+#define UCR1_UARTEN    (1 << 0)
+#define UCR1_RRDYEN    (1 << 9)
+#define UCR1_TRDYEN    (1 << 13)
+#define UCR2_SRST      (1 << 0)
+#define UCR2_RXEN      (1 << 1)
+#define UCR2_TXEN      (1 << 2)
+#define UCR2_WS        (1 << 5)
+#define UCR2_IRTS      (1 << 14)
+#define UCR3_RXDMUXSEL (1 << 2)
+#define UCR3_RI        (1 << 8)
+#define UCR3_DCD       (1 << 9)
+#define UCR3_DSR       (1 << 10)
+#define UCR3_FRAERREN  (1 << 11)
+#define UCR3_PARERREN  (1 << 12)
+#define UCR4_OREN      (1 << 1)
+
 void uart_thr(void *arg)
 {
 	uint32_t port = (uint32_t)arg;
@@ -409,12 +434,14 @@ int main(int argc, char **argv)
 		return 2;
 
 	set_clk(uart.dev_no);
-	*(uart.base + ucr2) &= ~0;
+
+	/* software reset */
+	*(uart.base + ucr2) = 0;
+	while ((*(uart.base + uts) & UTS_SOFTRST))
+		;
 
 	/* set correct daisy for rx input */
 	set_mux(uart.dev_no, use_rts_cts);
-
-	while (!(*(uart.base + ucr2) & 1));
 
 	if (mutexCreate(&uart.lock) != EOK)
 		return 2;
@@ -424,7 +451,6 @@ int main(int argc, char **argv)
 
 	interrupt(uart_intr_number[uart.dev_no - 1], uart_intr, NULL, uart.cond, &uart.inth);
 
-
 	/* set TX & RX FIFO watermark, DCE mode */
 	*(uart.base + ufcr) = (0x04 << 10) | (0 << 6) | (0x1);
 
@@ -432,16 +458,17 @@ int main(int argc, char **argv)
 	*(uart.base + ufcr) &= ~(0b111 << 7);
 	*(uart.base + ufcr) |= 0b010 << 7;
 
-	/* enable uart and rx ready interrupt */
-	*(uart.base + ucr1) |= 0x0201;
-
-	/* soft reset, tx&rx enable, 8bit transmit */
-	*(uart.base + ucr2) = 0x4027;
+	/* ignore RTS pin, 8-bit transmit, RX enable, TX enable, soft reset */
+	*(uart.base + ucr2) = UCR2_IRTS | UCR2_WS | UCR2_RXEN | UCR2_TXEN | UCR2_SRST;
 
 	set_cflag(&uart, &uart.tty_common.term.c_cflag);
 	set_baudrate(&uart, baud);
 
-	*(uart.base + ucr3) = 0x704;
+	/* set muxed mode */
+	*(uart.base + ucr3) |= UCR3_RXDMUXSEL;
+
+	/* enable UART and receiver ready interrupt */
+	*(uart.base + ucr1) |= UCR1_UARTEN | UCR1_RRDYEN;
 
 	beginthread(uart_intrthr, 3, &stack0, 2048, NULL);
 	beginthread(uart_thr, 3, &stack, 2048, (void *)port);
--- a/tty/imx6ull-uart/imx6ull-uart.c
+++ b/tty/imx6ull-uart/imx6ull-uart.c
@@ -29,6 +29,7 @@
 #include <sys/stat.h>
 #include <sys/debug.h>
 #include <posix/utils.h>
+#include <fcntl.h>
 
 #include <libtty.h>
 #include <libklog.h>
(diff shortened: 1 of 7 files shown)


tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -145,48 +286,104 @@ void uart_thr(void *arg)
...
+	for (;;) {
+		do {
+			flags = __atomic_exchange_n(&uart.int_flags, 0, __ATOMIC_RELAXED);
+
+			if (flags & PARITY_ERR)
+				log_warn("parity error (%u)", uart.counters.parity);
+			if (flags & FRAME_ERR)
+				log_warn("frame error (%u)", uart.counters.frame);
+			if (flags & HW_OVERRUN_ERR)
+				log_warn("hw_overrun (%u)", uart.counters.hw_overrun);
```
nalajcie, line 368: not sure if printing more chars (possibly on UART) on overflow is the best thing to do ;).

Side note: MISRA says we always need to use `{ }`

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -301,20 +501,34 @@ void set_cflag(void* _uart, tcflag_t* cflag)
...
-static void signal_txready(void* _uart)
+
+static void signal_txready(void *_uart)
 {
-	uart_t* uartptr = (uart_t*) _uart;
+	/* disable TX interrupt to ensure that interrupt handler doesn't touch TX buffer */
+	*(uart.base + ucr1) &= ~UCR1_TRDYEN;
+
+	/* push as many characters as possible to HW FIFO at once */
+	while (!fifo_is_empty(uart.tty_common.tx_fifo)) {
```
nalajcie, line 511: use `libtty_txready` and `libtty_getchar`?

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -107,16 +247,17 @@ void uart_thr(void *arg)
 
 		switch (msg.type) {
 		case mtOpen:
-			// TODO: set PGID?
+			msg.o.io.err = uart_open(msg.i.openclose.flags);
```
nalajcie, line 250: shame I needed to read kernel source code to understand how it works (but that's the problem of our API)

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -77,21 +95,143 @@ typedef struct {
...
+{
+	/* TODO: set PGID? */
+	int id = flags_to_id(flags);
+
+	if (id <= 0)
+		return -EACCES;
+
+	mutexLock(uart.openclose_lock);
+
+	if (uart.busy & id) {
```
nalajcie, line 189: ok, side effect is that I can't open UART twice (not that I need to, just it's possible on linux)

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -77,21 +95,143 @@ typedef struct {
...
+
+/* interrupt flags */
+#define RX_DONE        (1 << 0)
+#define TX_DONE        (1 << 1)
+#define PARITY_ERR     (1 << 2)
+#define FRAME_ERR      (1 << 3)
+#define HW_OVERRUN_ERR (1 << 4)
+#define SW_OVERRUN_ERR (1 << 5)
+
+/* register flags */
```
nalajcie, line 138: IMHO more readable than magic numbers but adding @HBuczynski to reviewers for kernel team to approve the style :P.

nalajcie (commented):
Maybe we could use libtty external API (either change the semantics of existing functions or add some new non-{blocking,signaling} ones?)

tty/imx6ull-uart/imx6ull-uart.c
ziemleszcz: You mean two readers?

agkaminski: Imho there is only a point in restricting readers. I see no problem with having multiple writers or reader and writer at the same time

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -403,45 +655,59 @@ int main(int argc, char **argv)
 	if (portCreate(&port) != EOK)
 		return 2;
 
+	uart.rx_sw_fifo = malloc(sizeof(lf_fifo_t) + RX_SW_FIFO_SIZE * sizeof(uart.rx_sw_fifo->data[0]));
```
agkaminski, line 658: All of malloc size arguments are const, this operation is always performed, imho let's avoid malloc if possible (it is possible here)

tty/imx6ull-uart/lf_fifo.h
```diff
@@ -0,0 +1,71 @@
+/*
```
agkaminski, line 1: Maybe not now, but imho this should be in corelibs

tty/libtty/libtty.c
```diff
@@ -170,21 +170,37 @@ ssize_t libtty_read_nonblock(libtty_common_t *tty, char *data, size_t size, unsi
...
-	if (fifo_freespace(tty->tx_fifo) >= TX_FIFO_NOTFULL_WATERMARK) {
-		if (wake_writer)
-			*wake_writer = 1;
-		condSignal(tty->tx_waitq);
+	if (wake_writer != NULL) {
+		*wake_writer = fifo_freespace(tty->tx_fifo) >= TX_FIFO_NOTFULL_WATERMARK;
+	}
+	else {
+		if (fifo_freespace(tty->tx_fifo) >= TX_FIFO_NOTFULL_WATERMARK)
+			condSignal(tty->tx_waitq);
```
agkaminski, line 184: `libtty_wake_writer` fits here perfectly

agkaminski (commented):
If you feel like it maybe adaptation to the new style (brackets everywhere)? But this is more a thing for separate PR. I see no point in adding a forest of debugging log_* macros, which are not even used anyway.

Register bits style is imho ok, it adheres to our new guidelines (avoiding enums used as int), but our convertion is rather to use bit number directly and comment it's use. `(1 << 2)` says not much, but so does `UCR3_RXDMUXSEL`. But `(1 << 2)` along with comment what we're doing (maybe even why we're doing it) is descriptive enough. You can define the bits however you like, but a comment is still advised and you still need to read the chip manual to perform changes, so TBH defines for bits seem to me like a waste of time and code lines.

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -29,12 +32,71 @@
...
+#define COL_NORMAL "\033[0m"
+
+#define LOG_TAG "imx6ull-uart"
+
+/* clang-format off */
+#define log_debug(fmt, ...) do { log_printf(LOG_DEBUG, fmt "\n", ##__VA_ARGS__); } while (0)
+#define log_info(fmt, ...)  do { log_printf(LOG_INFO, fmt "\n", ##__VA_ARGS__); } while (0)
+#define log_warn(fmt, ...)  do { log_printf(LOG_WARNING, COL_YELLOW fmt COL_NORMAL "\n", ##__VA_ARGS__); } while (0)
+#define log_error(fmt, ...) do { log_printf(LOG_ERR, COL_RED fmt COL_NORMAL "\n", ##__VA_ARGS__); } while (0)
+/* clang-format on */
```
agkaminski, line 54: Do we need this? I don't see these used... Only log_warn is used. Imho such logs should be for debug only, not enabled by default (so disabled under #if NDEBUG)

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -29,12 +32,71 @@
...
+
+#define READER 1
+#define WRITER 2
+
+#define MAX_READERS 255
```
agkaminski, line 59: I thought we want only one reader to be allowed

tty/imx6ull-uart/imx6ull-uart.c
ziemleszcz: No, I can use `printf` directly if you don't like these defines. Currently I use `log_warn` to print UART errors which is useful for checking problems on DCU. These errors are not printed by default (option `-e`).

tty/imx6ull-uart/imx6ull-uart.c
ziemleszcz: I thought you have requested it. One reader is fine by me. Linux allows multiple readers.

agkaminski:
Side note - don't ttys on NOMMU targets require change too?

tty/imx6ull-uart/imx6ull-uart.c
agkaminski: Ok, so misunderstanding, I thought we wanted to restrict to one reader only. I see no problem with allowing multiple readers on a driver side, it shouldn't really bother the driver who and how consumes the data.

tty/imx6ull-uart/imx6ull-uart.c
agkaminski: Ok, perhaps might be useful for the future debugging, developement

ziemleszcz:
By NOMMU targets you mean multi drives, right? libtty API has not changed so they should be fine. I can replace `libtty_getchar` with `libtty_popchar` if you want.

agkaminski:
> By NOMMU targets you mean multi drives, right? libtty API has not changed so they should be fine. I can replace `libtty_getchar` with `libtty_popchar` if you want.

It's ok, I'm just making sure these weren't overlooked.

PR #211, opened 2022-05-11
storage/imx6ull-flash: add mtd interface compatible with libstorage

HBuczynski:
## Description

 - flashdev intializes storage_dev_t and provides pointers to functions
 - flashdev provides interface to NAND flash memory for server as well,
 - current version supports only mtd char interface, mtd block is not defined,
 - some of the functions derived from the previous mtd implementation from[ phoenix-rtos-filesystems/jffs2/phoenix-rtos/mtd.c](https://github.com/phoenix-rtos/phoenix-rtos-filesystems/blob/master/jffs2/phoenix-rtos/mtd.c)

**Note for reviewers:**
 - I am not sure whether function `_flashmtd_errRead` is needed. I leave it to be compatible with previous implementation,
 - if bad block occurs, `flashmtd_erase` continues operation - compatibility with flash server implementation. `jffs2` checks bad blocks on its own, so it shoudn't bother it 
 - `flashmtd_metaRead` - should operation be breaked, if error occurs in `_flashmtd_errRead` ?

I have tested the code on imx6ull platform for DTR project. The following tests have been preformed:
 - **psd** - works as expected,
 - **jffs2** - running parallel on three partitions,
 - **factory** flashing for DTR project
 
## Motivation and Context

DTR-102

## Types of changes

## How Has This Been Tested?

- [ ] Tested by hand on: (imx6ull, DTR).

## Checklist:

## Special treatment

### DIFF
--- a/storage/imx6ull-flash/flashdev.c
+++ b/storage/imx6ull-flash/flashdev.c
@@ -1,0 +1,439 @@
+/*
+ * Phoenix-RTOS
+ *
+ * IMX6ULL NAND flash device interface
+ *
+ * Copyright 2022 Phoenix Systems
+ * Author: Hubert Buczynski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/threads.h>
+#include <sys/minmax.h>
+
+#include "imx6ull-flashdev.h"
+#include "imx6ull-flashdrv.h"
+
+#define ECC_BITFLIP_THRESHOLD 10
+#define ECCN_BITFLIP_STRENGHT 14
+#define ECC0_BITFLIP_STRENGHT 16
+
+
+/* Auxiliary functions */
+
+/* Function derived from the previous code phoenix-rtos-filesystems/jffs/phoenix-rtos/mtd.c */
+static int _flashmtd_errRead(int ret, flashdrv_meta_t *meta, int status, void *data)
+{
+	int max_bitflip = 0;
+	unsigned *blk;
+	int i, j, err = 0;
+	/* check current read status. if it is EBADMSG we have nothing to do here since
+	 * read is marked as uncorrectable */
+	if (status == -EBADMSG) {
+		return status;
+	}
+
+	/* check status for every block */
+	for (i = 0; i < sizeof(meta->errors) / sizeof(meta->errors[0]); i++) {
+		/* no error or all ones */
+		if (meta->errors[i] == 0x0 || meta->errors[i] == 0xff) {
+			continue;
+		}
+		/* uncorrectable errors but maybe block is erased and we can still use it */
+		/* TODO: check if this actually works */
+		else if (meta->errors[i] == 0xfe) {
+			err = 0;
+			if (i == 0) {
+				blk = (unsigned *)meta->metadata;
+				for (j = 0; j < sizeof(meta->metadata) / sizeof(unsigned); ++j) {
+					err += (sizeof(unsigned) * 8) - __builtin_popcount(*blk + j);
+				}
+
+				if (err > ECC0_BITFLIP_STRENGHT) {
+					return -EBADMSG;
+				}
+
+				memset(blk, 0xff, sizeof(meta->metadata));
+				max_bitflip = max(max_bitflip, err);
+			}
+			else {
+				blk = data + ((i - 1) * 512);
+				for (j = 0; j < 128; j++) {
+					err += (sizeof(unsigned) * 8) - __builtin_popcount(*(blk + j));
+				}
+
+				if (err > ECCN_BITFLIP_STRENGHT) {
+					return -EBADMSG;
+				}
+
+				memset(blk, 0xff, 512);
+				max_bitflip = max(max_bitflip, err);
+			}
+		}
+		/* correctable errors */
+		else {
+			max_bitflip = max(max_bitflip, (int)meta->errors[i]);
+		}
+	}
+
+	return max_bitflip >= ECC_BITFLIP_THRESHOLD ? -EUCLEAN : status;
+}
+
+
+/* MTD interface */
+
+static int flashmtd_erase(struct _storage_t *strg, off_t offs, size_t size)
+{
+	int res;
+	uint32_t pageID;
+	uint32_t eb, startBlock, endBlock, erased = 0; /* values in erase block units */
+	unsigned int pagesPerBlock;
+
+	if ((offs % strg->dev->mtd->erasesz) != 0 || (size % strg->dev->mtd->erasesz) != 0) {
+		return -EINVAL;
+	}
+
+	startBlock = offs / strg->dev->mtd->erasesz;
+	endBlock = (offs + size) / strg->dev->mtd->erasesz;
+	pagesPerBlock = strg->dev->mtd->erasesz / strg->dev->mtd->writesz;
+
+	mutexLock(strg->dev->ctx->lock);
+	for (eb = startBlock; eb < endBlock; ++eb) {
+		pageID = eb * pagesPerBlock;
+
+		res = flashdrv_isbad(strg->dev->ctx->dma, pageID);
+		if (res != 0) {
+			printf("flashmtd_erase: skipping bad block: %lld\n", eb);
+			continue;
+		}
+
+		res = flashdrv_erase(strg->dev->ctx->dma, pageID);
+		if (res < 0) {
+			printf("flashmtd_erase: error while erasing block: %lld - marking as badblock", eb);
+			res = flashdrv_markbad(strg->dev->ctx->dma, pageID);
+			if (res < 0) {
+				mutexUnlock(strg->dev->ctx->lock);
+				return res; /* can't mark as badblock, urecoverable error */
+			}
+		}
+		else {
+			++erased;
+		}
+	}
+	mutexUnlock(strg->dev->ctx->lock);
+
+	return erased;
+}
+
+
+static ssize_t flashmtd_read(struct _storage_t *strg, off_t offs, void *data, size_t len)
+{
+	int res, err = 0;
+	size_t tempsz = 0, chunksz;
+
+	mutexLock(strg->dev->ctx->lock);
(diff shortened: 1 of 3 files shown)


storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,423 @@
...
+#define ECCN_BITFLIP_STRENGHT 14
+#define ECC0_BITFLIP_STRENGHT 16
+
+
+/* Auxiliary functions */
+
+/* Function derived from the previous code phoenix-rtos-filesystems/jffs/phoenix-rtos/mtd.c */
+static int _flashmtd_errRead(int ret, flashdrv_meta_t *meta, int status, void *data)
+{
+	int max_bitflip = 0;
```
agkaminski, line 34: camelCase
```suggestion
	int maxBitflip = 0;
```

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,423 @@
...
+		/* no error or all ones */
+		if (meta->errors[i] == 0x0 || meta->errors[i] == 0xff) {
+			continue;
+		}
+		/* uncorrectable errors but maybe block is erased and we can still use it */
+		/* TODO: check if this actually works */
+		else if (meta->errors[i] == 0xfe) {
+			err = 0;
+			if (i == 0) {
+				blk = (unsigned *)meta->metadata;
```
agkaminski, line 54: This is undefined behaviour. metada should be defined as array of unsigned (or better uint32_t) or memcpy should be used

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,423 @@
...
+				for (j = 0; j < sizeof(meta->metadata) / sizeof(unsigned); ++j) {
+					err += (sizeof(unsigned) * 8) - __builtin_popcount(*blk + j);
```
agkaminski, line 56: Where does 8 come from? Some sizeof would be better

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,423 @@
...
+					return -EBADMSG;
+				}
+
+				memset(blk, 0xff, sizeof(meta->metadata));
+				max_bitflip = max(max_bitflip, err);
+			}
+			else {
+				blk = data + ((i - 1) * 512);
+				for (j = 0; j < 128; j++) {
+					err += (sizeof(unsigned) * 8) - __builtin_popcount(*(blk + j));
```
agkaminski, line 69: Not obvious numbers used: 512, 128 and 8. At least comment would be nice

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,423 @@
...
+	strg->dev->ctx = malloc(sizeof(storage_devCtx_t));
+	if (strg->dev->ctx == NULL) {
+		free(strg->dev);
+		return -ENOMEM;
+	}
+
+	strg->dev->ctx->dma = flashdrv_dmanew();
+	if (strg->dev->ctx->dma == MAP_FAILED) {
+		free(strg->dev);
+		free(strg->dev->ctx);
```
agkaminski, line 367: Oops, dangling pointer. Revert this order, applies to other places too

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,423 @@
...
+	/* Free device context */
+	munmap(strg->dev->ctx->metabuf, strg->dev->mtd->writesz);
+	munmap(strg->dev->ctx->databuf, 2 * strg->dev->mtd->writesz);
+	flashdrv_dmadestroy(strg->dev->ctx->dma);
+	resourceDestroy(strg->dev->ctx->lock);
+	free(strg->dev->ctx);
+
+	free(strg->dev->mtd);
+	free(strg->dev);
+
```
agkaminski, line 332: strg is freed outside?

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,423 @@
...
+	if (strg->dev->ctx->dma == MAP_FAILED) {
+		free(strg->dev);
+		free(strg->dev->ctx);
+		return -ENOMEM;
+	}
+
+	strg->dev->ctx->databuf = mmap(NULL, 2 * info->writesz, PROT_READ | PROT_WRITE, MAP_UNCACHED, OID_CONTIGUOUS, 0);
+	if (strg->dev->ctx->databuf == MAP_FAILED) {
+		free(strg->dev);
+		free(strg->dev->ctx);
```
agkaminski, line 374: These frees should be performed last (in reverse order)

Maxez:
Not a review, just my answers to questions from the PR description:
1. `_flashmtd_errRead()` has to be called after every read with ECC operation (for both data and metadata). In metadata case it should check only metadata corrections (errors[0]) - no need to check other page chunks. Maybe we should split it into 2 separate functions - `_flashmtd_metaReadErr()` and `_flashmtd_readErr()`?
2. We can't erase bad blocks so skipping them in `flashmtd_erase()` is fine. I think that `flashmtd_read()` and `flashmtd_write()` should skip bad blocks too.
3. Yes, every read with ECC error should cause immediate return with error.

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,434 @@
...
+
+	/* check status for every block */
+	for (i = 0; i < sizeof(meta->errors) / sizeof(meta->errors[0]); i++) {
+		/* no error or all ones */
+		if (meta->errors[i] == 0x0 || meta->errors[i] == 0xff) {
+			continue;
+		}
+		/* uncorrectable errors but maybe block is erased and we can still use it */
+		/* TODO: check if this actually works */
+		else if (meta->errors[i] == 0xfe) {
```
Maxez, line 51: I think that this 'page chunk rescue' may cause trouble. If BCH module reports uncorrectable chunk (number of bitflips is bigger than ECC strength for that chunk: 16 bits for metadata and 14 bits for data) we should just return with error and cause ECC read operation to fail.
I'd suggest to remove it in a separate PR.

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,434 @@
...
+		}
+	}
+
+	return max_bitflip >= ECC_BITFLIP_THRESHOLD ? -EUCLEAN : status;
+}
+
+
+/* MTD interface */
+
+static int flashmtd_erase(struct _storage_t *strg, off_t offs, size_t size)
```
Maxez, line 92: I'd use `storage_t * strg` instead of `struct _storage_t *strg`. Repeats in other places in this PR.

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,434 @@
...
+		if (res < 0) {
+			break;
+		}
+
+		tempsz += chunksz;
+		offs += strg->dev->mtd->writesz;
+	}
+	mutexUnlock(strg->dev->ctx->lock);
+
+	return res >= 0 ? len : res;
```
Maxez, line 266: Should be `return tempsz > 0 ? tempsz : res;` for consistent semantics with other functions.

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,434 @@
...
+			mutexUnlock(strg->dev->ctx->lock);
+			return err;
+		}
+
+		chunksz = (strg->dev->mtd->writesz - (offs % strg->dev->mtd->writesz));
+		if (chunksz > len) {
+			chunksz = len;
+		}
+
+		memcpy(data, strg->dev->ctx->databuf + (offs % strg->dev->mtd->writesz), chunksz);
```
Maxez, line 155: `void *` arithmetic on `databuf` buffer.
Maybe change buffers defition to `unsigned char *`?

HBuczynski: Thanks. I have also changed in the rest of the code.

storage/imx6ull-flash/flashdev.c
HBuczynski: Right, it was left out.

storage/imx6ull-flash/flashdev.c
HBuczynski: This change would require changes to the interface and dependent modules. I leave it as it is.

storage/imx6ull-flash/flashdev.c
kemonats: Maybe
'''
#include <limits.h>
'''
and use `CHAR_BIT`
```
					err += (sizeof(unsigned) * CHAR_BIT) - __builtin_popcount(*blk + j);
```

storage/imx6ull-flash/flashdev.c
```diff
@@ -0,0 +1,434 @@
...
+				return res; /* can't mark as badblock, urecoverable error */
+			}
+		}
+		else {
+			++erased;
+		}
+	}
+	mutexUnlock(strg->dev->ctx->lock);
+
+	return erased;
```
kemonats, line 132: Why `erased` is type `long long` if You use it in this place as type `int`?
The same aplies to `pageID`.
`paigeID` is type `long long` but functions are declared as:
```
extern int flashdrv_isbad(flashdrv_dma_t *dma, uint32_t paddr);
extern int flashdrv_markbad(flashdrv_dma_t *dma, uint32_t paddr);
```
the second argument is type `uint32_t`.

If `offs` is to be of `long long` type, code where the divide 64bit is usedt can be improved by reducing the number of calls to `__aeabi_ldivmod()`.

storage/imx6ull-flash/flashdev.c
kemonats: Maybe 
```
#include <limits.h>
```
and use `CHAR_BIT`

storage/imx6ull-flash/flashdev.c
HBuczynski: Changed to uint32_t

HBuczynski:
**To sum up:**
Function `_flashmtd_errRead` is left as it is and will be changed in the next PR by @Maxez.
Current implementation is compatible with the previous logic of `flashsrv` on `imx6ull`.

agkaminski (approved):

PR #217, opened 2022-05-16
zynq7000-flash: cleaning up code: fixing minor vulnerabilities

HBuczynski:
## Description

 - update tests,
 - check input arguments in flashsrv,
 - move erase block from blkWrite to sync function

PD-216
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/storage/zynq7000-flash/zynq7000-flash.c
+++ b/storage/zynq7000-flash/zynq7000-flash.c
@@ -43,8 +43,11 @@ static ssize_t flash_read(oid_t *oid, offs_t offs, void *buff, size_t len)
 	ssize_t res = -ENOSYS;
 	storage_t *strg = storage_get(GET_STORAGE_ID(oid->id));
 
-	if (strg == NULL || strg->dev == NULL || (offs + len) >= strg->size) {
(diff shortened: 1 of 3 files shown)


storage/zynq7000-flash/zynq7000-flash.c
```diff
@@ -43,7 +43,7 @@ static ssize_t flash_read(oid_t *oid, offs_t offs, void *buff, size_t len)
 	ssize_t res = -ENOSYS;
 	storage_t *strg = storage_get(GET_STORAGE_ID(oid->id));
 
-	if (strg == NULL || strg->dev == NULL || (offs + len) >= strg->size) {
+	if (strg == NULL || strg->dev == NULL || (offs + len) >= strg->size || buff == NULL) {
```
agkaminski, line 46: In case of `len` zero `buff` shouldn't matter - but I don't know if this is a use case. I think you can achieve it by doing `write(NULL, fd, 0)`, valid call - we should get `0` from it, not `-1` and `EINVAL`

agkaminski (approved):

PR #226, opened 2022-05-24
sensors: implementation of sensor manager

HBuczynski:
## Description

The main goal of this module concerns on providing a basic interface for variety of sensors.

Assumptions:
 - the client communicates with manager via messages to **/dev/sensors**,
 - there is an external library **libsensors** containing sensor structures and message events - external API for the client
 - sensors server takes sensors description as arguments list:  <sensor_name:bus:others>; arguments might be specific for the sensor driver,
 - each sensor registers itself in a sensor manager via constructor,

**Example usage of the server:**
`sensors -s lsm9dsxx:/dev/spi1.0:argv -s lps25xx:/dev/i2c0`

**Example code of client app:**
```c
#include <stdio.h>
#include <stdlib.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/ioctl.h>

#include <libsensors.h>

int main(int argc, char **argv)
{
	int fd;
	sensors_data_t *data;
	unsigned char buff[0x400];
	sensor_type_t types;
	sensors_ops_t ops = {0};

	fd = open("/dev/sensors", O_RDWR);
	if (fd < 0) {
		return EXIT_FAILURE;
	}

	/* Get available sensor types */
	ioctl(fd, SMIOC_SENSORSAVAIL, &types);

	/* Define sensor types request */
	ops.types = (types & SENSOR_TYPE_BARO) | (types & SENSOR_TYPE_ACCEL) | (types & SENSOR_TYPE_GYRO);
	ioctl(fd, SMIOC_SENSORSSET, &ops);

	if ((ops.evtSz * sizeof(sensor_event_t) + sizeof(((sensors_data_t *)0)->size)) > sizeof(buff)) {
		fprintf(stderr, "Buff is too small\n");
		return EXIT_FAILURE;
	}

	for (int i = 0; i < 15; ++i) {
		sleep(1);
		read(fd, buff, 0x600);

		data = (sensors_data_t *)buff;
		printf("Sz: %d\n", data->size);

		for (int j = 0; j < data->size; ++j) {
			switch (data->events[j].type) {
			case SENSOR_TYPE_ACCEL:
				printf("Accel, timestamp: %llu\n", data->events[j].timestamp);
				printf("acce - x: %u, y: %u, z: %u\n", data->events[j].accels.accelX, data->events[j].accels.accelY, data->events[j].accels.accelZ);
				break;

			case SENSOR_TYPE_BARO:
				printf("Baro, timestamp: %llu\n", data->events[j].timestamp);
				printf("baro - pressure: %u\n", data->events[j].baro.pressure);
				break;

			case SENSOR_TYPE_GYRO:
				printf("Gyro, timestamp: %llu\n", data->events[j].timestamp);
				printf("gyro - x: %u, y: %u, z: %u\n", data->events[j].gyro.gyroX, data->events[j].gyro.gyroY, data->events[j].gyro.gyroZ);
				break;

			default:
				break;
			}
		}
	}

	close(fd);

	return EXIT_SUCCESS;
}

```

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


sensors/include/libsensors.h
```diff
@@ -0,0 +1,136 @@
...
+	stTemp = SENSOR_BIT(5),
+} sensor_type_t;
+
+
+typedef struct {
+	uint32_t devId;
+	int16_t accelX; /* value [mm/s^2] */
+	int16_t accelY; /* value [mm/s^2] */
+	int16_t accelZ; /* value [mm/s^2] */
+	uint8_t res[2];
```
agkaminski, line 44: What does this do? Why 2 elements?

sensors/include/libsensors.h
```diff
@@ -0,0 +1,136 @@
...
+} accel_data_t;
+
+
+typedef struct {
+	uint32_t devId;
+	uint32_t pressure; /* pressure in [Pa] */
+	int32_t temp;      /* temperature value in 1E-3 [C] */
```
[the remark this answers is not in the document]
agkaminski: @niewim19 What's your opinion? What unit does better suit Kalman etc needs?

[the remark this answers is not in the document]
HBuczynski: I have discussed this issue with @niewim19 and we agree that change to Kelvin is ok.

sensors/include/libsensors.h
```diff
@@ -0,0 +1,136 @@
...
+#define SENSOR_BIT(n) (1 << n)
+
+typedef enum {
+	stAccel = SENSOR_BIT(0),
+	stBaro = SENSOR_BIT(1),
+	stGps = SENSOR_BIT(2),
+	stGyro = SENSOR_BIT(3),
+	stMag = SENSOR_BIT(4),
+	stTemp = SENSOR_BIT(5),
+} sensor_type_t;
```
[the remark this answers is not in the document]
HBuczynski: Ok. I have made changes according to Your suggestion. Sensor types are declared as `#defines` and `uint32_t` variable is used instead of `enum`.

sensors/sensors.h
```diff
@@ -0,0 +1,64 @@
...
+/* Info structure is filled by a sensor driver */
+typedef struct {
+	unsigned int id;  /* device identifier in sensor manager */
+	uint32_t sensors; /* supported sensor types */
+	void *ctx;        /* internal context for specific sensor */
+} sensor_info_t;
+
+
+/* Basic operations on sensors */
+typedef struct {
```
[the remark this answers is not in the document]
HBuczynski: This information is placed in `sensor_info_t` structure, adding it into `sensor_ops_t` is redundant.

sensors/include/libsensors.h
```diff
@@ -0,0 +1,133 @@
...
+	int32_t altEllipsoid; /* altitude in 1E-3 [m] (millimetres) above Ellipsoid */
+	int32_t groundSpeed;  /* GPS ground speed, [mm/s] */
+	int32_t velNorth;     /* GPS North velocity, [mm/s] */
+	int32_t velEast;      /* GPS East velocity, [mm/s] */
+	int32_t velDown;      /* GPS Down velocity, [mm/s] */
+	uint32_t eph;         /* GPS horizontal position accuracy 1E-3 [m] (millimetres)*/
+	uint32_t epv;         /* GPS vertical position accuracy 1E-3 [m] (millimetres)*/
+	int32_t heading;      /* value in [mrad] */
+	int32_t headingOffs;  /* value in [mrad] */
+	int32_t headingAccur; /* value in [mrad] */
```
niewim19, line 67: IMHO 16 bit precision can be used.

None of these values exceeds 2π. With 10^3 precision we are far below `uint16_t` limit. Also only `headingOffs` may be negative so unsigned types can be used for heading and heading accuracy

niewim19 (commented):
Minor remarks about variable types in `gps_data_t`.

sensors/baro/lps25xx.c
```diff
@@ -0,0 +1,98 @@
...
+	ctx->evBaro.type = SENSOR_TYPE_BARO;
+	ctx->evBaro.baro.devId = info->id;
+
+	info->ctx = ctx;
+
+	err = beginthread(lps25xx_publishthr, 4, ctx->stack, sizeof(ctx->stack), info);
+	if (err < 0) {
+		free(ctx);
+	}
+	printf("lps25xx: launched sensor\n");
```
agkaminski, line 73: So we don't get our hopes high on error
```suggestion
	if (err < 0) {
		free(ctx);
	}
	else {
		printf("lps25xx: launched sensor\n");
	}
```

sensors/sensors.c
```diff
@@ -0,0 +1,600 @@
...
+	}
+
+	mutexLock(sensors_common.cLock);
+	res = idtree_alloc(&sensors_common.clients, &client->node);
+	if (res < 0) {
+		free(client);
+	}
+	else {
+		res = CLIENT_SET_ID(res);
+	}
```
agkaminski, line 97: This can be moved out of mutex

sensors/sensors.c
```diff
@@ -0,0 +1,600 @@
...
+	client = lib_treeof(sensor_client_t, node, idtree_find(&sensors_common.clients, id));
+	if (client != NULL) {
+		err = EOK;
+		idtree_remove(&sensors_common.clients, &client->node);
+		free(client);
+	}
+	mutexUnlock(sensors_common.cLock);
+
+	return err;
+}
```
agkaminski, line 119: Is this needed? I guess it's future proofing, but int such case shouldn't we have some refcnt for clients, so we don't remove and free client still in use?

sensors/sensors.c
```diff
@@ -0,0 +1,600 @@
...
+
+
+/* Function return total number of events for selected types */
+static unsigned int sensors_evtsCnt(sensor_type_t types)
+{
+	uint8_t id;
+	unsigned int sz = 0;
+
+	for (id = __builtin_ffs(types); id != 0; types &= ~(1 << (id - 1)), id = __builtin_ffs(types)) {
+		sz += sensors_common.evtNb[id - 1];
```
agkaminski, line 179: Shouldn't this be protected by mutex?

sensors/sensors.c
```diff
@@ -0,0 +1,600 @@
...
+
+
+/* Function filters requested types with the available ones */
+static sensor_type_t sensors_checkTypes(sensor_type_t types)
+{
+	uint8_t id;
+	sensor_type_t avail = 0;
+
+	for (id = __builtin_ffs(types); id != 0; types &= ~(1 << (id - 1)), id = __builtin_ffs(types)) {
+		if (sensors_common.evtNb[id - 1] != 0) {
```
agkaminski, line 193: Shouldn't this be protected by mutex?

sensors/sensors.c
```diff
@@ -0,0 +1,600 @@
...
+
+				client->ops.types = ops.types;
+				client->ops.evtSz = ops.evtSz;
+
+				outData = (void *)&ops;
+				break;
+
+			case SMIOC_SENSORSAVAIL:
+				for (i = 0; i < NB_SENSOR_TYPES; ++i) {
+					if (sensors_common.evtNb[i] != 0) {
```
agkaminski, line 234: Shouldn't this be protected by mutex?

sensors/sensors.c
```diff
@@ -0,0 +1,600 @@
...
+		switch (c) {
+			case 's': /* <sensor:args> */
+				name = optarg;
+				args = strchr(optarg, ':');
+				if (args == NULL) {
+					fprintf(stderr, "sensors: missing a sensors arguments: %s\n", optarg);
+					return devsz;
+				}
+
+				*args++ = '\0';
```
agkaminski, line 469: Imho better make operation order explicit 
```suggestion
				*(args++) = '\0';
```

sensors/sensors.c
```diff
@@ -0,0 +1,600 @@
...
+	if (err) {
+		fprintf(stderr, "sensors: could not create a device\n");
+		return -EINVAL;
+	}
+
+	return EOK;
+}
+
+
+void sensors_cleanup(size_t devsz)
```
agkaminski, line 515: I am not sure if this is really needed - we would rather want to deinit device drivers, memory management is not really needed - it's faster to leave this to the kernel. @nalajcie your opinion?

storage/zynq7000-flash/flashdrv.c
```diff
@@ -191,7 +191,7 @@ static ssize_t _flashdrv_pageProgram(unsigned int id, addr_t offs, const void *b
 	ssize_t res;
 	time_t timeout;
 	const flash_cfi_t *cfi = &fdrv_common.info.cfi;
-	const flash_cmd_t *cmd = &fdrv_common.info.cmds[flash_cmd_pp];
+	const flash_cmd_t *cmd = &fdrv_common.info.cmds[flash_cmd_qpp];
```
agkaminski, line 194: Surprise change not related to the PR :)

sensors/sensors.c
HBuczynski: This table is only modified within init functions, after that it is read only.

sensors/baro/lps25xx.c
```diff
@@ -0,0 +1,98 @@
...
+	if (ctx == NULL) {
+		return -ENOMEM;
+	}
+
+	ctx->evBaro.type = SENSOR_TYPE_BARO;
+	ctx->evBaro.baro.devId = info->id;
+
+	info->ctx = ctx;
+
+	err = beginthread(lps25xx_publishthr, 4, ctx->stack, sizeof(ctx->stack), info);
```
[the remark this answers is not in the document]
HBuczynski: Ok, it is only an example. In the final driver, it can be changed.

damianloew (commented):
There are some new warnings, when building for `armv7a9-zynq7000-qemu`

sensors/sensors.c
HBuczynski: I have added reference counter for each of the client.

HBuczynski:
> There are some new warnings, when building for `armv7a9-zynq7000-qemu`
> 
> Screenshot from 2022-06-03 14-24-33

Thanks for taking a look on compilation outputs. According to https://github.com/phoenix-rtos/phoenix-rtos-build/blob/39c2ff5611d0e7599ba794302659389f3d0c442d/makes/binary.mk#L26 I've added `LOCAL_HEADERS_DIR := nothing`

sensors/sensors.c
```diff
@@ -0,0 +1,630 @@
...
+} sensors_common;
+
+
+/* Update data from sensors */
+
+int sensors_publish(unsigned int devId, const sensor_event_t *event)
+{
+	int err = EOK;
+	uint8_t id = __builtin_ffs(event->type);
+	uint8_t evtId = sensors_common.devEvents[devId][id];
```
agkaminski, line 61: This is a bad access - id hasn't been decremented yet, it's outside of a mutex

sensors/sensors.c
```diff
@@ -0,0 +1,630 @@
...
+
+	mutexLock(sensors_common.cLock);
+	res = idtree_alloc(&sensors_common.clients, &client->node);
+	mutexUnlock(sensors_common.cLock);
+
+	if (res < 0) {
+		free(client);
+	}
+	else {
+		client->refs = 1;
```
agkaminski, line 131: Set refs counter before adding to the tree

sensors/sensors.c
```diff
@@ -0,0 +1,630 @@
...
+
+
+static int sensors_close(id_t id)
+{
+	int err = -EINVAL;
+	sensor_client_t *client;
+
+	client = sensors_clientFind(id);
+	if (client != NULL) {
+		sensors_clientPut(client);
```
agkaminski, line 146: Find increments ref counter - you need to `put` it twice to remove the resource

sensors/sensors.c
HBuczynski: done

sensors/sensors.c
HBuczynski: Done

sensors/sensors.c
HBuczynski: I agree that there is a bad access. However, the `sensors_common.devEvent` is read only, so that it doesn't have to be protected by the mutex.

agkaminski (dismissed):
LGTM

niewim19 (approved):
LGTM

niewim19 (approved):
LGTM

PR #271, opened 2022-07-26
sensorhub: decrease magnetometer base sampling freq. from 80Hz to 64Hz

niewim19:
JIRA: PP-51

## Description

Longer wait is needed for magnetometer thread to not truncate data when sampling the sensor without uisng DRDY (data ready) register check. 

## Motivation and Context

We skip DRDY check as it is not important for us to achieve absolute maximum magnetometer data sampling performance. It comes out, that precise 80Hz sampling causes data loss on some axis. Thus the closest and possibly round value of 64 Hz was chosen as sampling frequency. Tested on zturn drone this fixes the problem. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: lsm9dsxx with zturn.

## Checklist:

## Special treatment

### DIFF
--- a/sensors/mag/lsm9dsxx.c
+++ b/sensors/mag/lsm9dsxx.c
@@ -158,7 +158,7 @@ static void lsm9dsxx_threadPublish(void *data)
 	uint8_t obuf;
 
 	while (1) {
-		usleep(12500); /* 80Hz sampling period */
+		usleep(1000 * 1000 / 64);
 
 		obuf = REG_DATA_OUT | SPI_READ_BIT | SPI_AUTOADDRINCR_BIT;
 		err = sensorsspi_xfer(&ctx->spiCtx, &ctx->spiSS, &obuf, sizeof(obuf), ibuf, sizeof(ibuf), sizeof(obuf));


nalajcie:
Please move the description as to WHY the change was needed to commit message body (instead of writing it only in PR) - the commit messages stay with us forever, the PR descriptions are as good as lost after the merge :).

Side note: it could be written as `usleep((1000*1000) / 64)` so I would see 64Hz without additional comment. Also, sleep-based sampling time will always be irregular (and below expected 64Hz) but I suppose it doesn't matter in this case.

niewim19:
> Please move the description as to WHY the change was needed to commit message body (instead of writing it only in PR) - the commit messages stay with us forever, the PR descriptions are as good as lost after the merge :).
> 
> Side note: it could be written as `usleep((1000*1000) / 64)` so I would see 64Hz without additional comment. Also, sleep-based sampling time will always be irregular (and below expected 64Hz) but I suppose it doesn't matter in this case.



> Please move the description as to WHY the change was needed to commit message body (instead of writing it only in PR) - the commit messages stay with us forever, the PR descriptions are as good as lost after the merge :).
> 
> Side note: it could be written as `usleep((1000*1000) / 64)` so I would see 64Hz without additional comment. Also, sleep-based sampling time will always be irregular (and below expected 64Hz) but I suppose it doesn't matter in this case.

- Moved description to preserve change cause. Thanks for that!
- changed notation to proposed and removed comment.
- Yes, I am keeping in mind that this frequency is somewhat theoretical max. Thus 80Hz was chosen at start so with processing included I was counting on good results. Comes out that more time is needed. Magnetometer data at 64Hz is still very good frequency (i think we would be good even at around 10Hz)

HBuczynski (approved):

PR #278, opened 2022-08-10
imxrt-multi: uart parity bit appending fix

gorbanchovie:
JIRA: NIL-287

## Description

A fix to appending parity bit to UART data frames for i.MX RT MCUs.
## Motivation and Context

UART configuration for i.MX RT MCUs is different then, for example, for i.MX 6ULL. Parity bit is treated as a part of data bits. So in order to append parity bit to data frame, data length must be incremented.
## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: i.MX RT1170.

## Checklist:

## Special treatment

### DIFF
--- a/multi/imxrt-multi/uart.c
+++ b/multi/imxrt-multi/uart.c
@@ -155,7 +155,7 @@ static void signal_txready(void *_uart)
 }
 
 
-static void set_cflag(void *_uart, tcflag_t* cflag)
+static void set_cflag(void *_uart, tcflag_t *cflag)
 {
 	uart_t *uartptr = (uart_t *)_uart;
 	uint32_t t;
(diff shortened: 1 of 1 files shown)


agkaminski (dismissed):
LGTM, cross-checked with the RM

multi/imxrt-multi/uart.c
```diff
@@ -155,24 +155,28 @@ static void signal_txready(void *_uart)
...
+
+	if ((*cflag & CSIZE) == CS7 && !(*cflag & PARENB)) {
+		t |= 1 << 11;
 	}
 	else { /* CS8 */
+		if (*cflag & PARENB) {
+			t |= 1 << 4;
+		}
 		*cflag &= ~CSIZE;
 		*cflag |= CS8;
```
anglov, line 177: It looks invalid.  When you have eg. 7E1  it will set CSIZE to CS8

anglov (changes requested):

anglov (approved):
Solution looks hacky but I think it's valid

agkaminski (approved):
@anglov  Why hacky? One thing I see (but it's present in the old code too) is that we silently set CS8 if CSIZE != CS7.

PR #282, opened 2022-08-22
stm32l4-multi: fix bugs in otp handling

Diabliz:
Add -EINVAL returning in case of unaligment address or size when read from otp.
Fix returning value from otp functions to be signed type.

JIRA: SKAL-527

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4x).

### DIFF
--- a/multi/stm32l4-multi/flash.c
+++ b/multi/stm32l4-multi/flash.c
@@ -143,13 +143,12 @@ size_t flash_readData(uint32_t offset, char *buff, size_t size)
 }
 
 
-size_t flash_readOtp(uint32_t offset, char *buff, size_t size)
+ssize_t flash_readOtp(uint32_t offset, char *buff, size_t size)
 {
 	size_t ret;
 
-	ret = otp_isValidAddress(offset + OTP_ADDR, size);
-	if (!ret) {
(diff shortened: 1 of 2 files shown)


multi/stm32l4-multi/flash.c
```diff
@@ -143,13 +143,13 @@ size_t flash_readData(uint32_t offset, char *buff, size_t size)
 }
 
 
-size_t flash_readOtp(uint32_t offset, char *buff, size_t size)
+ssize_t flash_readOtp(uint32_t offset, char *buff, size_t size)
 {
 	size_t ret;
 
 	ret = otp_isValidAddress(offset + OTP_ADDR, size);
 	if (!ret) {
```
agkaminski, line 151: otp_isValidAddress has no side-effects, so it's ok for direct if usage. On the other hand !ret is against MISRA rule regarding casting int -> bool
```suggestion
	if (otp_isValidAddress(offset + OTP_ADDR, size) == 0) {
```

agkaminski (approved):
One somewhat unrelated remark, otherwise LGTM

agkaminski (approved):

PR #298, opened 2022-09-27
stm32l4-multi: Add RTC backup storage implementation

agkaminski:
JIRA: ISC-198

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (stm32l4).

## Checklist:

- [x] My change requires a change to the documentation.

## Special treatment

### DIFF
--- a/multi/stm32l4-multi/rtc.c
+++ b/multi/stm32l4-multi/rtc.c
@@ -14,6 +14,7 @@
 
 #include <errno.h>
 #include <stdint.h>
+#include <string.h>
 #include <sys/threads.h>
 #include <sys/platform.h>
 
@@ -21,14 +22,21 @@
 #include "rcc.h"
 #include "exti.h"
 #include "rtc.h"
+#include "stm32l4-multi.h"
 
-
+/* clang-format off */
 enum { pwr_cr = 0, pwr_csr };
 
 
 enum { tr = 0, dr, cr, isr, prer, wutr, alrmar = wutr + 2, alrmbr, wpr, ssr, shiftr, tstr, tsdr, tsssr, calr,
 	tampcr, alrmassr, alrmbssr, or, bkp0r};
 
+#define BACKUP1_ID_REG    bkp0r
+#define BACKUP_PAYLOAD_SZ RTC_BACKUP_SZ
+#define BACKUP2_ID_REG    (BACKUP1_ID_REG + (BACKUP_PAYLOAD_SZ / sizeof(uint32_t)) + 1)
+#define ID_VALUE(id)      ((id) & 0xffff)
+#define ID_IS_VALID(id)   ((((id) >> 16) ^ ((id) & 0xffff)) == 0xffff)
+/* clang-format on */
 
 struct {
 	volatile unsigned int *base;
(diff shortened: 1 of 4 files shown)


multi/stm32l4-multi/rtc.c
```diff
@@ -232,6 +241,124 @@ int rtc_setAlarm(rtctimestamp_t *timestamp)
...
+
+	return ret;
+}
+
+
+int rtc_storeBackup(const void *buff, size_t bufflen)
+{
+	uint32_t idno;
+	int index = (rtc_getLastStorage(&idno) == 0) ? 1 : 0;
+	int idreg = (index == 0) ? BACKUP1_ID_REG : BACKUP2_ID_REG;
```
kemonats, line 300: No `index < 0 ` check?
Register swap?
Is this correct?

multi/stm32l4-multi/rtc.c
```diff
@@ -232,6 +241,124 @@ int rtc_setAlarm(rtctimestamp_t *timestamp)
...
+	int retval = -EINVAL;
+	size_t i;
+	uint32_t tmp[BACKUP_PAYLOAD_SZ / sizeof(uint32_t)] = { 0 };
+
+	if ((buff != NULL) || (bufflen != 0) || (bufflen <= BACKUP_PAYLOAD_SZ)) {
+		mutexLock(rtc_common.lock);
```
kemonats, line 306: I guess it should be:
```suggestion
	if ((buff != NULL) && (bufflen != 0) && (bufflen <= BACKUP_PAYLOAD_SZ)) {
		mutexLock(rtc_common.lock);
```

multi/stm32l4-multi/rtc.c
agkaminski: `index` cannot be < 0, if `rtc_getLastStorage` returns -1 then `index` = 0 (if there are no valid records I want to store data on position 1, i.e BACKUP1_ID_REG).

I swap the positions because I want to store data on the older record position (`rtc_getLastStorage` returns the newest one)

multi/stm32l4-multi/rtc.c
agkaminski: Whoops! Forgot to change that when I was reversing the logic. Thank you

multi/stm32l4-multi/rtc.c
```diff
@@ -232,6 +241,124 @@ int rtc_setAlarm(rtctimestamp_t *timestamp)
...
+			num = ID_VALUE(id[0]);
+			break;
+
+		case 2: /* Only record #2 is valid */
+			ret = 1;
+			num = ID_VALUE(id[1]);
+			break;
+
+		default: /* Both records are valid */
+			if (ID_VALUE(id[0]) > ID_VALUE(id[1])) {
```
anglov, line 277: On id[0] = 0xffff and id[1] = 0x0 because of overflow  this condition is invalid. What worse - it never repair because we will always write new records to id[1]

anglov (changes requested):
In general I like the solution but I think that it's not safe on overflow which is likely to occur (only 65535 values of id)

multi/stm32l4-multi/rtc.c
```diff
@@ -14,21 +14,30 @@
...
 #include <sys/platform.h>
 
 #include "common.h"
 #include "rcc.h"
 #include "exti.h"
 #include "rtc.h"
+#include "stm32l4-multi.h"
+
+/* clang-format off */
+#define BACKUP1_ID_REG    bkp0r
```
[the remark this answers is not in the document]
agkaminski: Tbh I don't really see a problem with this, I placed it before enum to adhere to our other coding style traditions (include, define, data types, functions order). Will change

multi/stm32l4-multi/rtc.c
agkaminski: Indeed, good catch. I'll think about a fix

multi/stm32l4-multi/rtc.c
```diff
@@ -232,6 +240,124 @@ int rtc_setAlarm(rtctimestamp_t *timestamp)
...
+	uint32_t idno;
+	int index = (rtc_getLastStorage(&idno) == 0) ? 1 : 0;
+	int idreg = (index == 0) ? BACKUP1_ID_REG : BACKUP2_ID_REG;
+	int retval = -EINVAL;
+	size_t i;
+	uint32_t tmp[BACKUP_PAYLOAD_SZ / sizeof(uint32_t)] = { 0 };
+
+	if ((buff != NULL) && (bufflen != 0) && (bufflen <= BACKUP_PAYLOAD_SZ)) {
+		mutexLock(rtc_common.lock);
+		_rtc_unlock();
```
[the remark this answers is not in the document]
agkaminski: Yes, I was just changing that, thx

anglov (dismissed):

PR #300, opened 2022-09-28
imx6ull/flashdev: Fix handling too many bitflips case

MaciejPurski:
JIRA: DTR-341

## Motivation and Context

This change is needed to fix handling the situation, when there are more than ECCN_BITFLIP_STRENGHT bitflips but less than ECC0_BITFLIP_STRENGHT on a data chunk. After this change we should return -EBADMSG.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


storage/imx6ull-flash/flashdev.c
```diff
@@ -99,113 +100,130 @@ static unsigned int _flashmtd_checkErased(const void *buff, size_t boffs, size_t
...
+			/* Let the user see this chunk as correctly erased */
+			_flashmtd_chunkCorrect(strg, chunkidx);
+			break;
+
+		default:
+			/* On any other case metadata error field contains number of corrected bits */
+			ret = meta->errors[chunkidx];
 	}
 
 	if (raw != NULL) {
```
nalajcie, line 170: && raw != MAP_FAILED

storage/imx6ull-flash/flashdev.c
```diff
@@ -99,113 +100,130 @@ static unsigned int _flashmtd_checkErased(const void *buff, size_t boffs, size_t
...
+				blen = metalen;
+			}
+			else {
+				boffs = metalen + (chunkidx - 1) * datalen;
+				blen = datalen;
+			}
+
+			ret = _flashmtd_checkErased(raw, boffs, blen);
+
+			/* Let the user see this chunk as correctly erased */
```
nalajcie, line 161: IMHO this should only happen if flips <= ecc_strenght (so it's a correctable error and we correct it to `0xff`), otherwise we don't know what the original data should be here

storage/imx6ull-flash/flashdev.c
```diff
@@ -99,113 +100,130 @@ static unsigned int _flashmtd_checkErased(const void *buff, size_t boffs, size_t
...
+			_flashmtd_chunkCorrect(strg, chunkidx);
+			break;
+
+		default:
+			/* On any other case metadata error field contains number of corrected bits */
+			ret = meta->errors[chunkidx];
```
nalajcie, line 167: probably good to have `break` here in case somebody would add new `case` not looking around

storage/imx6ull-flash/flashdev.c
```diff
@@ -99,113 +100,130 @@ static unsigned int _flashmtd_checkErased(const void *buff, size_t boffs, size_t
...
+	int ret, i;
+	unsigned int maxflips = 0;
+	struct _storage_devCtx_t *ctx = strg->dev->ctx;
+	flashdrv_meta_t *meta = ctx->metabuf;
+	const int nchunks = sizeof(meta->errors) / sizeof(meta->errors[0]);
+
+	for (i = 0; i < nchunks; i++) {
+		ret = _flashmtd_chunkFlips(strg, meta, pageaddr, i);
+		if (ret < 0) {
+			return ret;
```
nalajcie, line 190: If I understand correctly we can exit only with -ENOMEM here, still - how will `maxbitflips` react in that case? I think it won't output anything to `stdout` so our healthmon script will break

nalajcie (commented):
looks better than the original code, I've pointed out some potential issues.

storage/imx6ull-flash/flashdev.c
MaciejPurski: Our script now checks the exit code of `maxbitflips` command, so it will print a log, that there was a problem and won't do anything.

storage/imx6ull-flash/flashdev.c
MaciejPurski: I agree.

storage/imx6ull-flash/flashdev.c
```diff
@@ -99,113 +100,130 @@ static unsigned int _flashmtd_checkErased(const void *buff, size_t boffs, size_t
...
-					memset(data + (i - 1) * ECCN_DATA_SIZE, 0xff, ECCN_DATA_SIZE);
-				}
-
-				/* Chunk corrected, update max number of bitflips */
-				maxflips = max(flips, maxflips);
+		case flash_uncorrectable:
+			/* BCH reports chunk as uncorrectable in case of an erased page with bitflips within that chunk area */
+			/* Check if that's the case by counting the bitflips with an assumption that the page is fully erased */
+			/* Read raw page in order to check full chunk - both data and its ECC area */
+			raw = mmap(NULL, rawsz, PROT_READ | PROT_WRITE, MAP_UNCACHED, OID_CONTIGUOUS, 0);
```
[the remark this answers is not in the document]
MaciejPurski: Yes, it does, that's a one sad downside of my refactor. I assumed it does not happen so often to be worth making the code less readable, but okay, I'll try to optimize it.

storage/imx6ull-flash/flashdev.c
nalajcie: ok, I've missed that in the last CR patch iteration.

storage/imx6ull-flash/flashdev.c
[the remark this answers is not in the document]
nalajcie: ok, maybe it won't (I hope we won't have much unrecoverable errors while reading data as I'm assuming the client would skip badblocks (which might have invalid ECC due to being bad, no need to try to recover them). I think You can leave it as-is.

storage/imx6ull-flash/flashdev.c
```diff
@@ -30,6 +30,9 @@
 #define ECCN_BITFLIP_STRENGHT 14  /* ECC strength for data chunk */
 #define ECCN_DATA_SIZE        512 /* Data chunk size */
 
+#define CHUNK_IS_META(chunkidx)                ((chunkidx) == 0)
+#define CHUNK_IS_UNCORECTABLE(chunkidx, flips) ((CHUNK_IS_META(chunkidx) && (flips) > ECC0_BITFLIP_STRENGHT) || \
```
nalajcie, line 34: UNCORECTABLE -> UNCORRECTABLE

Also - missing `!` before CHUNK_IS_META in second line.

IMHO, this should be written as an inline function (much more readable, less likely to make an error):
```c
static inline int chunk_is_uncorrectable(chunkidx, flips)
{
  if (CHUNK_IS_META(chunkidx)) {
    return flips > ECC0_BITFLIP_STRENGHT;
  }
  else {
     return flips > ECCN_BITFLIP_STRENGHT;
  }
}
```

If You really want it to be a function-like macro, maybe use ternary operator, eg.:
```c
#define CHUNK_IS_UNCORECTABLE(chunkidx, flips) (CHUNK_IS_META(chunkidx) ? ((flips) > ECC0_BITFLIP_STRENGHT)) : ((flips) > ECCN_BITFLIP_STRENGHT)
```

storage/imx6ull-flash/flashdev.c
```diff
@@ -99,113 +102,141 @@ static unsigned int _flashmtd_checkErased(const void *buff, size_t boffs, size_t
...
 {
-	/* Metadata and data (with their ECC) chunk sizes in bits */
-	const size_t mlen = strg->dev->mtd->oobSize * CHAR_BIT + ECC_GF * ECC0_BITFLIP_STRENGHT;
-	const size_t dlen = ECCN_DATA_SIZE * CHAR_BIT + ECC_GF * ECCN_BITFLIP_STRENGHT;
 	flashdrv_meta_t *meta = strg->dev->ctx->metabuf;
 	unsigned char *data = strg->dev->ctx->databuf;
-	unsigned int i, flips, maxflips = 0;
-	size_t boffs, blen, rawsz;
+
+	if (nflips == 0) {
```
nalajcie, line 110: never happens, would be detected as `flash_erased`

storage/imx6ull-flash/flashdev.c
```diff
@@ -99,113 +102,141 @@ static unsigned int _flashmtd_checkErased(const void *buff, size_t boffs, size_t
...
+		/* Nothing to be corrected */
+		return;
+	}
+
+	if (CHUNK_IS_UNCORECTABLE(chunkidx, nflips)) {
```
nalajcie, line 115: IMHO it would be more readable to move this condition outside of this function (`_flashmtd_chunkCorrect` should correct the chunk, if it's conditional should be renamed to eg. `_flashmtd_chunkTryCorrect`)

storage/imx6ull-flash/flashdev.c
nalajcie: not applied

nalajcie (commented):
missing `!`, not applied MAP_FAILED and some nitpicking :v:

storage/imx6ull-flash/flashdev.c
MaciejPurski: I have applied it. I set raw to NULL in case mmap returns MAP_FAILED:
https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/4069c06d0c116e73117022a361fd4ae3db5b0550/storage/imx6ull-flash/flashdev.c#L153

storage/imx6ull-flash/flashdev.c
MaciejPurski: Okay, I can make it an inline function.

storage/imx6ull-flash/flashdev.c
nalajcie: sorry, my bad, not used to GH PRs and how to clearly see diffs between patchsets

storage/imx6ull-flash/flashdev.c
MaciejPurski: Yes, I agree it would be more readable to move this check outside this function.

nalajcie (approved):

PR #314, opened 2022-11-10
pa6h: more error calculations

niewim19:
## Description

 1) Addition of velocity error field to gps data structure
 2) Remove some of redundancies in parsing/interpreting nmea messages
    - nmea parsers do not return message type. They store it inside message structure passed to them
    - interpreter does not return message type.
    - pa6h_update returns information about the need of publishing new data

## Motivation and Context

 - hdop/vdop values are insufficient by itself to calculate error in position in SI units. For this an arbitrary value of base (minimum) error in position/velocity is needed which is device specific. Therefore calculation of such error must be done  in the driver module
 - update logic was corrected

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/sensors/gps/pa6h.c
+++ b/sensors/gps/pa6h.c
@@ -307,49 +307,50 @@ static int pa6h_paramSet(int fd, unsigned int msgId, const char *payload, enum p
 }
 
 
-int pa6h_update(nmea_t *message, pa6h_ctx_t *ctx)
+/* Returns 1 if `gpsEvt` was updated with data from `message`. Otherwise returns 0. */
+int pa6h_update(nmea_t *message, sensor_event_t *evtGps)
 {
 	/* timestamp update happens only on GPGGA message as it contains position */
 	switch (message->type) {
 		case nmea_gga:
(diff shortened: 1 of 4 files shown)


sensors/gps/pa6h.c
```diff
@@ -347,29 +359,24 @@ static void pa6h_threadPublish(void *data)
...
-			if (ret != nmea_broken && ret != nmea_unknown) {
-				pa6h_update(&message, ctx);
-				doUpdate = 1;
-			}
+			nmea_interpreter(inbox[i].msg, &message);
+			update |= pa6h_update(&message, &ctx->evtGps);
 		}
-		if (doUpdate == 1) {
+
+		if (update) {
```
agkaminski, line 378:
```suggestion
		if (update != 0) {
```

agkaminski (approved):

sensors/gps/pa6h.c
niewim19: Implemented, thanks!

agkaminski (approved):

PR #332, opened 2023-01-26
sensorhub: change default priorities of threads

niewim19:
## Description

Introduced and used defines for priorities of messaging thread and sensor threads.

## Motivation and Context

Sensorhub works between the low-level bus drivers (e.g spi) which are typically on priority 2 and higher level user applications which are typically on priorty 4. Thus it is sensible to run it on priority level 3.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


agkaminski (commented):
If we can assume that all sensors have the same priority it would be better to #define it in some header

niewim19:
> If we can assume that all sensors have the same priority it would be better to #define it in some header

Reimplemented in mentioned way.

agkaminski (approved):

PR #342, opened 2023-02-10
pa6h: startup configuration and lossless reading method

niewim19:
### Additions:
 - serial device configuration added to driver
 - driver startup configuration of PA6H device
   - pmtk packets templates
   - pmtk ack packet validation functions
   - gps parameters setup routine
 - new incoming message reading methods (mainly `pa6h_receiver()`):
   - message split between two `read()` operations are retrieved in the second one
   - message are checksum-validated before being passed to interpreters

Default configuration of GPS device is as follow:
 - communication is done  on baudrate 115200
 - pa6h outputs GPGGA, GPGSA and GPVTG messages
 - pa6h outputs at 10Hz
 - pa6h position fix frequency at 5Hz

## Motivation and Context

GPS PA6H device driver was implemented last year for phoenix-pilot project, but wasn`t yet used in it. Since then a need arose for the following features:
 - gps configuration on startup
 - faster output rate of data

Previous implementation was a single-read based, so all clipped data (either headless or tailless messages) was discarded. This is somehow sufficient for reading.  However successful communication through PMTK configuration protocol should be validated and this needs more _bulletproof_ message reading method.   

Proposed solution is not the most efficient however it's resilient. There are possible scenarios of data corruption that can cause redundant/pointless string operations/searches. Picking out all possible string-wise mishaps (e.g `$$$**$***$` will cause some pointless searches) would only obscure the goal in the code. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: zturn + pmodnav pa6h

## Checklist:

## Special treatment


sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+					break;
+				}
+				nextStartToken = strchr(tokenEnd, '$');
+			}
+
+			/* Ensure space for checksum and nul character not interfering with next message. Lack '/r/n' at the end is discarded */
+			len = ((nextStartToken == NULL) ? &buf[pos] : nextStartToken) - tokenEnd;
+			if (len < 3) {
+				break;
+			}
```
kemonats, line 156: The correct end of the record should look like this:
`*<ch1><ch2><CR><LF>`
and then
```suggestion
			if (len < 5) {
				break;
			}
			if ((*tokenEnd[3] != '\r') || (*tokenEnd[4] != '\n') {
				break;
			}
			*tokenEnd[3] = '\0';
```

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+		tokenEnd = tokenStart;
+		while (tokenStart != NULL) {
+			remStart = tokenStart;
+
+			/* Seek current message`s end. Avoid multiple search for the same 'tokenEnd' if 'tokenStart' is behind it */
+			if (tokenStart >= tokenEnd) {
+				tokenEnd = strchr(tokenStart, '*');
+				if (tokenEnd == NULL) {
+					break;
+				}
```
kemonats, line 148: You can seek `'*'` and calculate checksum.
```suggestion
				tokenEnd = tokenStart + 1;
				checksum = 0;
				while ((*tokenEnd != '\0') && (tokenEnd != '*') {
					checksum ^= *tokenEnd;
					tokenEnd++;
				}
				If (*tokenEnd == '\0') {
					tokenEnd = NULL;
				}
				if (tokenEnd == NULL) {
					break;
				}
```

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+	if (ret >= 0) {
+		/* update position and nul terminate the buffer */
+		pos += ret;
+		buf[pos] = '\0';
+
+		/* find beginning of first message */
+		tokenStart = strchr(buf, '$');
+		tokenEnd = tokenStart;
+		while (tokenStart != NULL) {
+			remStart = tokenStart;
```
kemonats, line 141: Checking free space in the `inbox` array.
```suggestion
		while (tokenStart != NULL) {
			if (inboxFill >= inboxSz) {
				break;
			}
			remStart = tokenStart;
```

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+				len = tokenEnd - tokenStart;
+
+				/* calculate checksum */
+				for (i = 1, checksum = 0; i <= len - 1; i++) {
+					checksum ^= tokenStart[i];
+				}
+
+				/* Save message if checksum matches and there is enough space */
+				if (checksum == val && inboxFill < inboxSz) {
+					inbox[inboxFill].msg = tokenStart;
```
kemonats, line 169: If the array `inbox` will be full, you loss valid records.
Here, only check for checksum validity.
```suggestion
				if (checksum == val) {
					inbox[inboxFill].msg = tokenStart;
```

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+					inbox[inboxFill].sz = len + 4; /* +1 for '*', +2 for checksum, +1 for nul */
+					inboxFill++;
+				}
+			}
+
+			/* find next starting token */
+			tokenStart = strchr(tokenStart + 1, '$');
```
kemonats, line 176:
```suggestion
			/* find next starting token */
			tokenStart = nextStartToken;
			if (tokenStart != NULL) {
				remStart = tokenStart;
			}
			
```
and remove line 141
```
			remStart = tokenStart;
```

sensors/gps/pa6h.c
```diff
@@ -176,19 +450,89 @@ static int pa6h_alloc(sensor_info_t *info, const char *args)
...
+	cnt = 0;
+	while (cnt++ < repeat) {
+		pa6h_paramSet(ctx->filedes, PMTK_ID_SET_NMEA_BAUDRATE, PMTK_SET_NMEA_BAUDRATE_115200, 0);
+		usleep(10 * 1000);
+	}
+
+	if (pa6h_serialSetup(ctx->filedes, B115200, NULL) < 0) {
+		printf("%s cannot set baud %d\n", PA6H_STR, 115200);
+		close(ctx->filedes);
+		return -1;
```
kemonats, line 493: I think the restore of the termios structure is missing.

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+	static char *remStart = buf;
+	static int pos = 0;
+	static size_t remLen;
+
+	size_t inboxFill = 0;
+	char *tokenStart = NULL, *tokenEnd = NULL, *nextStartToken, checksum;
+	int i, ret, val, len;
+
+	remLen = &buf[pos] - remStart;
+	pos = 0;
```
kemonats, line 117: Remove this line.
```suggestion
```

sensors/gps/pa6h.c
```diff
@@ -176,19 +450,89 @@ static int pa6h_alloc(sensor_info_t *info, const char *args)
...
+				return err;
+			}
 		}
-		ctx->filedes = open(path, O_RDONLY | O_NOCTTY | O_NONBLOCK);
+	} while (ctx->filedes < 0);
+
+	if (isatty(ctx->filedes) != 1) {
+		fprintf(stderr, "%s %s not a tty\n", PA6H_STR, path);
+		close(ctx->filedes);
+		return -1;
```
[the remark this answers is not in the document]
niewim19: - mIssed free(ctx), thanks for catching!
 - decided on `fprintf` for error signaling

sensors/gps/pa6h.c
```diff
@@ -176,19 +450,89 @@ static int pa6h_alloc(sensor_info_t *info, const char *args)
...
+	if (pa6h_serialSetup(ctx->filedes, B9600, &termBackup) < 0) {
+		printf("%s cannot set baud %d\n", PA6H_STR, 9600);
+		close(ctx->filedes);
+		return -1;
+	}
+
+	/* Bruteforce baudrate 115200. GPS changes baud instantly and we certainly will miss output message (no output check) */
+	cnt = 0;
+	while (cnt++ < repeat) {
+		pa6h_paramSet(ctx->filedes, PMTK_ID_SET_NMEA_BAUDRATE, PMTK_SET_NMEA_BAUDRATE_115200, 0);
```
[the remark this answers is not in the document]
niewim19: Initially i thought that baudrate setting message is acknowledged by gps receiver but we are unable to catch it. However I have found in one documentation that it is not. This allows for function refactor so that `ack` is returned via an argument, not a return value. Return value signals OS related error in transmission.

Now return value is checked.

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+
+
+/*
+* Sends PMTK message of `msgId` id and preformated `payload`  to `fd` and then listens for ackownelgment.
+* Timeouts after `timeout` microseconds passed.
+*/
+static int pa6h_paramSet(int fd, unsigned int msgId, const char *payload, time_t timeout)
+{
+	/* Default inter-read sleep set to 100 ms */
+	static const time_t pollSleep = 10 * 1000;
```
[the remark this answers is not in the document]
niewim19: Wrong comment. Corrected for 10ms

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+	pa6h_msg_t inbox[INBOX_SIZE];
+	time_t startTime, currTime;
+	int i, capacity, pos, ack = -1;
+
+	/* Compose message */
+	pos = snprintf(buf, sizeof(buf), "%s%03i,%s\r\n", PMTK_HEADER, msgId, payload);
+	if (pos < 0 || buf[pos - 1] != '\n') {
+		fprintf(stderr, "%s cannot compose message\n", PA6H_STR);
+		return -1;
+	}
```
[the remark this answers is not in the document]
niewim19: Corrected, thanks

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+	if (tcflush(fd, TCIOFLUSH) < 0) {
+		printf("%s tcflush failed.\n", PA6H_STR);
+		return -1;
+	}
+
+	/* Send message */
+	if (write(fd, buf, pos) < 0) {
+		printf("%s msg write failed.\n", PA6H_STR);
+		return -1;
+	}
```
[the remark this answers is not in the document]
niewim19: More safe, thanks! Implemented.

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+	fsync(fd);
+	tcdrain(fd);
+
+	/* Poll for ack */
+	gettime(&startTime, NULL);
+	currTime = startTime;
+	ack = -1;
+	while (currTime - startTime <= timeout && ack < 0) {
+		capacity = pa6h_receiver(fd, inbox, INBOX_SIZE);
```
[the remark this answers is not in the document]
niewim19: Implemented together with ack parameter addition.

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+	char checksum;
+
+	if (bufSz <= PMTK_ACK_MSG_LEN || msgId > 999 || msgId == 0) {
+		return -1;
+	}
+
+	/* validate general static tokens. Assures that atoi/strtod will work on data chunks of correct length/semantics */
+	if (buf[0] != '$' || buf[8] != ',' || buf[12] != ',' || buf[14] != '*') {
+		return -1;
+	}
```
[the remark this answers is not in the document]
niewim19: Dropped this check. This is now checked by `PMTK_ACK_HEADER` from your further remark.

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+
+	/* checksum validation */
+	val = strtol(&buf[15], (char **)NULL, 16);
+	if ((val == 0 && errno == EINVAL) || val > 0xff) {
+		return -1;
+	}
```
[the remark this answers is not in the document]
niewim19: - thanks for catching redundant checksum check. Removed.
 - changed to strtoul
 - well yeah... there is issue with errno setting by strtoul. But i would left this errno check even if it is now pointless as its the issue that should be resolved and not the code that should be amended to wrong-working function :)

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+
+	if (checksum != val) {
+		printf("%x %x\n", checksum, val);
+		return -1;
+	}
+
+	/* ACK message header check */
+	if (strncmp(&buf[0], PMTK_HEADER, sizeof(PMTK_HEADER) - 1) != 0) {
+		return -1;
+	}
```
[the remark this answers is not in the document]
niewim19: Implemented, but with additional "$" check. It is somehow redundant to prior `pa6h_receive` check, but more atomic: this function checks the full pmtk message integrity now. No matter if it would be used in `pa6h_receive` or not

sensors/gps/pa6h.c
```diff
@@ -41,8 +52,274 @@ typedef struct {
...
+	if (val != msgId) {
+		return -1;
+	}
+
+	/* status check */
+	val = strtol(&buf[13], (char **)NULL, 10);
+	if ((val == 0 && errno == EINVAL) || val > PMTK_ACK_SUCCESS) {
+		printf("%d %d\n", val, errno);
+		return -1;
+	}
```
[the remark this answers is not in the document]
niewim19: Decided to use `switch` on `buf[13]` ad this function now returns `enum pmtk_ack` value.

sensors/gps/pa6h.c
niewim19: Yes, you are correct. Complete message that is transmitted include `<CR>` or `<NL>` at the end but they bring no additional  value to both message payload and checksum validation. They are mainly in place so that gps output is correctly formatted when straight printed into terminal.

That is why I decided to just discard their presence in message validation. Lack of <CR> or <NL> would lead to scrambling, or carry into next `pa6h_receive()` even though it is meaningfully correct.

sensors/gps/pa6h.c
niewim19: Implemented, but moved `break` to the following form:
```c
				tokenEnd = tokenStart + 1;
				checksum = 0;
				while ((*tokenEnd != '\0') && (*tokenEnd != '*')) {
					checksum ^= *tokenEnd;
					tokenEnd++;
				}
				if (*tokenEnd == '\0') {
					tokenEnd = NULL;
					break;
				}
```

Its impossible for tokenEnd to be NULL by itself with only addition.

sensors/gps/pa6h.c
niewim19: I did not implement this. For buffer integrity we must recalculate tokenEnd and tokenStart up to the end of data read. Increased the inbox size to 64 by default.

sensors/gps/pa6h.c
niewim19: Yes, we will loose data if our reads/processing are not as frequent as necessary, Other approach would be to carry _over-inbox_ data to the next read. But this just moves the overload point from PA6H driver to tty driver.

sensors/gps/pa6h.c
niewim19: We don't want that. This way we would miss valid payload in the following sequence: `$broken_payload$valid_payload*`

sensors/gps/pa6h.c
niewim19: Oh, good catch. Thanks!

niewim19:
@kemonats Thanks for thorough reviews :) I

 have implemented majority of remarks. The checksum validation shift greatly increased readibility. Some remarks were left unimplemented and left on unresolved.

sensors/gps/pa6h.c
niewim19: Ok, I had to revert this implementation. Calculating actual checksum here does not handle the case where the input buffer looks like this:
```
$xxx$yyyyyyyy* 
```
where `$xxx` is only part of the message and `$yyyyyyyy*` is a valid message as then we calculate checksum once for the whole `$xxx$yyyyyyyy*` which is wrong.

niewim19:
I have reverted the checksum calculation place in code to the one prior to review. Proposed version did not handle the case when in buffer there was partially read message  (without *) and fully read one immediately after. Checksum was calculated for the two-message string at once which was wrong.

sensors/gps/pa6h.c
niewim19: This setting is needed here. Otherwise it should be in the `else` part of the next `if` statement after this line.

sensors/gps/pa6h.c
```diff
@@ -17,32 +17,295 @@
...
+
+			/* Ensure space for checksum and nul character not interfering with next message. Lack '/r/n' at the end is discarded */
+			len = ((nextStartToken == NULL) ? &buf[pos] : nextStartToken) - tokenEnd;
+			if (len < 3) {
+				break;
+			}
+
+			/* Read and validate checksum. Save message to inbox if there is space there */
+			val = strtol(tokenEnd + 1, (char **)NULL, 16);
+			if (!(val == 0 && errno == EINVAL) && val <= 0xff && inboxFill < inboxSz) {
```
agkaminski, line 161: errno is not thread safe by default yet - use `_errno_new` at the start of this thread

niewim19: I did not know that, thanks! Addressed this issue in separate commit.

agkaminski (approved):
LGTM

kemonats:
You merged this PR, but few comments:
1. the proposed version of checksum calculation was good, but you need to change the code litle.
2. the checksum can be equal to `0`.
example from PA6H datasheet:
```
$GPGSA,A,3,29,21,26,15,18,09,06,10,,,,,2.32,0.95,2.11*00
```
3. if record is correct and there is a next record  `nextStartToken != NULL`, then the second time you seek for the `$` char that `nextStartToken` points to.

niewim19:
> You merged this PR, but few comments:

I am eager to clean up the code for phoenix-pilot and not to rely on multiple rebased branches. Nothing is lost. If we work out some ideas here they can be added to the awaiting PR for pa6h in devices :)

___

> 1. the proposed version of checksum calculation was good, but you need to change the code litle.

Of course there was nothing bad with your checksum calculation method. It could be left in the place proposed by you but would need a code change so that we are sure there are no `$` between current `tokenStart` and `tokenEnd` (reverse search from `tokenEnd` found, or creeping towards `tokenEnd` looking for next `$` as long as we don't pass `tokenEnd`.

___

> 2. the checksum can be equal to `0`. example from PA6H datasheet:
> $GPGSA,A,3,29,21,26,15,18,09,06,10,,,,,2.32,0.95,2.11*00

Hmmm... do you refer to this place in code?
``` c
if (!(val == 0 && errno == EINVAL) && val <= 0xff && inboxFill < inboxSz) {
```
This allows for checksum to be equal 0. Correct me please if i missed something.

___


> 3. if record is correct and there is a next record  `nextStartToken != NULL`, then the second time you seek for the `$` char that `nextStartToken` points to.

Not in one particular case: when there are two consecutive messages where the first is truncated/overwritten by the second. Now we look for the `nextStartToken` starting from `tokenEnd`, not `tokenStart`. Given data read as below (`<CR>` and `<NL>` ommited in example):
```
$GPRMC,203522.00,A,5109.026$GPGSA,A,3,29,21,26,15,18,09,06,10,,,,,2.32,0.95,2.11*00$GPGGA
```
We would skip the whole and valid GPGSA message and go directly for GPGGA (with current approach). This skip resulted in poor wakeup time of the device driver as parameter setting function truncates the buffer so the situation presented above occurs with the PMTK ack message being skipped which results in timeouts and retries.  

## Conclusion

Indeed  checksum validation can be utilized in the place proposed by you and in the form proposed by you with the following addition: during checksum calculation any occurrence of `*` should stop the checksum validation, and any occurrence of `$` should change the position of `tokenStart` and `remStart` and reset the checksum to 0.

I will check this approach in the currently active pa6h PR. Feel free to elaborate :)

kemonats:
As for the checksum you are right, it can be zero, I was mistaken, but in the implementation of the `strtol(`) function in libphoenix the `errno` variable is not set.
I noticed two more problems:
1. possible writing outside the buffer,
2. if there is a `nul` `'\0'` character in the loaded data, some or even all of the loaded data may be lost.
To consider using `memcmp()` instead of `strchr()`.

niewim19:
> As for the checksum you are right, it can be zero, I was mistaken, but in the implementation of the `strtol(`) function in libphoenix the `errno` variable is not set. I noticed two more problems:
> 
>   1. possible writing outside the buffer,
>   2. if there is a `nul` `'\0'` character in the loaded data, some or even all of the loaded data may be lost.
>        To consider using `memcmp()` instead of `strchr()`.

As this PR is merged let`s migrate this discussion here:
 - https://github.com/phoenix-rtos/phoenix-rtos-project/issues/612

PR #344, opened 2023-02-24
sensors/gps: increase accuracy of lat/lon with `double`/`int64_t` type

niewim19:
## Description

 - Using `double` variable type for calculations on latitude and longitude to not loose accuracy.
 - Using `int64_t` as transfer data type for longitude and latitude to not loose accuracy
 - Transferring latitude and longitude in nanodegrees (10^9 * [deg]). This gives worst case scenario (on equator) accuracy of 100µm/LSB which is far more that necessary. 

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/sensors/gps/nmea.h
+++ b/sensors/gps/nmea.h
@@ -61,8 +61,8 @@ enum nmea_fields_vtg { field_vtg_track = 1, field_vtg_tracktype = 2, field_vtg_s
 
 /* GPGGA nmea type definitions: Global Positioning System Fix Data  */
 typedef struct {
-	float lat;
-	float lon;
+	double lat;
+	double lon;
 	unsigned int fix;
 	unsigned int sats;
 	float hdop;
--- a/sensors/gps/pa6h.c
+++ b/sensors/gps/pa6h.c
@@ -313,8 +313,8 @@ int pa6h_update(nmea_t *message, sensor_event_t *evtGps)
 	/* timestamp update happens only on GPGGA message as it contains position */
 	switch (message->type) {
 		case nmea_gga:
-			evtGps->gps.lat = message->msg.gga.lat * 1e7;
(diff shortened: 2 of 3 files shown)


kemonats:
Just what position accuracy does the receiver have?

niewim19:
> Just what position accuracy does the receiver have?

Currently implemented PA6H has measurement decimal precision of 10^-4 arc minute, translating to ~18cm on equator. However some COTS modules (ublox NEO series) can provide decimal precision of 10^-7 arc minute which translates to 0.2mm on equator.

When it comes to accuracy of currently implemented pa6h module it is stated in datasheet that accuracy (CEP) is 3m for position and 0.1 m/s for velocity.

agkaminski (approved):
LGTM

kemonats:
The accuracy of the 3m position is about 16 * 10^-4 minutes of arc on the meridian. that is, the value read from the receiver is +- 16 * 10^-4 arc minutes.
I wonder what the position accuracy is with these 10^-7 arc minutes.

niewim19:
> The accuracy of the 3m position is about 16 * 10^-4 minutes of arc on the meridian. that is, the value read from the receiver is +- 16 * 10^-4 arc minutes. I wonder what the position accuracy is with these 10^-7 arc minutes.

Currently i work with PA6H and I can tell that this 3m CEP estimate is pretty bloated :) I cant tell for the error (how far is the reported position from the true position) but speaking spread-wise it is quite easy to achieve a position spread below ~1m with just enough satelites. And if PA6H catches DGPS fix type then the full precision of returned NMEA is used as the spread is minimal - hopping on the last decimal place and giving +-15cm position changes when standing still.

The NMEA messages with 10^-7 arc minute precision gives the minimal step of about 1mm. It works for example in geodetic uses (using RTCM/RTK) where position estimation is important down to centimeters.

PR #348, opened 2023-03-17
simsensor reader: add queue and file reader for sensor event

PNieck:
## Description

This PR contains parser for test files prepared for a future tests, which uses sensors. Topic was discussed there: https://github.com/phoenix-rtos/phoenix-rtos-project/discussions/626

## Motivation and Context

Easier testing.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (list targets here). `armv7a9-zynq7000-qemu`, `host-generic-pc`

## Checklist:

- [x] My change requires a change to the documentation.

## Special treatment

### DIFF
--- a/sensors/simsensor_common/event_queue.c
+++ b/sensors/simsensor_common/event_queue.c
@@ -1,0 +1,98 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Queue for buffering sensor event data from reader to simsensors
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Piotr Nieciecki
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include "event_queue.h"
+
+#include <stdlib.h>
+
+
+int eventQueue_init(event_queue_t *q, unsigned int cnt)
+{
+	if (cnt == 0 || q == NULL) {
+		return -1;
+	}
+
+	q->queue = malloc(sizeof(sensor_event_t) * cnt);
+	if (q->queue == NULL) {
+		return -1;
+	}
+
+	q->capacity = cnt;
+	q->len = 0;
+
+	q->head = q->queue;
+	q->tail = q->queue;
+
+	return 0;
+}
+
+
+void eventQueue_free(event_queue_t *q)
+{
+	if (q != NULL) {
+		free(q->queue);
+		q->queue = NULL;
+	}
+}
+
+
+static inline sensor_event_t *eventQueue_next(event_queue_t *q, sensor_event_t *elem)
+{
+	return (elem == q->queue + q->capacity - 1) ? q->queue : elem + 1;
+}
+
+
+int eventQueue_enqueue(event_queue_t *q, const sensor_event_t *event)
+{
+	if (q == NULL || event == NULL) {
+		return -1;
+	}
+
+	if (q->len == q->capacity) {
+		return -1;
+	}
+
+	*q->head = *event;
+	q->head = eventQueue_next(q, q->head);
+	q->len++;
+
+	return 0;
+}
+
+
+bool eventQueue_full(const event_queue_t *q)
+{
+	if (q == NULL) {
+		return true;
+	}
+
+	return q->len >= q->capacity;
+}
+
+
+int eventQueue_dequeue(event_queue_t *q, sensor_event_t *event)
+{
+	if (q == NULL || event == NULL) {
+		return -1;
+	}
+
+	if (q->len == 0) {
+		return -1;
+	}
+
+	*event = *q->tail;
+	q->tail = eventQueue_next(q, q->tail);
+	q->len--;
+
+	return 0;
+}
--- a/sensors/simsensor_common/event_queue.h
+++ b/sensors/simsensor_common/event_queue.h
@@ -1,0 +1,53 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Queue for buffering sensor event data from reader to simsensors
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Piotr Nieciecki
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#ifndef EVENT_QUEUE_H
+#define EVENT_QUEUE_H
+
+#include <stdbool.h>
+#include <time.h>
+
+#include <libsensors.h>
+
+
+typedef struct {
+	sensor_event_t *queue;
+	unsigned int capacity; /* queue max length */
+	unsigned int len;      /* current length of queue */
+
+	sensor_event_t *head;
+	sensor_event_t *tail;
+} event_queue_t;
+
+
+/* Initiates `q` as an event queue. If succeeded returns 0. In other case -1. */
+extern int eventQueue_init(event_queue_t *q, unsigned int cnt);
+
+
+/* Deallocates queue pointed by `q` */
+extern void eventQueue_free(event_queue_t *q);
+
+
+/* Adds new sensor event to queue. If succeeded returns 0. In other case -1. */
+extern int eventQueue_enqueue(event_queue_t *q, const sensor_event_t *event);
+
+
+/* Returns information if the queue `q` is full. If `q` is NULL, then the behaviour is undefined. */
+extern bool eventQueue_full(const event_queue_t *q);
+
+
+/* Removes the last element from queue `q` and saves it into `event`. Returns 0 on success, or -1 if queue is empty */
+extern int eventQueue_dequeue(event_queue_t *q, sensor_event_t *event);
+
+
+#endif
--- a/sensors/simsensor_common/simsensor_reader.c
+++ b/sensors/simsensor_common/simsensor_reader.c
@@ -1,0 +1,286 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Simsensor reader reads and parses data form special scenario file
+ * than it pushes it to event_queue
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Piotr Nieciecki
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <limits.h>
+#include <sys/types.h>
+#include <string.h>
+#include <sys/time.h>
+
+#include "simsensor_reader.h"
+
+#define READER_END_SCENARIO_INDICATOR NO_SENSOR
+#define READER_DATA_SEPARATOR         ','
+
+#define READER_TIMESTAMP_NOT_SET (time_t) LLONG_MAX
+
+
+static int reader_getFieldLLong(const char *actField, char **nextField, long long *res)
+{
+	char *endptr;
+	long long value = strtoll(actField, &endptr, 10);
+	if (!(*endptr == READER_DATA_SEPARATOR || *endptr == '\0' || *endptr == '\n') || actField == endptr) {
+		printf("simsensor: Invalid file\n");
+		return -1;
+	}
+
+	if (*endptr == READER_DATA_SEPARATOR && nextField != NULL) {
+		*nextField = endptr + 1;
+	}
+
+	*res = value;
+
+	return 0;
+}
+
+
+static int reader_accelDataParse(char *startOfDataSection, time_t timestamp, sensor_event_t *result)
+{
+	char *actField, *nextField;
+	long long tmp;
+
+	actField = startOfDataSection;
+
+	result->type = SENSOR_TYPE_ACCEL;
+	result->timestamp = timestamp;
+
+	if (reader_getFieldLLong(actField, &nextField, &tmp) < 0) {
+		return -1;
+	}
+	result->accels.accelX = tmp;
+	actField = nextField;
+
+	if (reader_getFieldLLong(actField, &nextField, &tmp) < 0) {
+		return -1;
+	}
+	result->accels.accelY = tmp;
+	actField = nextField;
+
+	if (reader_getFieldLLong(actField, NULL, &tmp) < 0) {
+		return -1;
+	}
+	result->accels.accelZ = tmp;
+
+	return 0;
+}
+
+
+static int reader_baroDataParse(char *startOfDataSection, time_t timestamp, sensor_event_t *result)
+{
+	char *actField, *nextField;
+	long long tmp;
+
+	actField = startOfDataSection;
+
+	result->type = SENSOR_TYPE_BARO;
+	result->timestamp = timestamp;
+
+	if (reader_getFieldLLong(actField, &nextField, &tmp) < 0) {
+		return -1;
+	}
+	result->baro.pressure = tmp;
+	actField = nextField;
+
+	if (reader_getFieldLLong(actField, NULL, &tmp) < 0) {
+		return -1;
+	}
+	result->baro.temp = tmp;
+
+	return 0;
+}
+
+
+int reader_read(simsens_reader_t *rd, event_queue_t *queue)
+{
+	char *actField, *nextField;
+	long long tmp;
+	int sensorID, err = 0, emptyIter = 100;
+	ssize_t lineLen;
+	time_t timestamp, timeStart = READER_TIMESTAMP_NOT_SET;
+	sensor_event_t parsed;
+
+	if (rd == NULL || queue == NULL) {
+		return -1;
+	}
+
+	while ((eventQueue_full(queue) == 0) && (emptyIter > 0)) {
+		lineLen = getline(&rd->lineBuf, &rd->bufLen, rd->scenarioFile);
+		if (lineLen == -1) {
+			if (!feof(rd->scenarioFile)) {
+				err = -1;
+				break;
+			}
+
+			clearerr(rd->scenarioFile);
+
+			/* Going to the beginning of the file and omitting file header */
+			if (fseek(rd->scenarioFile, rd->headerLen, SEEK_SET) != 0) {
+				err = -1;
+				break;
+			}
+			emptyIter--;
+
+			continue;
+		}
+
+		if (reader_getFieldLLong(rd->lineBuf, &nextField, &tmp) < 0) {
+			err = -1;
+			break;
+		}
+		sensorID = tmp;
+		actField = nextField;
+
+		/* Checking if we are parsing our header */
+		if ((sensorID & rd->sensorTypes) == 0) {
+			if (sensorID == READER_END_SCENARIO_INDICATOR) {
+				parsed.type = READER_STOP_EVENT;
+				if (eventQueue_enqueue(queue, &parsed) < 0) {
+					err = -1;
+				}
+
+				break;
+			}
+			emptyIter--;
+
+			continue;
+		}
+
+		if (reader_getFieldLLong(actField, &nextField, &tmp) < 0) {
+			err = -1;
+			break;
+		}
+		timestamp = tmp;
+		actField = nextField;
+
+		/* Increasing timeOffset on scenario timestamp decrease (as on file loop) */
+		if (timestamp < rd->timeLast && rd->timeLast != READER_TIMESTAMP_NOT_SET) {
+			rd->timeOffset += rd->timeLast - timestamp;
+		}
+		rd->timeLast = timestamp;
+		timestamp += rd->timeOffset;
+
+		if (timeStart == READER_TIMESTAMP_NOT_SET) {
+			timeStart = timestamp;
+		}
+
+		if (timestamp - timeStart > rd->timeHorizon) {
+			/* Go to start of the line */
+			if (fseek(rd->scenarioFile, -lineLen, SEEK_CUR) != 0) {
+				err = -1;
+			}
+			break;
+		}
+
+		switch (sensorID) {
+			case SENSOR_TYPE_ACCEL:
+				err = reader_accelDataParse(actField, timestamp, &parsed);
+				break;
+			case SENSOR_TYPE_BARO:
+				err = reader_baroDataParse(actField, timestamp, &parsed);
+				break;
+			default:
+				fprintf(stderr, "%s: Unknown sensor type: %d\n", __FUNCTION__, sensorID);
+				err = -1;
+				break;
+		}
+
+		if (err != 0) {
+			break;
+		}
+
+		if (eventQueue_enqueue(queue, &parsed) < 0) {
+			err = -1;
+			break;
+		}
+	}
+
+	return err;
+}
+
+
+void reader_close(simsens_reader_t *reader)
+{
+	if (reader == NULL) {
+		return;
+	}
+
+	if (reader->scenarioFile != NULL) {
+		fclose(reader->scenarioFile);
+		reader->scenarioFile = NULL;
+	}
+
+	free(reader->lineBuf);
+	reader->lineBuf = NULL;
+}
+
+
+int reader_open(simsens_reader_t *rd, const char *path, int sensorTypes, time_t timeHorizon)
+{
+	ssize_t lineLen;
+	time_t act_time;
+
+	if (rd == NULL || path == NULL) {
+		return -1;
+	}
+
+	rd->scenarioFile = fopen(path, "r");
+	if (rd->scenarioFile == NULL) {
+		return -1;
+	}
+
+	rd->lineBuf = NULL;
+	rd->bufLen = 0;
+	rd->sensorTypes = sensorTypes;
+	rd->timeHorizon = timeHorizon;
+	rd->timeLast = READER_TIMESTAMP_NOT_SET;
+	rd->headerLen = 0;
+
+	lineLen = getline(&rd->lineBuf, &rd->bufLen, rd->scenarioFile);
+	if (lineLen < 1) {
+		fclose(rd->scenarioFile);
+		return -1;
+	}
+
+	/* Checking if current line is header */
+	if (rd->lineBuf[0] == 's' || rd->lineBuf[0] == 'S') {
+		/* Read next line */
+		if (getline(&rd->lineBuf, &rd->bufLen, rd->scenarioFile) < 1) {
+			free(rd->lineBuf);
+			fclose(rd->scenarioFile);
+			return -1;
+		}
+		rd->headerLen = lineLen;
+	}
+
+	/* Omitting first sensorID, to get first event timestamp for timeOffset */
+	char *p = strchr(rd->lineBuf, READER_DATA_SEPARATOR);
+	if (p == NULL || reader_getFieldLLong(p + 1, NULL, &rd->timeOffset) != 0) {
+		free(rd->lineBuf);
+		fclose(rd->scenarioFile);
+		return -1;
+	}
+
+	gettime(&act_time, NULL);
+	rd->timeOffset = act_time - rd->timeOffset;
+
+	if (fseek(rd->scenarioFile, rd->headerLen, SEEK_SET) != 0) {
+		free(rd->lineBuf);
+		fclose(rd->scenarioFile);
+		return -1;
+	}
+
+	return 0;
+}
--- a/sensors/include/libsensors.h
+++ b/sensors/include/libsensors.h
@@ -20,6 +20,7 @@
 
 /* Available sensor types */
 
+#define NO_SENSOR         0
 #define SENSOR_TYPE_ACCEL (1 << 0)
 #define SENSOR_TYPE_BARO  (1 << 1)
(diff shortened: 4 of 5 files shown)


sensors/simsensor_common/event_queue.c
```diff
@@ -0,0 +1,118 @@
...
+
+event_queue_t *eventQueue_init(unsigned int elem_cnt)
+{
+	event_queue_t *result;
+
+	if (elem_cnt == 0) {
+		return NULL;
+	}
+
+	result = malloc(sizeof(event_queue_t));
```
niewim19, line 37: The `event_queue_t` structure can be passed as an argument (pointer). No need to allocate it dynamically

sensors/simsensor_common/event_queue.c
```diff
@@ -0,0 +1,118 @@
...
+
+
+static sensor_event_t *eventQueue_nextElement(event_queue_t *q, sensor_event_t *elementPtr)
+{
+	if (elementPtr == q->queue + q->queue_sz) {
+		return q->queue;
+	}
+
+	return elementPtr + 1;
+}
```
niewim19, line 76: This is more of a code remark.

Let's try to keep things simple. More verbose naming should be used when the goal of the named object is not easily understandable from the context. For example: here we are in the queue module and we have a function that accepts some pointer and returns some pointer to the container type with the word `next` in the name. It is easily understandable that it may _return the next pointer to the passed pointer_. How about that:
```suggestion
static inline sensor_event_t *eventQueue_next(event_queue_t *q, sensor_event_t *elem)
{
	return (elem == q->queue + q->queue_sz) ? q->queue : elem + 1;
}
```
Shorter names benefit with overall shorter line lengths which helps readibility and allow for less lines in total without loosing clarity :)

sensors/simsensor_common/event_queue.c
```diff
@@ -0,0 +1,118 @@
...
+
+#include "event_queue.h"
+
+#include <stdlib.h>
+
+
+struct event_queue_t {
+	sensor_event_t *queue;
+	unsigned int queue_sz;
+	unsigned int act_elem_cnt;
```
niewim19, line 22: How about:
```suggestion
	sensor_event_t *queue;
	unsigned int capacity;   /* queue max length */
	unsigned int len;        /* current length of queue */
```
This way it conforms to out coding standard (camelCase) and is easily understandable later in code.

sensors/simsensor_common/event_queue.c
```diff
@@ -0,0 +1,118 @@
...
+	q->act_elem_cnt++;
+
+	return 0;
+}
+
+
+inline int eventQueue_canEnqueue(const event_queue_t *q)
+{
+	return q->act_elem_cnt != q->queue_sz;
+}
```
niewim19, line 100: Just a suggestion. I think this may be more clear later in implementation:
```suggestion
inline bool eventQueue_full(const event_queue_t *q)
{
	return q->act_elem_cnt >= q->queue_sz;
}
```

Also misra suggests boolean type output to be stored in boolean type variable.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,280 @@
...
+}
+
+
+static inline int reader_isLineHeader(const char *line)
+{
+	return line[0] == 's' || line[0] == 'S';
+}
+
+
+static long int reader_getFieldLong(const char *actField, char **nextField, int *error)
```
niewim19, line 94: More typical approach: let`s return the error and use pointer argument as the place to store the read value. This way you will evade situations like one couple lines below where you both: set error to -1 and return -1:
```suggestion
static int reader_getFieldLong(const char *fld, char **nextFld, long int *res)
```
btw - `long int` in our system is the length of `int`. How about using specific value length types from `stdint`: `int32`/`int64`/`uint64`? These are the storage type specifiers in `libsensors.h`.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,280 @@
...
+	simsens_time_t time_struct;
+};
+
+
+static void reader_timeInit(simsens_time_t *timeStruct)
+{
+	timeStruct->first_scenario_time = READER_TIMESTAMP_NOT_SET;
+	timeStruct->last_scenario_timestamp = READER_TIMESTAMP_NOT_SET;
+	timeStruct->loops_cnt = 0;
+}
```
niewim19, line 52: Implementing this function directly in the `init` does not impact clarity of code much. Let`s use it there

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,280 @@
...
+			break;
+		}
+
+		if (eventQueue_enqueue(queue, &parsed_result) < 0) {
+			free(lineBuf);
+			return -1;
+		}
+	}
+
+	free(lineBuf);
```
niewim19, line 277: This free suggests that we are doing dynamic allocation at each `read()` call (actually - `getline` does). We cannot escape the dynamic allocation completely, but we can not do it with every read if we store the line (and size of it) pointer in read structure then getline will only reallocate it if necessary.

We can also pre-allocate it to some expected length so that it is clear who allocates it first (`reader_init`/`reader_open`) and who frees it (`reader_done`/`reader_close`)

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,280 @@
...
+
+	return err < 0 ? err : 0;
```
niewim19, line 279: The operators precedence should be made explicit when it may be interpreted wrongfully 
```suggestion
	return (err < 0) ? err : 0;
```

niewim19 (changes requested):
- queue implementation needs some minor touches
 - reader implementation needs to have timestamp recalculation logic revised and has some memory management issues

sensors/simsensor_common/simsensor_reader.c
PNieck: Ok, but in this approach I have to define separate function for every type, which have to be parsed

PNieck: But, am also using it for parsing other fields, especially these specific for sensor event and there is many different like `int16`, `uint16`, `uint32`, `int32`, `int64` and so on

niewim19: So `int64` should be wide enough to cover all mentioned cases. The problem is only with sufficiently big `time_t` which is `unsigned long long`.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,280 @@
...
+
+#define READER_ACCEL_DEV_ID 1
+#define READER_BARO_DEV_ID  2
+
+
+typedef struct {
+	time_t first_scenario_time;
+	time_t last_scenario_timestamp;
+	int loops_cnt;
+} simsens_time_t;
```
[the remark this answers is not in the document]
PNieck: Ok, but this `offset` is equal to scenario timestamp span times number of full loops. So we still have to count loops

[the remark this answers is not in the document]
niewim19: On timestamp decrement we have to adjust the `offset` by the difference the timestamp has decremented.  This offsets the `offset` by the time span of scenario file.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,297 @@
...
+				err = -1;
+				break;
+			}
+
+			if (reader_skipHeader(rd) != 0) {
+				err = -1;
+				break;
+			}
+
+			continue;
```
niewim19, line 233: This `continue` has potential of endlessly spin the loop if there is some error in read and no data is pushed to queue. Some cap at failed reads is necessary.

Note: MISRA demands loops not bounded to be bounded with some cap

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,297 @@
...
+	if (lineLen < 1) {
+		return -1;
+	}
+
+	/* Checking if current line is header */
+	if (rd->lineBuf[0] == 's' || rd->lineBuf[0] == 'S') {
+		return 0;
+	}
+
+	return fseek(rd->scenario_file, 0, SEEK_SET);
```
niewim19, line 62: Is this function suppose to skip the header or only assert it? Currently on success of header check we will go back to the file start, possibly before the header.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,297 @@
...
+			break;
+		}
+
+
+		if ((sensorID & rd->sensor_types) == 0) {
+			if (sensorID == READER_END_SCENARIO_INDICATOR) {
+				break;
+			}
+
+			continue;
```
niewim19, line 249: The cap is also usable here - each time we pass through any `continue` some counter should be increased and if it reaches some error cap value we exit with `-1`

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,297 @@
...
+
+		if (timestamp - first_timestamp > rd->time_horizon) {
+			/* Go to start of the line */
+			if (fseek(rd->scenario_file, -curLineLen, SEEK_CUR) != 0) {
+				err = -1;
+			}
+			break;
+		}
+
+		switch (sensorID) {
```
niewim19, line 273: Please add `default` condition to this `switch`. (MISRA)

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,297 @@
...
+			case SENSOR_TYPE_ACCEL:
+				if (reader_accelDataParse(actField, timestamp, &parsed_result) < 0) {
+					err = -1;
+				}
```
niewim19, line 277: I think that these parsers offer return code `0/-1` by design. This can be used as `err` setting.

niewim19 (changes requested):
Only part of changes was implemented, but reviewed as requested.

As for timestamp recalculation scheme I propose the following solution
 - add variables to reader: 
   - `timeLast`: holds the last read timestamp from file
   - `offset`: holds current time offset between local time and file time (may be negative)
 - opening the reader will set these values to:
   - `lastTime`: TIME_MAX (or just `~0`) 
   - `offset`: this needs to be preread from file and set to the difference between current time and the first file time.
 - during reading from file:
   - if the timestamp read is greater than the `timeLast` shift the timestamp by `offset` and push to queue
   - if the timestamp read is smaller than the `timeLast` we increase the `offset` by the difference between `lastTime` and read value, shift the timestamp with the new `offset` and push to queue

I think this covers edge cases of file looping or sudden timestamp decrease. Alternatively we can perform such offset jump if and only if reader reaches `EOF` and discard all decrements inside the file.

sensors/simsensor_common/simsensor_reader.c
PNieck: It suppose to skip the header. It gets whole line and checks if it is a header line.
If it is - just returns, so the next read, which uses this file descriptor gets data from the next line. 
In other case goes to the begging of the file.

niewim19: Ok, I have misunderstood it. Looks good.

sensors/simsensor_common/simsensor_reader.c
PNieck: I have added max iterations counter.

niewim19 (changes requested):

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,273 @@
...
+
+
+int reader_read(simsens_reader_t *rd, event_queue_t *queue)
+{
+	char *actField, *nextField;
+	long long tmp;
+	int sensorID, err = 0, remainEmptyInter = 100;
+	size_t curLineLen;
+	time_t timestamp, first_timestamp = READER_TIMESTAMP_NOT_SET;
+	sensor_event_t parsedResult;
```
niewim19, line 178:
```suggestion
	int sensorID, err = 0, emptyIter = 100;
	size_t lineLen;
	time_t timestamp, timeStart = READER_TIMESTAMP_NOT_SET;
	sensor_event_t parsed;
```
 - `first_timestamp` does not follow coding convention
 - `curLineLen`, `remainEmptyInter` and `parsedResult` are a little too long. I gave propositions on shortening them without loosing the meaning. What do you think about these?

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,273 @@
...
+
+		actField = rd->lineBuf;
+
+		if (reader_getFieldLLong(actField, &nextField, &tmp) < 0) {
+			err = -1;
+			break;
+		}
+		sensorID = tmp;
+
+
```
niewim19, line 212: empty line
```suggestion
```

niewim19 (changes requested):
Structurally looks good to me. I have left some editorial remarsks

niewim19 (commented):
Ok, there is one more thing to correct in reader.

`simsens_reader_t:offset` is of type `time_t` which is unsigned. It will not handle well situation where data read from file have correct timestamps which are ahead of current time. In such scenario reader will offset data by `0` and push to queue. Usage of `long long` is required.

Generally speaking `time_t` signage is implementation defined. Better to assume it is unsigned and explicitly make it signed if necessary. 64 bit long can hold 5.9*10^4 years (in µs) so loosing half of it in representation is negligible.

sensors/simsensor_common/event_queue.c
```diff
@@ -0,0 +1,94 @@
...
+
+	*q->head = *event;
+	q->head = eventQueue_next(q, q->head);
+	q->len++;
+
+	return 0;
+}
+
+
+inline bool eventQueue_full(const event_queue_t *q)
```
kemonats, line 73:
```suggestion
bool eventQueue_full(const event_queue_t *q)
```

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,270 @@
...
+#include <stdlib.h>
+#include <ctype.h>
+#include <stdbool.h>
+
+#include "simsensor_reader.h"
+
+#define READER_END_SCENARIO_INDICATOR NO_SENSOR
+#define READER_DATA_SEPARATOR         ','
+
+#define READER_TIMESTAMP_NOT_SET -1
```
kemonats, line 25:
```suggestion
#define READER_TIMESTAMP_NOT_SET (time_t)-1
```

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,270 @@
...
+
+
+static int reader_skipHeader(simsens_reader_t *rd)
+{
+	size_t lineLen = getline(&rd->lineBuf, &rd->bufLen, rd->scenarioFile);
```
kemonats, line 30:
```suggestion
	ssize_t lineLen = getline(&rd->lineBuf, &rd->bufLen, rd->scenarioFile);
```
https://github.com/phoenix-rtos/libphoenix/blob/a514d66740f94232373d32ee7bb8d4ceef92ed9e/include/stdio.h#L360

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,270 @@
...
+	result->baro.pressure = tmp;
+	actField = nextField;
+
+	if (reader_getFieldLLong(actField, &nextField, &tmp) < 0) {
+		return -1;
+	}
+	result->baro.temp = tmp;
+	actField = nextField;
+
+	return 0;
```
kemonats, line 167:
```suggestion
	result->baro.temp = tmp;

	return 0;
```

sensors/simsensor_common/simsensor_reader.c
PNieck: I have decided to use `(time_t)LONG_MAX`, because time_t can be unsigned.

sensors/simsensor_common/event_queue.c
```diff
@@ -0,0 +1,94 @@
...
+	q->head = eventQueue_next(q, q->head);
+	q->len++;
+
+	return 0;
+}
+
+
+bool eventQueue_full(const event_queue_t *q)
+{
+	return q->len >= q->capacity;
```
kemonats, line 75: What if `q == NULL`

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,298 @@
...
+#include <limits.h>
+#include <stdbool.h>
+#include <sys/types.h>
+
+#include "simsensor_reader.h"
+
+#define READER_END_SCENARIO_INDICATOR NO_SENSOR
+#define READER_DATA_SEPARATOR         ','
+
+#define READER_TIMESTAMP_NOT_SET (time_t) LONG_MAX
```
kemonats, line 27: Not too small a number?

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,298 @@
...
+
+
+static int reader_getFieldLLong(const char *actField, char **nextField, long long *res)
+{
+	char *endptr;
+	long long value = strtoll(actField, &endptr, 10);
+	if (!(*endptr == READER_DATA_SEPARATOR || *endptr == '\0' || *endptr == '\n') || actField == endptr) {
```
kemonats, line 34: Can a field contain only spaces? If so, you need to add an additional code.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,298 @@
...
+int reader_read(simsens_reader_t *rd, event_queue_t *queue)
+{
+	char *actField, *nextField;
+	long long tmp;
+	int sensorID, err = 0, emptyIter = 100;
+	ssize_t lineLen;
+	time_t timestamp, timeStart = READER_TIMESTAMP_NOT_SET;
+	sensor_event_t parsed;
+
+	while (!eventQueue_full(queue) && emptyIter > 0) {
```
kemonats, line 114: What if `rd == NULL` or `queue == NULL`.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,298 @@
...
+				err = -1;
+				break;
+			}
+			clearerr(rd->scenarioFile);
+
+			/* Omitting file header */
+			if (fseek(rd->scenarioFile, rd->headerLen, SEEK_CUR) != 0) {
+				err = -1;
+				break;
+			}
```
kemonats, line 133: Only one `fseek()`.
```suggestion
			clearerr(rd->scenarioFile);
			/* Going to the beginning of the file and omitting file header */
			if (fseek(rd->scenarioFile, rd->headerLen, SEEK_SET) != 0) {
				err = -1;
				break;
			}
```

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,298 @@
...
+		fclose(rd->scenarioFile);
+		return -1;
+	}
+	rd->timeOffset = -rd->timeOffset;
+
+	if (fseek(rd->scenarioFile, -readBytes, SEEK_CUR) != 0) {
+		free(rd->lineBuf);
+		fclose(rd->scenarioFile);
+		return -1;
+	}
```
kemonats, line 295: My suggestion, a bit long.
```suggestion
	rd->headerLen = 0;
	/* Checking if current line is header */
	if (rd->lineBuf[0] == 's' || rd->lineBuf[0] == 'S') {
		/* Read next line */
		if (getline(&rd->lineBuf, &rd->bufLen, rd->scenarioFile) < 1) {
			free(rd->lineBuf);
			fclose(rd->scenarioFile);
			return -1;
		}
		rd->headerLen = lineLen;
	}

	/* Omitting first sensorID */
	char *p = strchr(rd->lineBuf, READER_DATA_SEPARATOR);
	if ((p == NULL) ||  (reader_getFieldLLong(rd->lineBuf, NULL, &rd->timeOffset) != 0) {
		free(rd->lineBuf);
		fclose(rd->scenarioFile);
		return -1;
	}
	rd->timeOffset = -rd->timeOffset;

	if (fseek(rd->scenarioFile, rd->headerLen, SEEK_SET) != 0) {
		free(rd->lineBuf);
		fclose(rd->scenarioFile);
		return -1;
	}
```
In addition:
```c
#include <string.h>
```
remove variable `read_bytes`

sensors/simsensor_common/simsensor_reader.c
kemonats: Not too small a number?

PNieck: I think, that our purposes it's enough

niewim19: kemonats is right on this: LONG_MAX = INT_MAX = 2^31 ~= 2e8 µs, which is about half a hour. Better use LLONG_MAX.

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,278 @@
...
+			break;
+		}
+		sensorID = tmp;
+		actField = nextField;
+
+		/* Checking if we are parsing our header */
+		if ((sensorID & rd->sensorTypes) == 0) {
+			if (sensorID == READER_END_SCENARIO_INDICATOR) {
+				break;
+			}
```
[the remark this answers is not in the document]
PNieck: I have add returning special stop event

niewim19 (dismissed):
LGTM

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,284 @@
...
+		fclose(rd->scenarioFile);
+		return -1;
+	}
+
+	/* Checking if current line is header */
+	if (rd->lineBuf[0] == 's' || rd->lineBuf[0] == 'S') {
+		/* Read next line */
+		if (getline(&rd->lineBuf, &rd->bufLen, rd->scenarioFile) < 1) {
+			fclose(rd->scenarioFile);
+			return -1;
```
kemonats, line 262: Missing `free()`.
```suggestion
			free(rd->lineBuf);
			fclose(rd->scenarioFile);
			return -1;
```
If you leave `Free()` and `fclose()` without setting:
```c
	reader->scenarioFile = NULL;
	reader->lineBuf = NULL;
```
then `flose()` and `free()` can be called again in the `reader_close()` function.
I have an issue regarding `getline()`. In which repository to put it libphoenix or phoenix-rtos-project?

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,284 @@
...
+		}
+		rd->headerLen = lineLen;
+	}
+
+	/* Omitting first sensorID, to get first event timestamp for timeOffset */
+	char *p = strchr(rd->lineBuf, READER_DATA_SEPARATOR);
+	if (p == NULL || reader_getFieldLLong(p + 1, NULL, &rd->timeOffset) != 0) {
+		fclose(rd->scenarioFile);
+		return -1;
```
kemonats, line 271:
```suggestion
		free(rd->lineBuf);
		fclose(rd->scenarioFile);
		return -1;
```

sensors/simsensor_common/simsensor_reader.c
niewim19: Please report it in [phoenix-rtos-project issues](https://github.com/phoenix-rtos/phoenix-rtos-project/issues).

sensors/simsensor_common/event_queue.h
```diff
@@ -0,0 +1,53 @@
...
+	unsigned int capacity; /* queue max length */
+	unsigned int len;      /* current length of queue */
+
+	sensor_event_t *head;
+	sensor_event_t *tail;
+} event_queue_t;
+
+
+/* Initiates `q` as an event queue. If succeeded returns 0. In other case -1. */
+extern int eventQueue_init(event_queue_t *q, unsigned int cnt);
```
agkaminski, line 34: This function is never used, seem like an oops

sensors/simsensor_common/event_queue.c
```diff
@@ -0,0 +1,98 @@
...
+#include <stdlib.h>
+
+
+int eventQueue_init(event_queue_t *q, unsigned int cnt)
+{
+	if (cnt == 0 || q == NULL) {
+		return -1;
+	}
+
+	q->queue = malloc(sizeof(sensor_event_t) * cnt);
```
agkaminski, line 25: If we can live with the cost of per object malloc imho list is a better choice for this

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,286 @@
...
+
+#define READER_TIMESTAMP_NOT_SET (time_t) LLONG_MAX
+
+
+static int reader_getFieldLLong(const char *actField, char **nextField, long long *res)
+{
+	char *endptr;
+	long long value = strtoll(actField, &endptr, 10);
+	if (!(*endptr == READER_DATA_SEPARATOR || *endptr == '\0' || *endptr == '\n') || actField == endptr) {
+		printf("Wrong file\n");
```
agkaminski, line 36:
```suggestion
		printf("simsensor: Invalid file\n");
```

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,286 @@
...
+
+	actField = startOfDataSection;
+
+	result->type = SENSOR_TYPE_BARO;
+	result->timestamp = timestamp;
+
+	if (reader_getFieldLLong(actField, &nextField, &tmp) < 0) {
+		return -1;
+	}
+	result->baro.pressure = tmp;
```
agkaminski, line 94: Why do we need `tmp`? On error we leave `result` struct destroyed anyway (e.g. first field is ok, next one is not). Applies to other places too

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,286 @@
...
+	int sensorID, err = 0, emptyIter = 100;
+	ssize_t lineLen;
+	time_t timestamp, timeStart = READER_TIMESTAMP_NOT_SET;
+	sensor_event_t parsed;
+
+	if (rd == NULL || queue == NULL) {
+		return -1;
+	}
+
+	while (!eventQueue_full(queue) && emptyIter > 0) {
```
agkaminski, line 119: MISRA
```suggestion
	while ((eventQueue_full(queue) == 0) && (emptyIter > 0)) {
```

sensors/simsensor_common/simsensor_reader.c
```diff
@@ -0,0 +1,286 @@
...
+			break;
+		}
+
+		if (eventQueue_enqueue(queue, &parsed) < 0) {
+			err = -1;
+			break;
+		}
+	}
+
+	return err;
```
agkaminski, line 210: No check if we exit early because of full queue

agkaminski (commented):
I don't really like omnipresent NULL checks - why do we expect function user to pass NULL anyway?

sensors/simsensor_common/event_queue.h
PNieck: Function will be used in modules called `simsensors`, which will be added in the future. Without this function you cannot use whole queue

agkaminski: Ack

sensors/simsensor_common/simsensor_reader.c
PNieck: We use `tmp`, because `reader_getFieldLLong` takes pointer to `long long` as an argument and data what we want to read is not always a `long long`. For example there `baro.pressure` is `unit32_t`

sensors/simsensor_common/simsensor_reader.c
PNieck: Because exiting early is not considered as an error

sensors/simsensor_common/event_queue.c
PNieck: I will stick to this implementation. For our use case queue with constant size is enough. Moreover we are concerned about performance, some sensors can report new measurements really often.

agkaminski: Ack

sensors/simsensor_common/simsensor_reader.c
agkaminski: Perhaps this field should be read a `long` instead of `long long` then, but ack

agkaminski (approved):
NULL checks aren't a problem for me, they don't hurt anything for sure (perhaps performance anc code size, marginally), but they also tend to be not needed and represent dead code path in the final code

PNieck:
I have cleared the commit history

niewim19 (approved):

PR #368, opened 2023-05-10
imxrt117x-cm4: small improvements

ziemleszcz:
- add custom vector table offset
- reset M4 core only

JIRA: NIL-107

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (rt1176).

## Checklist:

## Special treatment

https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/176

### DIFF
--- a/cpu/imxrt117x-cm4/imxrt117x-cm4.c
+++ b/cpu/imxrt117x-cm4/imxrt117x-cm4.c
@@ -33,7 +33,7 @@ enum { atr0 = 0, atr1, atr2, atr3, arr0, arr1, arr2, arr3, asr, acr };
 enum { chan0 = 0, chan1, chan2, chan3, chanLen };
 
 
-enum { scr_scr = 0, scr_smsr };
+enum { src_scr = 0, src_srmr };
 // clang-format on
 
 
@@ -58,7 +58,7 @@ struct {
 	volatile void *m4memory;
 	size_t m4memorysz;
 
-	volatile uint32_t *scr;
+	volatile uint32_t *src;
 	volatile uint32_t *mu;
 
 	struct {
@@ -210,10 +210,10 @@ static ssize_t chanWrite(id_t id, const void *buff, size_t bufflen)
 }
 
 
-static int runCore(void)
+static int runCore(unsigned int offset)
 {
 	platformctl_t pctl;
-	static const unsigned int vectors = (unsigned int)M4_MEMORY_START + 0x100;
+	unsigned int vectors = (unsigned int)M4_MEMORY_START + offset;
 
 	/* Set CM4's VTOR to the start of it's memory */
 	pctl.action = pctl_set;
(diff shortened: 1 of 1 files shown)


cpu/imxrt117x-cm4/imxrt117x-cm4.c
```diff
@@ -259,6 +259,7 @@ static int loadFromFile(const char *path)
 		ptr += r;
 		total += r;
 	}
+	// FIXME: clean & invalidate dcache
```
agkaminski, line 262:
```suggestion
	/* FIXME: clean & invalidate dcache */
```

cpu/imxrt117x-cm4/imxrt117x-cm4.c
```diff
@@ -270,6 +271,7 @@ static int loadFromBuff(const void *buff, size_t bufflen)
 		bufflen = m4_common.m4memorysz;
 
 	memcpy((void *)m4_common.m4memory, buff, bufflen);
+	// FIXME: clean & invalidate dcache
```
agkaminski, line 274:
```suggestion
	/* FIXME: clean & invalidate dcache */
```

cpu/imxrt117x-cm4/imxrt117x-cm4.c
agkaminski, line 262:
```suggestion
	/* FIXME: clean & invalidate dcache */
```

agkaminski (commented):
Nitpick only

cpu/imxrt117x-cm4/imxrt117x-cm4.c
```diff
@@ -233,9 +233,9 @@ static int runCore(void)
 		return -EIO;
 
 	/* Release the Kraken */
-	*(m4_common.scr + scr_smsr) |= 0xf << 10;
+	*(m4_common.src + src_srmr) |= 0xf << 10;
```
gerard5, line 236: If the M7 core is the primary `LR` role and the M4 is the secondary `NLR` role, this needs to be changed.
Setting bits 12-13 (M7REQ_RESET_MODE) to "do not reset anything" prevents the NVIC from resetting the M7 core (it will hang, due to the "do not reset anything" bits set), the reboot command will not work (it will freeze the core until valid reset from other source comes, you will be able to do a reset only via other enabled sources TEMP, OVERVOLT, CSU, WDOG, JTAG because these bits are 00 = "reset system - both M4 and M7").

So, set 10-11 (M4REQ_RESET_MODE) just to prevent M4 core from resetting both cores (suspend M4 during reset), and leave 12-13 bits as 0 to allow M7 core to reset system (as M7 is the Legally Relevant core)
```suggestion
	*(m4_common.src + src_srmr) |= 0x3u << 10;
```
and/or it may be required to set M4_LOCKUP_RESEST_MODE (bits 5-6) not to reset the system (both cores) during M4 lockup. The M7 core (as LR core) can reset the M4 core (NLR core) as needed:
```suggestion
	*(m4_common.src + src_srmr) |= (0x3u << 10) | (0x3u << 6);
```

PR #377, opened 2023-05-31
remove LIBS sets before calling static-lib.mk

xvuko:
LIBS is not used in static-lib and this value can corrupt next binary.mk

This needs to be merged before https://github.com/phoenix-rtos/phoenix-rtos-build/pull/144


nalajcie (approved):
shame we can't express cross-repo dependencies on other libs (their header files), maybe in monorepo approach we would reintroduce it as `DEPS`. LGTM.

PR #381, opened 2023-06-02
spi: low volume transmision optimization

niewim19:
JIRA: PP-152

## Description

- use `raw` message memory if sufficient to store data
- don`t use `condWait` if transmission is below the threshold

## Motivation and Context

When data to be transferred and to be received are below `raw` message storage then it is better to use this storage for faster processing time.

If expected data is below some arbitrary threshold then do not invoke `condWait` as it is faster to just busy-wait for completion. 

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


niewim19:
This solution is currently implemented only in `libzynq7000-xspi`. It needs implementation in other drivers

niewim19:
Closing, as superseeded with:
 - https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/383

PR #382, opened 2023-06-06
imxrt117x: Move cm4 driver to imxrt-multi

agkaminski:
JIRA: NIL-428

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work (https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/180).


multi/imxrt-multi/cm4.c
```diff
@@ -107,10 +110,11 @@ static inline int fifo_popPacket(packet_t *packet, fifo_t *fifo)
 }
 
 
-static inline int fifo_push(char byte, fifo_t *fifo)
+static inline int fifo_push(unsigned char byte, fifo_t *fifo)
 {
-	if ((fifo->wptr + 1) % sizeof(fifo->buff) == fifo->wptr)
+	if (((fifo->wptr + 1) % sizeof(fifo->buff)) == fifo->wptr) {
```
gerard5, line 115: Doing `(fifo->wptr + 1) % sizeof(fifo->buff)` twice is not effective, maybe use temporary variable to store the `nextwptr`.

multi/imxrt-multi/cm4.c
```diff
@@ -123,14 +127,17 @@ static inline int fifo_pushPacket(packet_t *packet, fifo_t *fifo)
...
 		size = sizeof(packet->byte.payload);
-	else
+	}
+	else {
 		size = packet->byte.bno;
+	}
 
 	for (i = 0; i < size; ++i) {
-		if (!fifo_push(packet->byte.payload[i], fifo))
+		if (!fifo_push(packet->byte.payload[i], fifo)) {
```
gerard5, line 138:
```suggestion
		if (fifo_push(packet->byte.payload[i], fifo) == 0) {
```
because that PR changes the style later in the code anyway.

multi/imxrt-multi/cm4.c
```diff
@@ -169,68 +178,75 @@ static int mu_irqHandler(unsigned int n, void *arg)
...
-	char *tbuff = buff;
-	ssize_t i;
+	unsigned char *tbuff = buff;
+	size_t i;
 	fifo_t *fifo;
 
-	if (id < chan0 || id > chan3 || buff == NULL)
+	if ((id < chan0) || (id > chan3) || (buff == NULL)) {
 		return -1;
+	}
```
gerard5, line 187: `return -EINVAL;`, instead of `-1`.

multi/imxrt-multi/cm4.c
```diff
@@ -169,68 +178,75 @@ static int mu_irqHandler(unsigned int n, void *arg)
...
-	const char *tbuff = buff;
-	int i;
+	const unsigned char *tbuff = buff;
+	size_t i;
 	fifo_t *fifo;
 
-	if (id < chan0 || id > chan3 || buff == NULL)
+	if ((id < chan0) || (id > chan3) || (buff == NULL)) {
 		return -1;
+	}
```
gerard5, line 209: `return -EINVAL;`, instead of `-1`.

multi/imxrt-multi/cm4.c
```diff
@@ -243,17 +259,25 @@ static int runCore(unsigned int offset)
...
+	f = fopen(path, "r");
+	if (f == NULL) {
 		return -ENOENT;
+	}
 
-	while ((r = fread(buff, sizeof(buff), 1, f)) > 0 && total < m4_common.m4memorysz) {
-		if (total + r > m4_common.m4memorysz)
+	while (total < m4_common.m4memorysz) {
+		r = fread(buff, sizeof(buff), 1, f);
+		if (r <= 0) {
```
gerard5, line 274: Currently `r < 0` is never true, `r` is declared as `size_t`, when changed to e.g. `ssize_t r;` will be ok.

multi/imxrt-multi/imxrt-multi.h
```diff
@@ -79,27 +85,35 @@ typedef struct {
...
 
 
-typedef struct _multi_i_t {
+typedef struct {
 	id_t id;
 
 	union {
 		gpio_t gpio;
 		spi_t spi;
+		int cm4_type;
```
gerard5, line 108: cm4 is only valid for the imxrt1170 target, maybe something like:
```suggestion
#ifdef TARGET_IMXRT1170
		int cm4_type;
#endif
```
btw. a similar check would be good for all `#define CM4_*` that are specific to the target.

agkaminski: Doesn't really hurt anything, I'd rather keep one, coherent API

multi/imxrt-multi/cm4.c
agkaminski: Fixed, as a side note, there was a nasty bug hiding in a plain sight, fixed ;)

multi/imxrt-multi/cm4.c
agkaminski: `fread()` returns `size_t`, changed condition to `r == 0`

agkaminski:
I also added a lock I forgot about, it's needed as cm4 driver entered multi-threaded environment

multi/imxrt-multi/cm4.c
gerard5: That's right.

multi/imxrt-multi/cm4.c
```diff
@@ -243,17 +264,25 @@ static int runCore(unsigned int offset)
...
-	if ((f = fopen(path, "r")) == NULL)
+	f = fopen(path, "r");
+	if (f == NULL) {
 		return -ENOENT;
+	}
+
+	while (total < m4_common.m4memorysz) {
+		r = fread(buff, sizeof(buff), 1, f);
+		if (r == 0) {
+			break;
```
ziemleszcz, line 280: Return error if not feof()?

agkaminski: Done

multi/imxrt-multi/cm4.c
```diff
@@ -243,17 +264,25 @@ static int runCore(unsigned int offset)
...
+		}
 
-	while ((r = fread(buff, sizeof(buff), 1, f)) > 0 && total < m4_common.m4memorysz) {
-		if (total + r > m4_common.m4memorysz)
+		if (total + r > m4_common.m4memorysz) {
 			r = m4_common.m4memorysz - total;
```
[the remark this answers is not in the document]
agkaminski: Done

multi/imxrt-multi/imxrt-multi.c
```diff
@@ -296,6 +307,14 @@ static int createDevFiles(void)
 		return -1;
 #endif
 
+#if CM4
+	for (i = 0; i < 4; ++i) {
+		sprintf(name, "cpuM4%d", i);
+		if (mkFile(&dir, id_cm4_0 + i - 1, name, multi_port) < 0)
+			return -1;
+	}
```
kemonats, line 315:
```suggestion
	for (i = 0; i < 4; ++i) {
		sprintf(name, "cpuM4%d", i);
		if (mkFile(&dir, id_cm4_0 + i - 1, name, multi_port) < 0) {
			return -1;
		}
	}
```
Since `i < 10` this can be done easily without `sprintf()`.

multi/imxrt-multi/imxrt-multi.h
```diff
@@ -17,14 +17,20 @@
...
 	id_uart9, id_uart10, id_uart11, id_uart12, id_gpio1, id_gpio2, id_gpio3, id_gpio4, id_gpio5,
 	id_gpio6, id_gpio7, id_gpio8, id_gpio9, id_gpio10, id_gpio11, id_gpio12, id_gpio13, id_spi1,
 	id_spi2, id_spi3, id_spi4,
 #ifdef TARGET_IMXRT1170
 	id_spi5, id_spi6,
 #endif
-	id_i2c1, id_i2c2, id_i2c3, id_i2c4, id_trng };
+	id_i2c1, id_i2c2, id_i2c3, id_i2c4, id_trng,
+#ifdef TARGET_IMXRT1170
+	id_cm4_0, id_cm4_1, id_cm4_2, id_cm4_3
```
ziemleszcz, line 30: IDs are not available outside of phoenix-rtos-devices.

multi/imxrt-multi/imxrt-multi.c
```diff
@@ -296,6 +307,14 @@ static int createDevFiles(void)
 		return -1;
 #endif
 
+#if CM4
+	for (i = 0; i < 4; ++i) {
+		sprintf(name, "cpuM4%d", i);
+		if (mkFile(&dir, id_cm4_0 + i - 1, name, multi_port) < 0)
```
ziemleszcz, line 313: `id_cm4_0 + i - 1` -> `id_cm4_0 + i`

multi/imxrt-multi/imxrt-multi.c
```diff
@@ -81,13 +82,23 @@ static void multi_dispatchMsg(msg_t *msg)
...
+			break;
+
 #ifdef TARGET_IMXRT1170
 		case id_spi5:
 		case id_spi6:
-#endif
 			spi_handleMsg(msg, id);
 			break;
 
+		case id_cm4_0:
```
ziemleszcz, line 94: I'm not sure if it's good idea to demux messages based on custom multi ID. Why not use `msg->i.openclose.oid.id` instead? Anyway read() & write() messages will be not delivered because multi ID is not set (it probably applies to TRNG driver as well).

agkaminski: In mtDevCtl we don't have this field - we only have msg.i/o.raw, so we need to pass oid/id by ourself. Oid in common msg field is planned on API change

multi/imxrt-multi/imxrt-multi.c
agkaminski: Wanted to stick to the style of the rest of the file, but fixed (along with other missing brackets)

multi/imxrt-multi/imxrt-multi.h
agkaminski: Wym, they are available, but I agree shouldn't. Removed from API, lookup should be used

multi/imxrt-multi/imxrt-multi.c
agkaminski: Ok. I see what you mean. It worked by a miracle, fixed

multi/imxrt-multi/cm4.c
```diff
@@ -243,32 +264,48 @@ static int runCore(unsigned int offset)
...
+
 	return (int)total;
 }
 
 
 static int loadFromBuff(const void *buff, size_t bufflen)
 {
-	if (bufflen > m4_common.m4memorysz)
+	if (bufflen > m4_common.m4memorysz) {
 		bufflen = m4_common.m4memorysz;
```
ziemleszcz, line 307: Return EINVAL error?

multi/imxrt-multi/cm4.c
```diff
@@ -169,68 +183,75 @@ static int mu_irqHandler(unsigned int n, void *arg)
...
 {
 	platformctl_t pctl;
-	unsigned int vectors = (unsigned int)M4_MEMORY_START + offset;
+	unsigned int vectors = (unsigned int)CM4_MEMORY_START + offset;
 
 	/* Set CM4's VTOR to the start of it's memory */
 	pctl.action = pctl_set;
 	pctl.type = pctl_iolpsrgpr;
 	pctl.iogpr.field = 0;
 	pctl.iogpr.val = vectors & 0xffff;
```
gerard5, line 241: btw. may be corrected, this should be:
```suggestion
	pctl.iogpr.val = vectors & 0xfff8;
```
LPSR_GPR0 bits 2-0 of CM4_INIT_VTOR_LOW are reserved.

multi/imxrt-multi/cm4.c
```diff
@@ -279,127 +316,84 @@ static int loadFromBuff(const void *buff, size_t bufflen)
...
-		/* Unsupported operations */
-		case mtCreate:
-			msg.o.create.err = -ENOSYS;
-			break;
-		case mtLookup:
-			msg.o.lookup.err = -ENOSYS;
-			break;
 		default:
-			msg.o.io.err = -ENOSYS;
+			msg->o.io.err = -EINVAL;
```
gerard5, line 378: Why changed to `EINVAL` (Invalid argument) ?
`ENOSYS` a.k.a function not implemented, for me suits better. Other handlers in imxrt-multi default err is ENOSYS, see gpio, spi.

ziemleszcz (approved):
LGTM

PR #394, opened 2023-07-10
sensors: add lis2mdl magnetometer driver

niewim19:
JIRA: PP-159

## Description

Adding driver for lis2mdl 3 axis +-50gauss magnetometer.

Datasheet: https://www.st.com/resource/en/datasheet/lis2mdl.pdf

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: zturn drone.

## Checklist:

## Special treatment

### DIFF
--- a/sensors/mag/lis2mdl.c
+++ b/sensors/mag/lis2mdl.c
@@ -1,0 +1,227 @@
+/*
+ * Phoenix-RTOS
+ *
+ * Driver for lis2mdl magnetometer
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Mateusz Niewiadomski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <sys/time.h>
+#include <sys/threads.h>
+#include <spi.h>
+#include <sensors-spi.h>
+#include <string.h>
+
+#include "../sensors.h"
+
+/* self-identification register of magnetometer */
+#define REG_WHOAMI 0x4f
+#define VAL_WHOAMI 0x40
+
+#define SPI_READ_BIT 0x80
+
+#define REG_CFG_REG_A               0x60
+#define VAL_CFG_REG_A_COMP_TEMP_EN  0x80
+#define VAL_CFG_REG_A_ODR_10HZ      0x00
+#define VAL_CFG_REG_A_ODR_20HZ      0x04
+#define VAL_CFG_REG_A_ODR_50HZ      0x08
+#define VAL_CFG_REG_A_ODR_100HZ     0x0c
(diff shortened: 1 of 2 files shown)


sensors/mag/lis2mdl.c
```diff
@@ -0,0 +1,229 @@
...
+	sensor_event_t evt;
+	char stack[512] __attribute__((aligned(8)));
+} lis2mdl_ctx_t;
+
+
+static int16_t translateMag(uint8_t hbyte, uint8_t lbyte)
+{
+	int16_t val;
+
+	val = (hbyte << 8) | lbyte;
```
agkaminski, line 68: MISRA problem, not sure about simplest solution to avoid it - this is UB (signed overflow), it works, but it is a rule violation.

Please note that making `lbyte` signed will break the code - during `int8_t` -> `int16_t` cast `lbyte` will be sign extended (so 8 MSB of `val` will become ones and will make `OR` operation no-op)

sensors/mag/lis2mdl.c
```diff
@@ -0,0 +1,229 @@
...
+
+	if (val > MAG_OVERFLOW) {
+		return MAG_OVERFLOW * MAG_CONV_MGAUSS;
+	}
+	else if (val < -MAG_OVERFLOW) {
+		return -MAG_OVERFLOW * MAG_CONV_MGAUSS;
+	}
+
+	/* conversion to milligaus (10^-7 T), as it is already sensor manager storage scale */
+	return val * MAG_CONV_MGAUSS;
```
agkaminski, line 78: MISRA lesser evil fix
```suggestion
	else {
		/* conversion to milligaus (10^-7 T), as it is already sensor manager storage scale */
		return val * MAG_CONV_MGAUSS;
	}
```

sensors/mag/lis2mdl.c
```diff
@@ -0,0 +1,229 @@
...
+	int err;
+	sensor_info_t *info = (sensor_info_t *)data;
+	lis2mdl_ctx_t *ctx = info->ctx;
+	uint8_t ibuf[SENSOR_OUTPUT_SIZE] = { 0 };
+	uint8_t obuf;
+
+	while (1) {
+		usleep(10 * 1000);
+
+		obuf = REG_DATA_OUT | SPI_READ_BIT;
```
agkaminski, line 143: Nitpicky optimization 
```suggestion
	const uint8_t obuf = REG_DATA_OUT | SPI_READ_BIT;

	while (1) {
		usleep(10 * 1000);
```

agkaminski (dismissed):
I'm going to slowly start to torture you with MISRA ;)

sensors/mag/lis2mdl.c
niewim19: I have added a comment that this part does not comply with MISRA. We have discussed other solution but the readability to amount of code ratio of them is very poor comparing to this approach.

agkaminski (approved):

PR #396, opened 2023-07-11
sensors: int32 accelerometer reading

niewim19:
## Description

Change in sensorhub accelerometer reading variable type from `int16_t` to `int32_t`. This increases maximum accelerometer value from small +-3G to more than +-200kG which is much more than sufficient.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: zturn drone.

## Checklist:

## Special treatment

### DIFF
--- a/sensors/imu/lsm9dsxx.c
+++ b/sensors/imu/lsm9dsxx.c
@@ -117,24 +117,12 @@ static int32_t translateGyr(uint8_t hbyte, uint8_t lbyte)
 }
 
 
-static int16_t translateAcc(uint8_t hbyte, uint8_t lbyte)
+static int32_t translateAcc(uint8_t hbyte, uint8_t lbyte)
 {
-	int32_t val = 0;
-
 	/* MISRA incompliant - casting u16 to s16 with no regard to sign */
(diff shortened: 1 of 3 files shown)


sensors/imu/lsm9dsxx.c
```diff
@@ -117,24 +117,12 @@ static int32_t translateGyr(uint8_t hbyte, uint8_t lbyte)
...
-	if (val > INT16_MAX) {
-		return INT16_MAX;
-	}
-	else if (val < INT16_MIN) {
-		return INT16_MIN;
-	}
-	else {
-		return val; /* sensor value to [mm/s^2] */
-	}
+	return ACC8G_CONV_MG * MG_CONV_MMS2 * val; /* sensor value to [mm/s^2] */
```
agkaminski, line 125: I think not strictly needed, but I'm not sure, so it would nice to have
```suggestion
	return ACC8G_CONV_MG * MG_CONV_MMS2 * (int32_t)val; /* sensor value to [mm/s^2] */
```

niewim19: Macros are float, so there is already promotion to float from int16

agkaminski (approved):

PR #405, opened 2023-07-31
gr716: implement multi driver

lukileczo:
JIRA: RTOS-534

## Types of changes

## How Has This Been Tested?

- [x] New test added: https://github.com/phoenix-rtos/phoenix-rtos-tests/pull/232

## Checklist:

## Special treatment

### DIFF
--- a/multi/gr716-multi/adc.c
+++ b/multi/gr716-multi/adc.c
@@ -1,0 +1,269 @@
+/*
+ * Phoenix-RTOS
+ *
+ * GR716 ADC driver
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Lukasz Leczkowski
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <errno.h>
+#include <sys/interrupt.h>
+#include <sys/platform.h>
+#include <sys/types.h>
+#include <sys/threads.h>
+#include <phoenix/arch/gr716.h>
+
+#include "adc.h"
+#include "common.h"
+#include "gr716-multi.h"
+
+
+#define ADC_CNT 8
+
+#define ADC0_BASE ((void *)0x80400000)
+#define ADC1_BASE ((void *)0x80401000)
+#define ADC2_BASE ((void *)0x80402000)
+#define ADC3_BASE ((void *)0x80403000)
+#define ADC4_BASE ((void *)0x80404000)
+#define ADC5_BASE ((void *)0x80405000)
+#define ADC6_BASE ((void *)0x80406000)
+#define ADC7_BASE ((void *)0x80407000)
+
+#define ADC0_IRQ 28
+#define ADC1_IRQ 29
+#define ADC2_IRQ 30
+#define ADC3_IRQ 31
+#define ADC4_IRQ 28
+#define ADC5_IRQ 29
+#define ADC6_IRQ 30
+#define ADC7_IRQ 31
+
+#define ADC_DEFAULT_SR (200 * 1000) /* 200 kSps */
+
+/* ADC registers */
+#define ADC_CTRL      0  /* Control register : 0x00 */
+#define ADC_SAMP_CTRL 1  /* Sampling control register : 0x04 */
+#define ADC_SEQ_CTRL  2  /* Sequence control register : 0x08 */
+#define ADC_SEQ_SYNC  3  /* Sequence sync register : 0x0C */
+#define ADC_STATUS    4  /* Status register : 0x10 */
+#define ADC_IRQ       5  /* Interrupt register : 0x14 */
+#define ADC_IRQ_MSK   6  /* Interrupt mask register : 0x18 */
+#define ADC_PRE_AMP   7  /* Pre-amplifier control register : 0x1C */
+#define ADC_HI_LVL    8  /* High level detection register : 0x20 */
+#define ADC_LO_LVL    9  /* Low level detection register : 0x24 */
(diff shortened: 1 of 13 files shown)


multi/gr716-multi/adc.h
```diff
@@ -0,0 +1,26 @@
...
+#define _MULTI_ADC_H_
+
+
+#include <sys/msg.h>
+
+
+extern void adc_handleMsg(msg_t *msg, int dev);
+
+
+extern int adc_init(void);
```
agkaminski, line 23: Lets phase out these `extern`s
```suggestion
void adc_handleMsg(msg_t *msg, int dev);


int adc_init(void);
```

multi/gr716-multi/adc.c
```diff
@@ -0,0 +1,261 @@
...
+}
+
+
+static void adc_handleDevCtl(msg_t *msg, int dev)
+{
+	multi_i_t *idevctl = (multi_i_t *)msg->i.raw;
+	multi_o_t *odevctl = (multi_o_t *)msg->o.raw;
+
+	dev -= id_adc0;
+	if ((dev >= ADC_CNT) || adc_info[dev].active == 0) {
```
agkaminski, line 164:
```suggestion
	if ((dev < 0) || (dev >= ADC_CNT) || adc_info[dev].active == 0) {
```

multi/gr716-multi/adc.c
```diff
@@ -0,0 +1,261 @@
...
+}
+
+
+void adc_handleMsg(msg_t *msg, int dev)
+{
+	switch (msg->type) {
+		case mtOpen:
+		case mtClose:
+		case mtWrite:
+		case mtRead:
```
agkaminski, line 193: We could handle `mtRead` based on `msg->i.io.oid.id`. Ofc we need a file in /dev per each adc

agkaminski (commented):
Partial review for now

multi/gr716-multi/adc.c
```diff
@@ -0,0 +1,268 @@
+/*
+ * Phoenix-RTOS
+ *
+ * GR716 ADC driver
+ *
+ * Copyright 2023 Phoenix Systems
```
agkaminski, line 6: Author is missing in all files

multi/gr716-multi/common.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+#ifndef _GR716_MULTI_COMMON_H_
+#define _GR716_MULTI_COMMON_H_
+
+
+#include <stdint.h>
+
+
+#define SYSCLK_FREQ 50 * 1000 * 1000
```
agkaminski, line 20:
```suggestion
#define SYSCLK_FREQ (50 * 1000 * 1000)
```

multi/gr716-multi/gpio.c
```diff
@@ -0,0 +1,226 @@
...
+	volatile uint32_t *gpio = gpio_portToAddr(port);
+
+	if (gpio == NULL) {
+		return -EINVAL;
+	}
+
+	mutexLock(gpio_common.lock);
+	clearedMsk = *(gpio + GPIO_OUT) & ~clear;
+	*(gpio + GPIO_DIR) = clearedMsk | set;
+	mutexUnlock(gpio_common.lock);
```
agkaminski, line 107: I guess it's not possible to perform atomic `and` and `or` here (from `common`)? This would allow us to skip mutex

agkaminski: If it would be possible in case of direction too, mutex could be removed altogether

multi/gr716-multi/gr716-multi.c
```diff
@@ -0,0 +1,323 @@
...
+
+			case mtCreate:
+				msg.o.create.err = -ENOSYS;
+				break;
+
+			case mtTruncate:
+			case mtDestroy:
+			case mtLookup:
+			case mtLink:
+			case mtUnlink:
```
agkaminski, line 163: Better to return 0 on link and unlink - if we return error on unlink sad stuff happens when removing special file - error is returned in shell, but the file is still removed from the filesystem

multi/gr716-multi/gr716-multi.c
```diff
@@ -0,0 +1,323 @@
...
+
+	while (1) {
+		while (msgRecv(multi_common.multiOid.port, &msg, &rid) < 0) {
+		}
+
+		switch (msg.type) {
+			case mtRead:
+			case mtWrite:
+			case mtGetAttr:
+			case mtSetAttr:
```
agkaminski, line 145: Do we really handle these messages in `multi_dispatchMsg`?

multi/gr716-multi/spi.c
```diff
@@ -0,0 +1,359 @@
...
+			break;
+	}
+}
+
+
+void spi_handleMsg(msg_t *msg, int dev)
+{
+	switch (msg->type) {
+		case mtGetAttr:
+		case mtSetAttr:
```
agkaminski, line 284: These msg types return error in the different field, applies to adc too

agkaminski (commented):
Sorry for late review 👉👈

multi/gr716-multi/gpio.c
lukileczo: well, I tried to use it but I think it didn't want to work, I'll check again

lukileczo: I couldn't get atomic set & clear to work, but I use gpio logical `or` and `and` registers now

agkaminski (approved):
LGTM

PR #412, opened 2023-08-31
mpu6000: add lpf selection argument

niewim19:
JIRA: PP-171

## Description

Adding argument to driver start so it is selectable. No argument defaults to no filtering. Examples of correct arguments to mpu6000:
| Command                                    | spi device   | ss selection | lpf selection |
|--------------------------------------------|--------------|--------------|---------------|
|   `sbin/sensors -s mpu6000:/dev/spi0.2::3` |  /dev/spi0.2 | none         | 3             |
|   `sbin/sensors -s mpu6000:/dev/spi0.2`    |  /dev/spi0.2 | none         | 0 (default)   |
| `sbin/sensors -s mpu6000:/dev/spi0.2:1`    |  /dev/spi0.2 | 1            | 0 (default)   |
| `sbin/sensors -s mpu6000:/dev/spi0.2:1:4`  | /dev/spi0.2  | 1            | 4             |

## Motivation and Context

mpu6000 has a low pass filter which offers good quality filtering.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/sensors/imu/mpu6000.c
+++ b/sensors/imu/mpu6000.c
@@ -80,6 +80,7 @@ typedef struct {
 	oid_t spiSS;
 	sensor_event_t evtAccel;
 	sensor_event_t evtGyro;
+	uint8_t lpfSel;
 	char stack[512] __attribute__((aligned(8)));
 } mpu6000_ctx_t;
 
@@ -156,7 +157,7 @@ static int mpu6000_hwSetup(mpu6000_ctx_t *ctx)
 	}
 
 	/* Sampling rate and LPF config */
-	if (spiWriteReg(ctx, REG_CONFIG, VAL_CONFIG_DLPF_CFG_256HZ) < 0) {
+	if (spiWriteReg(ctx, REG_CONFIG, ctx->lpfSel) < 0) {
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


sensors/imu/mpu6000.c
```diff
@@ -280,7 +282,21 @@ static int mpu6000_alloc(sensor_info_t *info, const char *args)
...
 	if (ss != NULL) {
 		*(ss++) = '\0';
+
+		lpfChr = strchr(ss, ':');
+		if (lpfChr != NULL) {
+			*(lpfChr++) = '\0';
+
+			errno = EOK;
+			lpfSel = strtoul(lpfChr, NULL, 10);
+			if (lpfSel > 7 || (lpfSel == 0 && errno != EOK)) {
```
agkaminski, line 292: `lpfSel == 0` is somewhat redundant but it gives performance boost (as no access to `errno` is needed when it's not zero). It causes us to miss ERANGE, but lpfSel > 7 protect against it. So the current code is 100%, but errno check should enough. I won't insist a change here. 
```suggestion
			if ((lpfSel > 7) || (errno != EOK)) {
```

agkaminski (approved):

sensors/imu/mpu6000.c
niewim19: You`re right - this can be written shorter. I`ll consider such syntax in future use of `strto*` functions - thanks!. Leaving as is to not force additional changes as already approved :)

PR #421, opened 2023-09-14
imxrt-multi: support for non-blocking operations on CM4 devices

ziemleszcz:
JIRA: NIL-441

Fix non-blocking read() & write() calls.

## Motivation and Context

Non-blocking mode is used by serial device in core-libs.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: (imxrt1176).

## Checklist:

## Special treatment

### DIFF
--- a/multi/imxrt-multi/cm4.c
+++ b/multi/imxrt-multi/cm4.c
@@ -11,6 +11,7 @@
 
 #include <stdio.h>
 #include <errno.h>
+#include <fcntl.h>
 #include <string.h>
 #include <stdint.h>
 #include <sys/msg.h>
@@ -69,6 +70,10 @@ struct {
 	struct {
 		fifo_t rx;
 		fifo_t tx;
+		struct {
+			uint8_t busy : 1;
+			uint8_t nonblock : 1;
+		} state;
 	} channel[chanLen];
 
 	handle_t lock;
@@ -189,6 +194,39 @@ static int mu_irqHandler(unsigned int n, void *arg)
 }
 
 
+static ssize_t chanOpen(id_t id, int flags)
+{
+	if ((id < chan0) || (id > chan3)) {
+		return -EINVAL;
+	}
+
+	if (m4_common.channel[id].state.busy == 1) {
+		return -EBUSY;
+	}
+
+	m4_common.channel[id].state.busy = 1;
+	m4_common.channel[id].state.nonblock = ((flags & O_NONBLOCK) != 0) ? 1 : 0;
+
+	return EOK;
+}
+
+
+static ssize_t chanClose(id_t id)
+{
+	if ((id < chan0) || (id > chan3)) {
+		return -EINVAL;
+	}
+
+	if (m4_common.channel[id].state.busy == 0) {
+		return -EBADF;
+	}
+
+	m4_common.channel[id].state.busy = 0;
+
+	return EOK;
+}
+
+
 static ssize_t chanRead(id_t id, void *buff, size_t bufflen)
 {
 	unsigned char *tbuff = buff;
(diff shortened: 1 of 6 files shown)


multi/imxrt-multi/cm4.c
```diff
@@ -69,6 +70,10 @@ struct {
 	struct {
 		fifo_t rx;
 		fifo_t tx;
+		struct {
+			uint8_t busy : 1;
+			uint8_t nonblock : 1;
+		} state;
```
agkaminski, line 76: As bitfields these might prove dangerous to use without a lock. Perhaps separate `volatile int`s?

multi/imxrt-multi/cm4.c
```diff
@@ -207,6 +245,10 @@ static ssize_t chanRead(id_t id, void *buff, size_t bufflen)
 		}
 	}
 
+	if (i == 0 && m4_common.channel[id].state.nonblock == 1) {
+		i = -EWOULDBLOCK;
+	}
```
agkaminski, line 250: I'm not sure about this, as it's always non-blocking, but if we open with `O_NONBLOCK` then we'll return `EWOULDBLOCK`, otherwise zero anyway... Maybe for now always assume `O_NONBLOCK`? But I guess as a hot-fix you need it's ok.

multi/imxrt-multi/cm4.c
```diff
@@ -233,6 +275,10 @@ static ssize_t chanWrite(id_t id, const void *buff, size_t bufflen)
 		setTXirq(id, 1);
 	}
 
+	if (i == 0 && m4_common.channel[id].state.nonblock == 1) {
+		i = -EWOULDBLOCK;
```
gerard5, line 279: On some systems `EWOULDBLOCK != EAGAIN`, and `EWOULDBLOCK` is used for sockets. In phoenix-rtos [EAGAIN is defined first](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/include/errno.h#L30), and [EWOULDBLOCK is an alias](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/include/errno.h#L59). Quoting the [Open Group](https://pubs.opengroup.org/onlinepubs/009604599/functions/read.html):

normal fd:
2023-09-15-102159_789x44_scrot
pipe:
2023-09-15-102144_955x77_scrot
socket:
2023-09-15-102016_1096x122_scrot

multi/imxrt-multi/cm4.c
```diff
@@ -207,6 +245,10 @@ static ssize_t chanRead(id_t id, void *buff, size_t bufflen)
 		}
 	}
 
+	if (i == 0 && m4_common.channel[id].state.nonblock == 1) {
+		i = -EWOULDBLOCK;
```
gerard5, line 249: On some systems `EWOULDBLOCK != EAGAIN`, and `EWOULDBLOCK` is used for sockets. In phoenix-rtos [EAGAIN is defined first](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/include/errno.h#L30), and [EWOULDBLOCK is an alias](https://github.com/phoenix-rtos/phoenix-rtos-kernel/blob/master/include/errno.h#L59). Quoting the [Open Group](https://pubs.opengroup.org/onlinepubs/009604599/functions/read.html):

normal fd:
2023-09-15-102159_789x44_scrot
pipe:
2023-09-15-102144_955x77_scrot
socket:
2023-09-15-102016_1096x122_scrot

multi/imxrt-multi/cm4.c
ziemleszcz: chanOpen() & chanClose() are protected by common mutex.

multi/imxrt-multi/cm4.c
ziemleszcz: Blocking mode is still used by `cm4-tool` so we can't always return `EWOULDBLOCK`.

multi/imxrt-multi/cm4.c
agkaminski: Right, sorry

multi/imxrt-multi/cm4.c
gerard5: I pointed to POSIX, phoenix-rtos (code-base) aspires to be POSIX compliant.

agkaminski (approved):

PR #426, opened 2023-10-02
imx6ull-uart: drop oldest bytes in case of overrun

badochov:
## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -79,32 +79,36 @@
 #define UCR3_DCD       (1 << 9)
 #define UCR3_DSR       (1 << 10)
 
+
+/* clang-format off */
 enum { urxd = 0, utxd = 16, ucr1 = 32, ucr2, ucr3, ucr4, ufcr, usr1, usr2,
-	uesc, utim, ubir, ubmr, ubrc, onems, uts, umcr };
+       uesc, utim, ubir, ubmr, ubrc, onems, uts, umcr };
```
agkaminski, line 85: This should be simple indentation, no need for alignment

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -455,39 +459,45 @@ static void set_baudrate(void *_uart, speed_t baud)
...
-		if (md % 2 == 0)
-			md = md / 2;
-		else if (in % 2 == 0)
-			in = in / 2;
-		else if (md > in)
+		if ((md % 2) == 0) {
+			md /= 2;
+		}
+		else if ((in % 2) == 0)
+			in /= 2;
```
agkaminski, line 484:
```suggestion
		else if ((in % 2) == 0) {
			in /= 2;
		}
```

nalajcie:
was this tested on DTR project? we're using 460800 baud there with a lot of data to pass both ways, so it's crucial to ensure the implementation performance hasn't degraded.

ziemleszcz:
Note that you can achieve the same result without changing LF-FIFO implementation by popping one element if push fails and pushing again (this is SPSC FIFO).

badochov:
> was this tested on DTR project? we're using 460800 baud there with a lot of data to pass both ways, so it's crucial to ensure the implementation performance hasn't degraded.

No, I'll test it today on DTR

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -254,36 +258,35 @@ static void uart_thr(void *arg)
...
 				unsigned long request;
 				const void *in_data = ioctl_unpack(&msg, &request, NULL);
 				const void *out_data = NULL;
 				pid_t pid = ioctl_getSenderPid(&msg);
 
 				int err = libtty_ioctl(&uart.tty_common, pid, request, in_data, &out_data);
 				ioctl_setResponse(&msg, request, err, out_data);
-			}
-			break;
+			} break;
```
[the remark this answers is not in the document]
badochov: I'll move break into the block

badochov:
> was this tested on DTR project? we're using 460800 baud there with a lot of data to pass both ways, so it's crucial to ensure the implementation performance hasn't degraded.

I've tested this change on nandtest(part of dtr repo) project and this change for some strange reason only have sped up the uart. I've tested this change by sending 95 629 chars at once from pc to dtr, while program reading from tty in raw mode with O_NONBLOCK was executing. I've detected on average ~35 overruns on old code and ~22 on new code, average of 5 runs on each code. Also only overruns I've noticed are libtty buffer overruns which are not part of uart's statistics(maybe they should be?, however change in libtty_putchar api would be needed). Also increase of size of libtty buffer maybe considered. I didn't test sending data from dtr, to host as this change didn't change this part of uart driver.

agkaminski (approved):
LGTM

PR #428, opened 2023-10-05
Add pseudo devices on small targets

gerard5:
## Description

This change adds the standard pseudo devices (as library) for small targets (without using posixsrv)
* `/dev/null`
* `/dev/zero`
* `/dev/full`
* `/dev/urandom`

First, I add the library use-cases to  the`imxrt10xx`, `imxrt117x` and `sparc-gr716` targets, because of the similar implementation of multi-driver.

Should work well with `psh` stream stdio redirections, try e.g.:

```bash
(psh)% dd < /dev/urandom > /dev/full
(psh)% dd count=1024 bs=1024 if=/dev/urandom > /dev/null
(psh)% dd if=/dev/zero of=/dev/null count=1024
```

```bash
(psh)% help &> /dev/null
(psh)% cat notexisting 2> /dev/null
```

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imxrt1176`, `imxrt1064-evk`.

## Checklist:

## Special treatment

### DIFF
--- a/libpseudodev/pseudodev.c
+++ b/libpseudodev/pseudodev.c
@@ -1,0 +1,143 @@
+/*
+ * Phoenix-RTOS
+ *
+ * pseudo devices
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Gerard Swiderski
+ *
+ * %LICENSE%
+ */
+
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/msg.h>
+#include <sys/file.h>
+#include <sys/time.h>
+#include "pseudodev.h"
+
+
+static struct {
+	unsigned int randSeed;
+} pseudo_common;
+
+
+static int pseudo_open(int id, int flags)
+{
+	(void)flags;
+
+	switch (id) {
+		case pseudo_idNull:
+		case pseudo_idZero:
+		case pseudo_idFull:
+		case pseudo_idRandom:
+			return EOK;
+
+		default:
+			break;
+	}
+
+	return -ENODEV;
+}
+
+
+static int pseudo_close(int id)
+{
+	switch (id) {
+		case pseudo_idNull:
+		case pseudo_idZero:
+		case pseudo_idFull:
+		case pseudo_idRandom:
+			return EOK;
+
+		default:
+			break;
+	}
+
+	return -ENODEV;
+}
+
+
+static ssize_t pseudo_read(int id, void *buf, size_t count)
+{
+	switch (id) {
+		case pseudo_idNull:
+			return (ssize_t)0;
+
+		case pseudo_idZero:
+		case pseudo_idFull:
+			memset(buf, 0, count);
+			return (ssize_t)count;
+
+		case pseudo_idRandom:
+			for (size_t ofs = 0; ofs < count; ofs += sizeof(pseudo_common.randSeed)) {
+				int res = rand_r(&pseudo_common.randSeed);
+				memcpy((unsigned char *)buf + ofs, &res,
+					((count - ofs) > sizeof(res)) ? sizeof(res) : count - ofs);
+			}
+			return (ssize_t)count;
+
+		default:
+			break;
+	}
+
+	return -ENODEV;
+}
+
+
+static ssize_t pseudo_write(int id, const void *buf, size_t count)
+{
+	(void)buf;
+
+	switch (id) {
+		case pseudo_idNull:
+		case pseudo_idZero:
+		case pseudo_idRandom:
+			return (ssize_t)count;
+
+		case pseudo_idFull:
+			return -ENOSPC;
+
+		default:
+			break;
+	}
+
+	return -ENODEV;
+}
+
+
+int pseudo_handleMsg(msg_t *msg, int id)
+{
+	switch (msg->type) {
+		case mtOpen:
+			msg->o.io.err = pseudo_open(id, msg->i.openclose.flags);
+			return 0;
+
+		case mtClose:
+			msg->o.io.err = pseudo_close(id);
+			return 0;
+
+		case mtRead:
+			msg->o.io.err = pseudo_read(id, msg->o.data, msg->o.size);
+			return 0;
+
+		case mtWrite:
+			msg->o.io.err = pseudo_write(id, msg->i.data, msg->i.size);
+			return 0;
+
+		default:
+			break;
+	}
+
+	return -ENOSYS;
+}
+
+
+void pseudo_init(void)
+{
+	time_t now;
+	gettime(&now, NULL);
+
+	pseudo_common.randSeed = (unsigned int)now;
+}
--- a/multi/gr716-multi/gr716-multi.c
+++ b/multi/gr716-multi/gr716-multi.c
@@ -28,6 +28,10 @@
 #include <sys/threads.h>
 #include <posix/utils.h>
 
+#if PSEUDODEV
+#include <pseudodev.h>
+#endif
+
 #include "adc.h"
 #include "gpio.h"
 #include "spi.h"
@@ -45,7 +49,11 @@
 #define LOG_ERROR(fmt, ...) fprintf(stderr, "gr716-multi: " fmt "\n", ##__VA_ARGS__)
 
 
-static const char *multi_devs[] = { "gpio0", "gpio1", "spi0", "spi1", "uart2", "adc0", NULL };
+static const char *multi_devs[] = { "gpio0", "gpio1", "spi0", "spi1", "uart2", "adc0",
+#if PSEUDODEV
+	"null", "zero", "full", "urandom",
+#endif
+	NULL };
 
 static struct {
 	oid_t multiOid;
@@ -90,6 +98,23 @@ static id_t multi_getId(msg_t *msg)
 	return id;
 }
 
+#if PSEUDODEV
+static inline int multi2pseudo(id_t id)
+{
+	switch (id) {
+		default:
+		case id_pseudoNull:
+			return pseudo_idNull;
+		case id_pseudoZero:
+			return pseudo_idZero;
+		case id_pseudoFull:
+			return pseudo_idFull;
+		case id_pseudoRandom:
+			return pseudo_idRandom;
+	}
+}
+#endif
+
 
 static void multi_dispatchMsg(msg_t *msg)
 {
@@ -116,6 +141,17 @@ static void multi_dispatchMsg(msg_t *msg)
 		case id_adc7:
 			adc_handleMsg(msg, id);
 			break;
+
+#if PSEUDODEV
+		case id_pseudoNull:
+		case id_pseudoZero:
+		case id_pseudoFull:
+		case id_pseudoRandom:
+			if (pseudo_handleMsg(msg, multi2pseudo(id)) < 0) {
+				msg->o.io.err = -EPERM;
+			}
+			break;
+#endif
 
 		default:
 			msg->o.io.err = -EINVAL;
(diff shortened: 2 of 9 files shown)


nalajcie:
just an idea: maybe we could compile posixsrv into lib to be reused in multidrv on small platforms? That way we could have single implementation and possibly introduce other functionalities (pipes, event queues, pseudo-terminals) easily if needed in the future?

libpseudodev/pseudodev.c
```diff
@@ -0,0 +1,190 @@
...
+			msg->o.io.err = pseudo_write(id, msg->i.data, msg->i.size);
+			break;
+
+		case mtGetAttr:
+			msg->o.attr.err = pseudo_getattr(id, msg->i.attr.type, &msg->o.attr.val);
+			break;
+
+		case mtSetAttr:
+			msg->o.attr.err = -EINVAL;
+			break;
```
agkaminski, line 177: These stuff shouldn't be needed - stat() uses file oid, not device one. On the other hand, mtLink and mtUnlink handlers are needed to avoid this issue again: https://github.com/phoenix-rtos/phoenix-rtos-project/issues/718

libpseudodev/pseudodev.c
```diff
@@ -0,0 +1,190 @@
...
+			msg->o.io.err = -EINVAL;
+			break;
+	}
+}
+
+
+void pseudo_init(oid_t oid)
+{
+	pseudo_common.oid = oid;
+	pseudo_common.randSeed = 0xcafebeba;
```
agkaminski, line 189: Perhaps something gettime based? This will result in very predictable "random" pattern

libpseudodev/pseudodev.c
```diff
@@ -0,0 +1,190 @@
...
+#include "pseudodev.h"
+
+
+static struct {
+	oid_t oid;
+	unsigned int randSeed;
+} pseudo_common;
+
+
+static int pseudo_open(int id, int flags)
```
agkaminski, line 26: Please note that oid->id is not int. It's platform dependent (at least for now): https://github.com/phoenix-rtos/libphoenix/blob/2a948484a6414a9363c6c33646d517f26d2b9d73/include/arch/armv7/types.h#L27C1-L27C1

Best solution imho is to pass whole oid here to make code oid-id-type-agnostic

multi/gr716-multi/gr716-multi.h
```diff
@@ -21,7 +21,8 @@
 
 /* clang-format off */
 /* IDs of special files OIDs */
-enum { id_gpio0 = 0, id_gpio1, id_spi0, id_spi1, id_uart0, id_uart1, id_uart2, id_uart3, id_uart4,
+enum { id_pseudoNull = 0, id_pseudoZero, id_pseudoFull, id_pseudoRandom,
```
agkaminski, line 24: It's a good idea to add new stuff at the end, to preserve ABI compatibility.

I hope that gpio doesn't assume somewhere `id_gpio0 = 0` (I don't think it does though)

multi/imxrt-multi/imxrt-multi.c
```diff
@@ -49,6 +51,8 @@
 
 #define STACKSZ 512
 
+#define MODE_DEFAULT 0666
```
agkaminski, line 54: We already have this: https://github.com/phoenix-rtos/libphoenix/blob/2a948484a6414a9363c6c33646d517f26d2b9d73/include/sys/stat.h#L76

multi/imxrt-multi/imxrt-multi.c
```diff
@@ -258,125 +270,138 @@ static int createDevFiles(void)
...
+		return -1;
+	}
+
+	if (mkFile(&dir, id_pseudoZero, "zero", multi_port, MODE_DEFAULT) < 0) {
+		return -1;
+	}
+
+	if (mkFile(&dir, id_pseudoFull, "full", multi_port, MODE_DEFAULT) < 0) {
+		return -1;
+	}
```
agkaminski, line 285: Why isn't pseudodev lib optional? We might want to use posixsrv instead

multi/imxrt-multi/imxrt-multi.c
```diff
@@ -258,125 +270,138 @@ static int createDevFiles(void)
...
 
 	/* UARTs */
 
 #if UART1
-	if (mkFile(&dir, id_uart1, "uart1", common.uart_port) < 0) {
+	if (mkFile(&dir, id_uart1, "uart1", common.uart_port, MODE_DEFAULT) < 0) {
```
agkaminski, line 290: I'm not sure there's a point in passing mode as an argument - it's default all the way

gerard5:
> just an idea: maybe we could compile posixsrv into lib to be reused in multidrv on small platforms? That way we could have single implementation and possibly introduce other functionalities (pipes, event queues, pseudo-terminals) easily if needed in the future?

I also thought about it at the beginning, but I also wanted the pseudo devices to be on stm32l4 (there is only 360kB of RAM). Pseudo devices in this PR are lightweight, they do not use additional threads or reserve buffers.
I agree that on `imxrt1176` can try with `(lib)posixsrv`, there is a good chance it will fit, on `imxrt1064` maybe, but on `imxrt1052` there is no chance (just like with stm32).
Of course, in each case, assuming that at least one additional custom user's app should fit (be able to malloc/mmap, etc) of course.

lukileczo:
I think it _should_ work on GR716 but, as @agkaminski said, I think it'd be better to add the new `id_`s at the end (even though it's not assumed that e.g. `id_gpio0 = 0`).

multi/gr716-multi/gr716-multi.h
gerard5: Yup, I see that in case of (gr716-multi) devices are created using string LUT `multi_devs[]`, and the last name is `adc0` but id's are `id_adc0-7`. So, I wanted that name match its id (I was unsure `adc0` some gap, and then pseudo devices - also I can't test it on target). In case of imxrt-multi I was able to easly add stuff at the end, as you write.

lukileczo: Yeah, but the `multi_devs[]` LUT is only used for creating dev files, it's not indexed by `id_` in any place - and e.g. not all ADCs are here because of pin confilcts if we were to enable them. So I think it's safe to just add the strings in the LUT directly after `"adc0"`

lukileczo: Also there's already no mapping in `id` :arrow_right: `multi_devs[]` - with `uart` and `console`

libpseudodev/pseudodev.c
gerard5: Right. I'll add a translator for `pseudo_handleMsg()` per multidrv implementation, but I'll keep the `id` as an `int` (it is treated here as a minor index of the device, not `id_t`).

multi/gr716-multi/gr716-multi.h
gerard5: but (just wondering) from the other hand these pseudo devices are available always ("common" posix devices), while gpios, adcs, etc may be setup/sorted differently. So, it is legit to put them (common) first, so later (common) ABI does not change.

agkaminski: We don't want pseudo devices to be available always - to be able to use posixsrv if we decide to

gerard5: ack

libpseudodev/pseudodev.c
gerard5: Done.

libpseudodev/pseudodev.c
gerard5: Removed. Now for all unsupported message types the multidrv is resonsible.

libpseudodev/pseudodev.c
gerard5: Done.

multi/imxrt-multi/imxrt-multi.c
gerard5: Done.

multi/imxrt-multi/imxrt-multi.c
gerard5: Removed, but in `mkFile(..)` impl. I replaced default `0` (no-access) value with  [DEFFILEMODE](https://github.com/phoenix-rtos/libphoenix/blob/2a948484a6414a9363c6c33646d517f26d2b9d73/include/sys/stat.h#L76) (all rw).

multi/imxrt-multi/imxrt-multi.c
gerard5: Thanks for it.

libpseudodev/pseudodev.c
```diff
@@ -0,0 +1,143 @@
...
+			return (ssize_t)0;
+
+		case pseudo_idZero:
+		case pseudo_idFull:
+			memset(buf, 0, count);
+			return (ssize_t)count;
+
+		case pseudo_idRandom:
+			for (size_t ofs = 0; ofs < count; ofs += sizeof(pseudo_common.randSeed)) {
+				rand_r(&pseudo_common.randSeed);
```
agkaminski, line 75: Nitpick - return value should be used, not seed

multi/gr716-multi/gr716-multi.c
```diff
@@ -90,6 +98,23 @@ static id_t multi_getId(msg_t *msg)
...
+	switch (id) {
+		default:
+		case id_pseudoNull:
+			return pseudo_idNull;
+		case id_pseudoZero:
+			return pseudo_idZero;
+		case id_pseudoFull:
+			return pseudo_idFull;
+		case id_pseudoRandom:
+			return pseudo_idRandom;
```
agkaminski, line 113: Nitpick: default should be last
```suggestion
		case id_pseudoZero:
			return pseudo_idZero;
		case id_pseudoFull:
			return pseudo_idFull;
		case id_pseudoRandom:
			return pseudo_idRandom;
		case id_pseudoNull:
		default:
			return pseudo_idNull;
```

libpseudodev/pseudodev.c
```diff
@@ -0,0 +1,143 @@
...
+
+		case mtWrite:
+			msg->o.io.err = pseudo_write(id, msg->i.data, msg->i.size);
+			return 0;
+
+		default:
+			break;
+	}
+
+	return -ENOSYS;
```
agkaminski, line 133: Why not `msg->o.io.err = -ENOSYS;`? API is inconsistent and forces outside correction.
We can handle mtGet/SetAttr and mtDevCtl separately to be super correct

multi/gr716-multi/gr716-multi.c
gerard5: Rule 16.5 - (required) - A default label shall appear as either the first or the last switch label of a switch statement.

libpseudodev/pseudodev.c
gerard5: The path `msg->o.io.err` is not always correct, e.g. `mtGetAttr` and `mtSetAttr` are no longer handled (I removed after [comment](https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/428#discussion_r1346993503)) so the `default` switch-case clasuse would be incorrect (should be `msg->o.attr.err` instead of `msg->o.io.err`), the function `pseudo_handleMsg()` returns error if something has not been handled and requires further handling by the calling code (multidrv implementation). I removed all not handled/unsupported `mt*`s to simplyfy API and delegated handling to the driver (because it knows what to do in the default case).

In multidrv:
```C
static void multi_dispatchMsg(msg_t *msg)
...
		case id_pseudoRandom:
			if (pseudo_handleMsg(msg, multi2pseudo(id)) < 0) {
				multi_handleError(msg, -EPERM);
			}
			break;
...
```

libpseudodev/pseudodev.c
gerard5: Done..

agkaminski (approved):

PR #430, opened 2023-10-11
mag: add als31300 driver

nalajcie:
JIRA: DTR-444

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `DTR`, `DTR2`

## Checklist:

## Special treatment

### DIFF
--- a/mag/mag-als31300.c
+++ b/mag/mag-als31300.c
@@ -1,0 +1,425 @@
+/*
+ * Phoenix-RTOS
+ *
+ * ALS31300 i2c Magnetometer Sensor driver
+ *
+ * Copyright 2023 Phoenix Systems
+ * Author: Marek Bialowas
+ *
+ * This file is part of Phoenix-RTOS.
+ *
+ * %LICENSE%
+ */
+
+#include <stdint.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdbool.h>
+#include <string.h>
+#include <errno.h>
+#include <signal.h>
+#include <unistd.h>
+#include <sys/minmax.h>
+
+#include <sys/msg.h>
+#include <posix/utils.h>
+
+#include <i2c.h>
+
+#define LOG_TAG       "mag-als31300: "
+#define log(fmt, ...) printf(LOG_TAG fmt "\n", ##__VA_ARGS__)
+
+
+#define I2C_DEV_ADDR 0x60u
+#define CFG_REG1     0x02u
+#define DATA_REG1    0x28u
+#define DATA_REG2    0x29u
+
+#define CFG_REG1_I2C_ADDR_MASK    0x0001fc00u
+#define CFG_REG1_I2C_THRESH       0x00000200u
+#define CFG_REG1_ENABLE_Z         0x00000100u
+#define CFG_REG1_ENABLE_Y         0x00000080u
+#define CFG_REG1_ENABLE_X         0x00000040u
+#define CFG_REG1_CUSTOMER_EE_MASK 0x0000001fu
+
+/* 12-bit signed data */
+#define DATA_SIGN_BIT           0x0800u
+#define DATA_GAUSS_MULTIPLIER_X 1 /* ALS31300EEJASR-JOY: 1 LSB/G */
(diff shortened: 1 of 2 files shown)


mag/mag-als31300.c
```diff
@@ -0,0 +1,413 @@
...
+	return 0;
+}
+
+
+/* verify CFG1 contents to ensure we have ALS31300 under I2C_DEV_ADDR */
+static int verifyCfgReg(uint32_t val)
+{
+	if ((val & CFG_REG1_CUSTOMER_EE_MASK) != 0) {
+		return -1; /* we don't store anything in customer EEPROM - should be 0 */
+	}
```
agkaminski, line 140: Kinda risky, write would brick it permanently (until driver update, that is)

mag/mag-als31300.c
```diff
@@ -0,0 +1,413 @@
...
+#define LOG_TAG       "mag-als31300: "
+#define log(fmt, ...) printf(LOG_TAG fmt "\n", ##__VA_ARGS__)
+
+
+#define I2C_DEV_ADDR 0x60u
+#define CFG_REG1     0x02u
+#define DATA_REG1    0x28u
+#define DATA_REG2    0x29u
+
+#define CFG_REG1_I2C_ADDR_MASK    0x0001FC00u
```
agkaminski, line 37: nitpick - we use lowercase hex numbers

mag/mag-als31300.c
```diff
@@ -0,0 +1,413 @@
...
+
+static int handleRead(unsigned int devId, char *buf, size_t size)
+{
+	struct mag_data_s data;
+	int len;
+
+	if (getData(&data) < 0) {
+		return -EIO;
+	}
+	else {
```
agkaminski, line 183: else is redundant but causes indentation increase, we return anyways in if

mag/mag-als31300.c
```diff
@@ -0,0 +1,413 @@
...
+
+	return 0;
+}
+
+
+static void handleDumpMode(void)
+{
+	struct mag_data_s data;
+
+	while (1) {
```
agkaminski, line 210: I don't know if you case about MISRA, but while (1) causes implicit essential type cast int -> bool
Examples of allowed infinite loops:
while (1 == 1)
while (true) (it's ok, I guess, but we tend to avoid stdbool)
for (;;) (advised)

mag/mag-als31300.c
```diff
@@ -0,0 +1,413 @@
...
+}
+
+
+static void msgLoop(uint32_t port)
+{
+	msg_t msg;
+	msg_rid_t rid;
+	int running = 1;
+
+	while (running) {
```
agkaminski, line 226:
```suggestion
	while (running != 0) {
```

mag/mag-als31300.c
```diff
@@ -0,0 +1,413 @@
...
+
+int main(int argc, char **argv)
+{
+	unsigned int i2cBusNo;
+	uint32_t port;
+	int dumpMode = 0;
+	int daemonize = 0;
+
+	int c;
+	while ((c = getopt(argc, argv, "Ddh")) != -1) {
```
agkaminski, line 324: There're examples of MISRA-friednly getopt in psh

agkaminski (commented):
Feel free to ignore MISRA stuff, I've pointed it out just in case

mag/mag-als31300.c
nalajcie: intentional, right now support for writing EEPROM is not planned, If the feature would be implemented I expect the implementer to tackle the issue fully.

Side note: part of the configuration is also stored in eeprom (like enabled axes / 1v8 i2c compatibility) so any configuration change would need chnages to identification process. Unfortunately there is no const register we might enquiry:
image

mag/mag-als31300.c
nalajcie: daamn, we use uppercase numbers, confusing.

mag/mag-als31300.c
agkaminski: Ack

agkaminski (approved):

PR #448, opened 2023-12-18
improve stm32l4 IRQ uart driver

xvuko:
This change improves IRQ uart performance. Changing inter octet timeout to absolute timeout allows passing bigger read buffers without risk of long block.

## Motivation and Context
Currently IRQ uart does not work reliably on _redacted_ project that uses STM32 with limited CPU clock. Problems seem to appear mostly when both TX and RX are used. With this changes 9600 seems to be working without issues.

## Types of changes

- [x ] Breaking change (fix or feature that would cause existing functionality to change)

### DIFF
--- a/multi/stm32l4-multi/libmulti/libuart.c
+++ b/multi/stm32l4-multi/libmulti/libuart.c
@@ -3,8 +3,8 @@
  *
  * Multidrv-lib: STM32L4 UART driver
  *
- * Copyright 2017, 2018, 2020 Phoenix Systems
- * Author: Jan Sikorski, Aleksander Kaminski, Andrzej Glowinski
+ * Copyright 2017, 2018, 2020, 2023 Phoenix Systems
+ * Author: Jan Sikorski, Aleksander Kaminski, Andrzej Glowinski, Jan Wisniewski
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -18,6 +18,7 @@
 #include <stdint.h>
 #include <string.h>
 #include <stdlib.h>
+#include <assert.h>
 #include <sys/pwman.h>
 #include <sys/interrupt.h>
 #include <sys/minmax.h>
@@ -59,6 +60,13 @@ static const struct {
 };
 
 
+static inline size_t libuart_incrementWrap(size_t value, size_t size)
+{
+	value += 1;
+	return (value < size) ? value : 0;
+}
+
+
 static int libuart_txirq(unsigned int n, void *arg)
 {
 	libuart_ctx *ctx = (libuart_ctx *)arg;
@@ -96,19 +104,20 @@ static int libuart_rxirq(unsigned int n, void *arg)
 		*(ctx->base + icr) |= (1 << 3);
 
 		/* Rxd buffer not empty */
-		ctx->data.irq.rxdfifo[ctx->data.irq.rxdw++] = *(ctx->base + rdr);
-		ctx->data.irq.rxdw %= ctx->data.irq.rxdfifosz;
+		ctx->data.irq.rxdfifo[ctx->data.irq.rxdw] = *(ctx->base + rdr);
+		ctx->data.irq.rxdw = libuart_incrementWrap(ctx->data.irq.rxdw, ctx->data.irq.rxdfifosz);
 
 		if (ctx->data.irq.rxdr == ctx->data.irq.rxdw) {
-			ctx->data.irq.rxdr = (ctx->data.irq.rxdr + 1) % ctx->data.irq.rxdfifosz;
+			ctx->data.irq.rxdr = libuart_incrementWrap(ctx->data.irq.rxdr, ctx->data.irq.rxdfifosz);
 		}
 	}
 
 	if (ctx->data.irq.rxbeg != NULL) {
 		while (ctx->data.irq.rxdr != ctx->data.irq.rxdw && ctx->data.irq.rxbeg != ctx->data.irq.rxend) {
-			*(ctx->data.irq.rxbeg++) = ctx->data.irq.rxdfifo[ctx->data.irq.rxdr++];
-			ctx->data.irq.rxdr %= ctx->data.irq.rxdfifosz;
-			(*ctx->data.irq.read)++;
+			*(ctx->data.irq.rxbeg) = ctx->data.irq.rxdfifo[ctx->data.irq.rxdr];
+			ctx->data.irq.rxdr = libuart_incrementWrap(ctx->data.irq.rxdr, ctx->data.irq.rxdfifosz);
+			ctx->data.irq.rxbeg += 1;
+			*ctx->data.irq.read += 1;
 		}
 
 		if (ctx->data.irq.rxbeg == ctx->data.irq.rxend) {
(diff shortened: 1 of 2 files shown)


multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -368,14 +396,22 @@ static int libuart_dmaRead(libuart_ctx *ctx, void *buff, unsigned int count, cha
...
 	volatile unsigned int read = 0;
 
 	mutexLock(ctx->data.irq.rxlock);
 	mutexLock(ctx->data.irq.lock);
 
+	assert(ctx->data.irq.rxbeg == NULL);
+
 	ctx->data.irq.read = &read;
 	ctx->data.irq.rxend = (char *)buff + count;
+	ctx->data.irq.rxnblock = 0;
```
badochov, line 408: I'd move to to else branch of the if.

multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -91,32 +113,38 @@ static int libuart_rxirq(unsigned int n, void *arg)
...
 
-	if (*(ctx->base + isr) & ((1 << 5) | (1 << 3))) {
-		/* Clear overrun error bit */
-		*(ctx->base + icr) |= (1 << 3);
-
+	/*
+	 * Reading twice as we can have bytes in both data an shift registers
+	 * It is not obvous from manual that second byte is available in this scenario
+	 */
+	for (int i = 0; i < 2 && (*(ctx->base + isr) & (1 << 5)); ++i) {
```
badochov, line 120: bracket missing
```
for (int i = 0; (i < 2) && (*(ctx->base + isr) & (1 << 5)); ++i) {
```

multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -59,21 +60,42 @@ static const struct {
 };
 
 
+static inline size_t increment_wrap(size_t value, size_t size)
```
badochov, line 63: ALl function names should start with module name.
I'd suggest
`libuart_incMod` or `libuart_incWrap` on `libuart_incrementWrap`

multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -91,32 +113,38 @@ static int libuart_rxirq(unsigned int n, void *arg)
...
-			ctx->data.irq.rxdr %= ctx->data.irq.rxdfifosz;
-			(*ctx->data.irq.read)++;
+			*(ctx->data.irq.rxbeg) = ctx->data.irq.rxdfifo[ctx->data.irq.rxdr];
+			ctx->data.irq.rxdr = increment_wrap(ctx->data.irq.rxdr, ctx->data.irq.rxdfifosz);
+			ctx->data.irq.rxbeg += 1;
+			*ctx->data.irq.read += 1;
 		}
 
-		if (ctx->data.irq.rxbeg == ctx->data.irq.rxend) {
+		if (ctx->data.irq.rxbeg == ctx->data.irq.rxend || ctx->data.irq.rxnblock != 0) {
```
badochov, line 144:
```c
(ctx->data.irq.rxbeg == ctx->data.irq.rxend) || (ctx->data.irq.rxnblock != 0)
```
branckets missing

badochov (commented):
Unfortunatelly I haven't tested the changes, but they seem very good.
I only have some small nitpicks

anglov (commented):
Not tested but looks good. Point some findings

multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -91,32 +113,38 @@ static int libuart_rxirq(unsigned int n, void *arg)
...
+		/* discard oldest byte if queue full */
 		if (ctx->data.irq.rxdr == ctx->data.irq.rxdw) {
-			ctx->data.irq.rxdr = (ctx->data.irq.rxdr + 1) % ctx->data.irq.rxdfifosz;
+			ctx->data.irq.rxdr = increment_wrap(ctx->data.irq.rxdr, ctx->data.irq.rxdfifosz);
 		}
 	}
 
+	if (*(ctx->base + isr) & (1 << 3)) {
+		/* Clear overrun error bit */
+		*(ctx->base + icr) |= (1 << 3);
```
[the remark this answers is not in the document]
agkaminski: As far, as I remember, we need to handle to clear it properly and restore rx functionality

multi/stm32l4-multi/libmulti/libuart.c
anglov: I agree that it shouldn't have underscore (incrementWrap)
But I don't like idea to add libuart prefix here as it's static inline helper that should never be exposed in libuart API.
We have [some](https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/a9cbd88f96e5e2380c1f48f23b5363d8a4712440/storage/imxrt-flash/flashsrv.c#L853C21-L853C21) [examples](https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/a9cbd88f96e5e2380c1f48f23b5363d8a4712440/multi/imxrt-multi/cm4.c#L83) in current codebase that we not always do it with prefixes. 

I also think that note in coding convention allowing to omit subsystem  

> Functions used internally in C file should be declared as static. Functions used only inside the selected subsystem could be named with the name of the module instead of the name of subsystem.

Makes me thing that's it's valid to have it as `incrementWrap` without libuart prefix

What do you think @agkaminski ?

agkaminski: In this statement it is still required to use some prefix, even for `static`s.

I guess it doesn't really matter that much for a static function. Using one common naming scheme makes it easier to evolve the code, i.e. no need to rename function if we in the end decide to share it and remove `static`. That's the pretty much only objective factor here imho

multi/stm32l4-multi/libmulti/libuart.c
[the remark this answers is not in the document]
xvuko: changed

multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -59,21 +60,42 @@ static const struct {
...
+
+			if (ctx->data.irq.txbeg == ctx->data.irq.txend) {
+				ctx->data.irq.txbeg = NULL;
+				ctx->data.irq.txend = NULL;
+				release = 1;
+				*(ctx->base + cr1) &= ~(1 << 7);
+			}
 		}
 		else {
+			/* wakeup from libuart_triggerInterrupt */
```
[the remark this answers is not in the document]
xvuko: changed to wake-up

multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -59,21 +60,42 @@ static const struct {
 };
 
 
+static inline size_t increment_wrap(size_t value, size_t size)
+{
+	assert(value < size);
```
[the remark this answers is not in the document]
xvuko: Good point. removed

multi/stm32l4-multi/libmulti/libuart.c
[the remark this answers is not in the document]
xvuko: Added TODO comment

agkaminski (approved):

PR #454, opened 2024-01-15
imxrt-flash: add custom flashsrv init option

gorbanchovie:
Allows to perform direct operations (read/write/sector erase) on specific device/partition from user code.

JIRA:NIL-471

## Description
as above

## Motivation and Context
Overriding custom init function and utilizing internal flashsrv functions creates a safe way for operating on partitions with sensitive data (e.g. program code). Message interface for such partition can be stripped of write/read/erase functions to increase safety.  

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7m7-imxrt117x-nil

## Checklist:

## Special treatment

### DIFF
--- a/storage/imxrt-flash/flashsrv.c
+++ b/storage/imxrt-flash/flashsrv.c
@@ -46,7 +46,7 @@
 #endif
 
 
-enum { flash_memory_unactive = 0, flash_memory_active = 0xff };
+enum { flashsrv_memory_inactive = 0, flashsrv_memory_active = 0xff };
 
 
 typedef struct {
@@ -73,11 +73,11 @@ typedef struct {
 	uint32_t rawPort;
 
 	handle_t lock;
-} flash_memory_t;
+} flashsrv_memory_t;
 
 
 struct {
-	flash_memory_t flash_memories[FLASH_MEMORIES_NO];
+	flashsrv_memory_t flash_memories[FLASH_MEMORIES_NO];
 	uint32_t flexspi_addresses[FLASH_MEMORIES_NO];
 } flashsrv_common;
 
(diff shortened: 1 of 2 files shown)


storage/imxrt-flash/flashsrv.c
```diff
@@ -79,10 +81,10 @@ typedef struct {
 struct {
 	flash_memory_t flash_memories[FLASH_MEMORIES_NO];
 	uint32_t flexspi_addresses[FLASH_MEMORIES_NO];
+	flashsrv_partitionCtx_t ctx;
```
gerard5, line 84: nitpick: maybe `ops` better suits (operations) function vector, instead of `ctx`, also  in type name `flashsrv_partitionOps_t`

storage/imxrt-flash/imxrt-flashsrv.h
```diff
@@ -34,17 +34,37 @@ typedef struct {
...
+ * call.
+ */
+typedef struct {
+	int fID;
+
+	ssize_t (*read)(uint8_t fID, size_t offset, void *data, size_t size);
+	ssize_t (*write)(uint8_t fID, size_t offset, const void *data, size_t size);
+	int (*eraseSector)(uint8_t fID, uint32_t offs);
+	int (*getProperties)(uint8_t fID, flashsrv_properties_t *p);
+} flashsrv_partitionCtx_t;
```
gerard5, line 67: nitpick: maybe `flashsrv_partitionOps_t` ?

storage/imxrt-flash/flashsrv.c
gorbanchovie: Good idea, thx

storage/imxrt-flash/imxrt-flashsrv.h
```diff
@@ -34,17 +34,37 @@ typedef struct {
...
 
+/*
+ * Device/partition operations
+ *
+ * Includes parameters and callbacks which allow to perform direct operations (read/write/sector erase) on specific
+ * device/partition. Callbacks are initialized by flashsrv. User should provide *fID* inside flashsrv_customIntInit()
+ * call.
+ */
+typedef struct {
+	int fID;
```
gerard5, line 61: Looks like `int fID` is not used, callbacks use `uint8_t fID`. Is it used by user code later like in the /*  description */ above, reserved for future or it's redundant struct member ?

gorbanchovie: It should be `uint8_t fID`. My intention was to use it as argument to callback, so the typical use would be:
```
flashsrv_partitionOps_t *gOps;

void flashsrv_customIntInit(flashsrv_partitionOps_t *ops)
{
	ops->fID = PARTITION_ID;
        gOps = ops;
}

int write(size_t offset, const void *data, size_t size){
        gOps->write(gOps->fID, offset, data, size);`
}
```

gerard5: Ack.

gerard5 (approved):
Looks good to me.

agkaminski (approved):

PR #459, opened 2024-01-24
stm32l4-multi: add mask for reading bytes of size 7 and 6  from tty

BartoroDev:
JIRA: DEND-39

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/multi/stm32l4-multi/tty.c
+++ b/multi/stm32l4-multi/tty.c
@@ -298,6 +298,7 @@ static void tty_irqthread(void *arg)
 {
 	tty_ctx_t *ctx = (tty_ctx_t *)arg;
 	int keptidle = 0;
+	uint8_t mask;
 
 	for (;;) {
 		uint8_t rxbyte;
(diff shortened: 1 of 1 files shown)


anglov (approved):
LGTM

agkaminski (approved):

multi/stm32l4-multi/tty.c
```diff
@@ -309,12 +310,22 @@ static void tty_irqthread(void *arg)
 		}
 		mutexUnlock(ctx->irqlock);
```
gorbanchovie, line 311: On imx driver I got such comment:

"imho this check should be performed under mutex"

Maybe it applies also here?

multi/stm32l4-multi/tty.c
```diff
@@ -309,12 +310,22 @@ static void tty_irqthread(void *arg)
...
 
+		if ((ctx->ttyCommon.term.c_cflag & CSIZE) == CS7) {
+			mask = 0x7f;
+		}
+		else if ((ctx->ttyCommon.term.c_cflag & CSIZE) == CS6) {
```
gorbanchovie, line 316: You could update commit message with CS6

multi/stm32l4-multi/tty.c
agkaminski: In this case we need to synchronize with `tty_setCflag()` - but it is not protected by any lock. So you're right, but this is non-trivial change

BartoroDev: so I will merge it as is, synchronization should be done as an independent change.

PR #462, opened 2024-01-29
imxrt-multi/spi: Fix RX FIFO being read too early

Darchiv:
DONE: RTOS-755

`spi_transaction` msg may return (read) less bytes than requested (e.g. 8 out of 10). RX FIFO watermark is set to 0 on `spi_config` msg and never changed, meaning that _Receive Data_ event is asserted after putting one word into FIFO. This change sets the RX watermark to an appropriate value so that data is read after it is fully received. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt-1170.

## Checklist:

## Special treatment


agkaminski:
So... Do we merge this? :)

PR #465, opened 2024-02-01
stm32l4x6-multi: improve libuart DMA

anglov:
## Description

This change provide to improvement to libuart DMA code:
1. Adding missing volatile specifier for pointer that's used to synchronisation between interrupt and reader code
2. Removing modulo from interrupt code, what potentially can speedup execution

## Motivation and Context

I found that sometimes I got following result when stressing up this driver:

Received pattern:
```
00155718_ABCDEFGHIJKLMNOPQRSTUVWXY\
00WXY\19_ABCDEFGHIJKLMNOPQRSTUVWXY\
00155720_ABCDEFGHIJKLMNOPQRSTUVWXY\
```
Expected pattern:
```
00155718_ABCDEFGHIJKLMNOPQRSTUVWXY\
00155719_ABCDEFGHIJKLMNOPQRSTUVWXY\
00155720_ABCDEFGHIJKLMNOPQRSTUVWXY\
```
(new lines replaced with \ for readability)

Code analysis make me find missing volatile for pointer. But it's not resolve the issue.

I found that this issue is not connected with overrun detected by the driver. 

Digging it deeper brings me conclusion that overrun happening while interrupt is handled, after check for overrun. 

This changes still doesn't really fix the issue (really unpredictable overrun), but removes unneded defensive modulos replacing it with simpler operations and improves situation a little.

Some more work required  to handle it properly and simplify interrupt code. 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7m4-stm32l4x6

## Checklist:

## Special treatment

### DIFF
--- a/multi/stm32l4-multi/libmulti/libuart.c
+++ b/multi/stm32l4-multi/libmulti/libuart.c
@@ -481,18 +481,22 @@ static void libuart_infiniteRxHandler(void *arg, int type)
 	size_t cpysz;
 	char *rxbuf = (char *)ctx->data.dma.rxbuf;
 
-	endPos = (ctx->data.dma.rxfifosz - libdma_leftToRx(ctx->data.dma.per)) % ctx->data.dma.rxfifosz;
+	endPos = ctx->data.dma.rxfifosz - libdma_leftToRx(ctx->data.dma.per);
 
-	/* Check overrun */
 	rxfifopos = ctx->data.dma.rxfifopos;
 	rxfifoprevend = ctx->data.dma.rxfifoprevend;
 
+	/* Check overrun */
 	if (((rxfifoprevend < rxfifopos) && ((rxfifopos < endPos) || (endPos < rxfifoprevend))) ||
 		((rxfifopos <= rxfifoprevend) && ((rxfifopos < endPos) && (endPos < rxfifoprevend)))) {
 		rxfifopos = endPos;
 	}
 
-	infifo = (ctx->data.dma.rxfifosz + endPos - rxfifopos) % ctx->data.dma.rxfifosz;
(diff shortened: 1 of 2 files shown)


agkaminski (approved):

multi/stm32l4-multi/libmulti/libuart.c
```diff
@@ -481,18 +481,22 @@ static void libuart_infiniteRxHandler(void *arg, int type)
 	size_t cpysz;
 	char *rxbuf = (char *)ctx->data.dma.rxbuf;
```
nalajcie, line 482: didn't read the code to check if rxbuf is needed to really be volatile, buuut:

 - is volatile dropped here consciously? If so - maybe add explicit comment?
 - please note that `memcpy` is not guaranteed to work with volatile memory but probably ok in this use case (assuming compiler won't optimize in some strange way)?

badochov (approved):
LGTM

multi/stm32l4-multi/libmulti/libuart.c
anglov: Welp. I have many more questions here. 

- If DMA write data to fifo, and interrupt gathers data from that fifo, is it required to make this fifo volatile ?
- As memcpy not working for volatile types, should it be implemented as loop?
- As we should have fast and optimized cpy, should we do it as some assembly? 

In this change I added volatile for ctx->data.dma.rxbuf pointer (not data!)  which improves situation. As i write in PR it not resolving issue fully, it's rather obvious things that improves current situation (synchronisation is no longer done on non-volatile pointer) . So I would merge it as-is and do more work in the future, after more analysis.

PR #468, opened 2024-02-09
imxrt-multi: More flexible peripheral configuration

Darchiv:
## Description

New behavior:
- peripheral-enabling flags (UARTn, SPIn) must be either set to 0, 1, or
  be left undefined. They are assumed to be 0 or 1 in device driver
  modules for the purpose of counting/offsetting devices, so in case of
  an invalid value an error is thrown;
- fallback pin/mux definitions in config.h are slated for removal (they
  will be moved to evk boards' board_config.h files for reference and
  convenience), so deprecation warnings are thrown whenever a pin/mux
  for an enabled peripheral is left undefined and a default is used;
- SPIn_SD0 (zero) has been deprecated in favor of SPIn_SDO (letter O);
- pin/mux definitions may be defined and left empty - this is an
  explicit way to leave this pin/mux not configured at all by multidrv;
- if a pin for a peripheral is required to be configured and it is not
  defined or is empty, an error is thrown.

Breaking changes:
- peripherals enabled by default (UART1, UART11, SPI1, I2C1) are now
  disabled by default;
- UARTn_HW_FLOWCTRL==1 was enough to configure CTS/RTS muxes - now
  UARTn==1 is also required.

More work is required for uniform experience: in devices (i2c), plo, and kernel (commonize UART_CONSOLE, etc.). Deprecation warnings are meant to be left for some time until projects are migrated to a new style, without (hopefully!) breaking any current builds. Errors for "required" pins (e.g. UART TX and RX pins) are thrown to minimize the potential for bugs (in case e.g. only TX or RX are configured), but this may be too aggressive and limiting (e.g. when we really want to leave RX at default and only configure TX).

This change also fixes a problem with SPI driver overriding mux values.

## Motivation and Context

The general idea was to enable overriding the default pin definitions without touching any non-project code.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


multi/imxrt-multi/helpers.h
```diff
@@ -0,0 +1,32 @@
...
+#define _HELPERS_H_
+
+#define CONCAT2(x, y)    x##y
+#define CONCAT3(x, y, z) x##y##z
+
+#define PIN2MUX(x) CONCAT2(pctl_mux_gpio_, x)
+#define PIN2PAD(x) CONCAT2(pctl_pad_gpio_, x)
+
+/* For the following macros, anything that expands to a numerical expression (or to an empty value) is fine.
+String literals don't work.
```
agkaminski, line 22:
```suggestion
String literals don't work. */
```

multi/imxrt-multi/config.h
```diff
@@ -219,37 +703,338 @@
...
 
+
 /* SPI */
 
 #ifndef SPI1
-#define SPI1 1
+#define SPI1 0
+#elif !ISBOOLEAN(SPI1)
+#error "SPI1 must have a value of 0, 1, or be undefined"
+#endif
```
anglov, line 713: In final solution I would move  error/warning messages and checkers into .c file. In case of warning you will  have wall of warnings. 

However I know that currently it may be hard to achieve because of deprecation of old approach.

anglov (commented):
Didn't check values and not testing, but looking at approach. I like it that you move all configuration into board config and do not have any unobvious defaults.

multi/imxrt-multi/config.h
```diff
@@ -26,186 +28,668 @@
...
+#if UART1
+
+#ifndef UART1_TX_PIN
+#warning "[Deprecated] UART1_TX_PIN should be explicitly defined in board_config.h. Using the default"
+#define UART1_TX_PIN UART1_TX_PIN_DEFAULT
+#elif ISEMPTY(UART1_TX_PIN)
+#error "UART1_TX_PIN must not be empty"
+#endif
+
+#ifndef UART1_RX_PIN
```
nalajcie, line 56: not sure what would be more readable ^C^V almost the same configuration 12 times or specifying it once in external file and including 12 times, eg:

```c
#define _UART_N 1
#include "_uart_def.h"
#undef _UART_N
```

Edit:
I think we could keep it as-is, a the code is simple (albeit kindof long) with comment describing what can be set in `board_config.h`. Probably we will migrate to device tree in the future, so let's not over-complicate the solution right now.

multi/imxrt-multi/config.h
```diff
@@ -26,186 +28,668 @@
...
+#warning "[Deprecated] UART1_CTS_PIN should be explicitly defined (may be empty) in board_config.h. Using the default"
+#define UART1_CTS_PIN UART1_CTS_PIN_DEFAULT
 #endif
 
 #ifndef UART1_BAUDRATE
+#warning "[Deprecated] UART1_BAUDRATE should be explicitly defined in board_config.h. Using the default"
 #define UART1_BAUDRATE 115200
 #endif
 
+#endif
```
nalajcie, line 89: as we're not using indentation for ifdefs (maybe worth reconsidering?) - write what context we're exiting -  `/* UART1 */`

multi/imxrt-multi/spi.c
```diff
@@ -470,116 +466,140 @@ static void spi_initPins(void)
...
+			{
+				PIN2MUX(SPI1_SCK),
+				PIN2MUX(SPI1_SD0),
+				PIN2MUX(SPI1_SDI),
+#if defined(SPI1_PCS0) && !ISEMPTY(SPI1_PCS0)
+				PIN2MUX(SPI1_PCS0),
 #endif
-#ifdef SPI1_PCS1
-		PIN2MUX(SPI1_PCS1),
+#if defined(SPI1_PCS1) && !ISEMPTY(SPI1_PCS1)
```
nalajcie, line 478: seems like  `defined` should be inside `isempty` as it's always used in pair here?

multi/imxrt-multi/helpers.h
```diff
@@ -0,0 +1,32 @@
...
+
+/* True if X is defined, is not empty, and has a non-zero value. */
+#define ISTRUTHY(X) ((X + 0) != 0)
```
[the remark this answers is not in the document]
Darchiv: Do you mean `#define ISTRUTHY(X) (defined(X) && (X != 0))`? This is undefined behavior (see `-Wexpansion-to-defined`) and also causes a syntax error if `X` is defined but empty. That's why `defined()` are used directly in `#if`s.

multi/imxrt-multi/spi.c
Darchiv: This would be undefined behavior (see `-Wexpansion-to-defined`).

multi/imxrt-multi/config.h
```diff
@@ -26,186 +28,668 @@
 #include "imxrt1060.h"
 #endif
 
+#include <board_config.h>
```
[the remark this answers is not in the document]
Darchiv: The idea is to place all the possible configuration defines in evaluation kits' `board_config.h` (make them "templates"). Peripherals which are not used would also be included (but disabled) and have their default pins defined. Possibly with comments to make it clear how it works.

multi/imxrt-multi/helpers.h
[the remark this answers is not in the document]
nalajcie: thanks for bringing it to my attention, I wasn't aware of that. Seems the implementation for function-like macros in all major compilers is consistent (https://reviews.llvm.org/D15866) but it's still undefined behavior so let's not change it.

Maybe explicit comment regarding UB should be added here to avoid undesired changes in the future.

gerard5 (commented):
Along with the new approach to peripheral configuration, it would be good to be able to individually configure buffers for each UART instance, instead of one common size ([like now](https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/f4f4412618df2f42a9fbf0013602fbc5b8c29c9a/multi/imxrt-multi/uart.c#L50-L52)). Sometimes some protocols require more space to fit the frame than the default size, and increasing all instead of the dedicated one can lead to poor fit, especially on small targets.

multi/imxrt-multi/config.h
```diff
@@ -3,7 +3,7 @@
  *
  * iMX RT multi driver config.
  *
- * Copyright 2019 Phoenix Systems
+ * Copyright 2019-2024 Phoenix Systems
```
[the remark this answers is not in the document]
Darchiv: I changed that, but we should probably track this better and list all years/ranges which appeared in git history (maybe some CI check?)

multi/imxrt-multi/config.h
```diff
@@ -26,186 +28,668 @@
...
+#define UART1_CTS_PIN UART1_CTS_PIN_DEFAULT
+#else
+#define UART1_RTS_PIN
+#define UART1_CTS_PIN
+#endif
+#endif
+
+#ifndef UART1_RTS_PIN
+#warning "[Deprecated] UART1_RTS_PIN should be explicitly defined (may be empty) in board_config.h. Using the default"
+#define UART1_RTS_PIN UART1_RTS_PIN_DEFAULT
```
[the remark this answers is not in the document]
Darchiv: Reverted back to using `UARTn_HW_FLOWCTRL` as it is more explicit and states "configure hardware flow control, and use the following RTS and/or CTS..."

multi/imxrt-multi/helpers.h
Darchiv: Done

multi/imxrt-multi/config.h
Darchiv: Done

multi/imxrt-multi/uart.c
```diff
@@ -12,22 +12,18 @@
  */
 
 
-#include <string.h>
-#include <stdint.h>
-#include <stdio.h>
 #include <errno.h>
+#include <stdint.h>
```
nalajcie, line 16: `string.h` is needed for `memcpy`:
image

multi/imxrt-multi/spi.c
```diff
@@ -470,133 +466,127 @@ static int spi_muxVal(int spi, int mux)
...
 static void spi_initPins(void)
 {
-	int i, j, isel, val;
+	int i, isel, val;
 	static const struct {
 		int spi;
-		int muxes[7];
+		int mux;
 	} info[] = {
+	/* clang-format off */
```
nalajcie, line 474: not needed

multi/imxrt-multi/spi.c
```diff
@@ -470,133 +466,127 @@ static int spi_muxVal(int spi, int mux)
...
-		PIN2MUX(SPI6_PCS3),
+#if defined(SPI6_PCS3) && !ISEMPTY(SPI6_PCS3)
+		{ 5, PIN2MUX(SPI6_PCS3) },
 #endif
-		},
-		},
 #endif
-
 	};
+	/* clang-format on */
```
nalajcie, line 583: not needed

multi/imxrt-multi/helpers.h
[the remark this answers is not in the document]
nalajcie: still, might be good to add comment regarding WHY we can't add `defined(X)` here (eg. "using `defined(X)` in macro body is UB (see -Wexpansion-to-defined for details))

multi/imxrt-multi/config.h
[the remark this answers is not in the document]
nalajcie: Ok for now, I still lean towards simple description here instead of looking at board_config.h which will also be long - but we will see how it looks when board_config.h would be ready :)

nalajcie (commented):
added some nitpicks, otherwise LGTM

multi/imxrt-multi/helpers.h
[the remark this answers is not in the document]
Darchiv: Added a comprehensive (I think) comment

nalajcie (approved):

PR #478, opened 2024-03-20
imxrt-multi/uart: add intermediate rx fifo between irq handler and thread

gerard5:
#### imxrt-multi/uart: add intermediate rxfifo (irq-thread)
    
The rx overrun flag is cleared in the interrupt handler. The hardware fifo turns out to be too small (four octets) and cannot be increased (read only register on imxrt1176). RX bytes were lost in irq handler. To solve the issue this change adds small FIFO between the interrupt and the thread to improve performance significantly. "Lock-free FIFO" from libtty implementation used to accomplish this.
    
Additionally, statistics regarding the counting of (software and hardware) overrun events are introduced.

Removed unused variable in uart_common, and corrected clock settings for imxrt1176 to behave like on imxrt10xx (just make clock run).

#### imxrt-multi/uart: allow per instance UARTx_BUFSIZE

This change allows to set libtty buffers for each uart instance separately. By default, the UART_BUFSIZE setting for an instance will be overridden by the UARTx_BUFSIZE (x - instance number) setting when defined in the board configuration.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: `imxrt1176-nil`

## Checklist:

## Special treatment

### DIFF
--- a/multi/imxrt-multi/uart.c
+++ b/multi/imxrt-multi/uart.c
@@ -26,6 +26,7 @@
 #include <sys/platform.h>
 
 #include <libtty.h>
+#include <libtty-lf-fifo.h>
 #include <board_config.h>
 
 #include "common.h"
@@ -51,6 +52,10 @@
 #define UART_BUFSIZE 512
 #endif
 
+#ifndef UART_RXFIFOSIZE
+#define UART_RXFIFOSIZE 128
+#endif
+
 
 typedef struct uart_s {
 	char stack[1024] __attribute__ ((aligned(8)));
@@ -67,6 +72,17 @@ typedef struct uart_s {
 	size_t txFifoSz;
 
 	libtty_common_t tty_common;
+
+	/* fifo between irq and thread */
+	lf_fifo_t rxFifoCtx;
+	uint8_t rxFifoData[UART_RXFIFOSIZE];
+
+	/* statistics */
+	struct {
+		/* NOTE: to read statistics use debugger */
+		size_t hw_overrunCntr;
+		size_t sw_overrunCntr;
+	} stat;
 } uart_t;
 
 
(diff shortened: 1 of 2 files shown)


multi/imxrt-multi/uart.c
```diff
@@ -86,52 +97,66 @@ static const int uartPos[] = { UART1_POS, UART2_POS, UART3_POS, UART4_POS, UART5
...
+		}
+	}
 
+	/* Clear errors: parity, framing, noise, overrun and idle flag */
+	*(uart->base + statr) = (status & (0x1fuL << 16));
 
-static inline int uart_getTXcount(uart_t *uart)
-{
-	return (*(uart->base + waterr) >> 8) & 0xff;
+	/* reschedule */
```
agkaminski, line 138: Nitpick - not exactly, wakeup registered cond (uart->cond)

multi/imxrt-multi/uart.c
```diff
@@ -67,12 +69,21 @@ typedef struct uart_s {
...
+
+	/* fifo between irq and thread */
+	lf_fifo_t rxFifoCtx;
+	uint8_t rxFifoData[UART_RXFIFOSIZE];
+
+	/* statistics */
+	struct {
+		size_t hw_overrunCntr;
+		size_t sw_overrunCntr;
+	} stat;
```
agkaminski, line 81: This struct seems write-only

gerard5: imxrt-multi does not have mechanisms for reading counters (yet), this structure is helpful for reading with a debugger. I can add a comment.

multi/imxrt-multi/uart.c
gerard5: Right, thanks.

agkaminski (approved):

PR #492, opened 2024-05-09
imxrt-flash: allow exclusive access to raw partitions 

gorbanchovie:
## Description

Meterfs stack size needs to be increased as it overflows (happened when flash was empty and files were to be created)
Add exclusive access for single PID. According to POSIX behavior of O_EXCL mode of open (when O_CREAT is unset) is undefined. This change interprets O_EXCL flag as exclusive access for a process that performs `open` on partition until `close`.

## Motivation and Context

Adding exclusive access to partition is an easy, probably temporary solution to restrict access to sensitive memory spaces 

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt1176.

## Checklist:

## Special treatment

### DIFF
--- a/storage/imxrt-flash/flashsrv.c
+++ b/storage/imxrt-flash/flashsrv.c
@@ -74,7 +74,7 @@ typedef struct {
 
 	uint32_t pCnt;
 	uint8_t fStatus;
-	uint8_t rawActive;
+	bool rawActive;
 
 	oid_t fOid;
 	uint32_t rawPort;
@@ -770,7 +770,7 @@ static int flashsrv_mountPart(flashsrv_partition_t *part)
 			}
 
 			if (!flashsrv_common.flash_memories[part->fID].rawActive) {
-				flashsrv_common.flash_memories[part->fID].rawActive = 1;
+				flashsrv_common.flash_memories[part->fID].rawActive = true;
 
 				mem = malloc(THREAD_STACKSZ);
 				if (mem == NULL) {
@@ -961,7 +961,7 @@ static int flashsrv_flashMemoriesInit(void)
 			return err;
 		}
 
-		memory->rawActive = 0;
+		memory->rawActive = false;
 		memory->pCnt = 0;
 
 		err = create_dev(&memory->fOid, path);
--- a/storage/imxrt-flash/flashsrv.c
+++ b/storage/imxrt-flash/flashsrv.c
@@ -14,8 +14,10 @@
 #include <endian.h>
 #include <errno.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <string.h>
 #include <unistd.h>
+#include <fcntl.h>
 #include <sys/msg.h>
 #include <sys/file.h>
 #include <sys/threads.h>
@@ -59,6 +61,8 @@ typedef struct {
 	oid_t oid;
 	uint8_t fID;
 	uint8_t pStatus;
+
+	pid_t authProcPid;
 
 	void *fsCtx;
 } flashsrv_partition_t;
@@ -261,6 +265,12 @@ static int flashsrv_fsEraseSectorf1(struct _meterfs_devCtx_t *devCtx, off_t offs
 	(void)devCtx;
 
 	return flashsrv_eraseSector(1, offs);
+}
+
+
+static inline bool hasAccess(flashsrv_partition_t *part, pid_t pid)
+{
+	return part->authProcPid == 0 || part->authProcPid == pid;
 }
 
 
@@ -572,6 +582,9 @@ static void flashsrv_rawThread(void *arg)
 		if (msg.oid.id >= memory->pCnt) {
 			msg.o.err = -EINVAL;
 		}
+		else if (!hasAccess(&memory->parts[msg.oid.id], msg.pid)) {
+			msg.o.err = -EACCES;
+		}
 		else {
 			switch (msg.type) {
 				case mtRead:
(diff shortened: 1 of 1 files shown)


storage/imxrt-flash/flashsrv.c
```diff
@@ -82,9 +84,16 @@ struct {
...
 	uint32_t flexspi_addresses[FLASH_MEMORIES_NO];
 	flashsrv_partitionOps_t ops;
+	uint8_t excludingOpen;
 } flashsrv_common;
 
 
+static void usage(const char *progname)
+{
+	ps_log("usage : %s [-o]\n", progname);
+}
```
anglov, line 94: Looks like remaing of some old approach?

storage/imxrt-flash/flashsrv.c
```diff
@@ -492,6 +501,9 @@ static void flashsrv_rawThread(void *arg)
 		if (msg.oid.id >= memory->pCnt) {
 			msg.o.err = -EINVAL;
 		}
+		else if (!(memory->parts[msg.oid.id].authProcPid == 0 || memory->parts[msg.oid.id].authProcPid == msg.pid)) {
+			msg.o.err = -EACCES;
```
anglov, line 505: Blocking gettatr can cause that ls will stop work

storage/imxrt-flash/flashsrv.c
```diff
@@ -982,8 +997,25 @@ static int flashsrv_customInit(void)
...
+int main(int argc, char **argv)
 {
+	for (;;) {
+		int c = getopt(argc, argv, "o");
+		if (c < 0) {
+			break;
+		}
+
+		switch (c) {
+			case 'o': /* Allows to restrict operations on raw partitions to process that opens partition first */
```
anglov, line 1009: Better mechanism would be if you could pass partition(s) to restrict, instead of enable/disable restriction globally.

storage/imxrt-flash/flashsrv.c
nalajcie: @gorbanchovie please look at the compilation results before adding anyone to the review - not to waste their time
image

storage/imxrt-flash/flashsrv.c
```diff
@@ -58,6 +58,8 @@ typedef struct {
 	uint8_t fID;
 	uint8_t pStatus;
 
+	uint8_t authProcPid;
```
nalajcie, line 61: needs to be `pid_t` - otherwise I can use uint8_t overflow to impersonate the other process

storage/imxrt-flash/flashsrv.c
```diff
@@ -82,9 +84,16 @@ struct {
 	flashsrv_memory_t flash_memories[FLASH_MEMORIES_NO];
 	uint32_t flexspi_addresses[FLASH_MEMORIES_NO];
 	flashsrv_partitionOps_t ops;
+	uint8_t excludingOpen;
```
nalajcie, line 87: not `excluding` but `exclusive`

storage/imxrt-flash/flashsrv.c
nalajcie: not sure, this is gettatr for the device, getattr for filesystem should be taken care of by dummyfs?

Still, don't use `!` if not needed - makes it harder to understand, eg:
`if (memory->parts[msg.oid.id].authProcPid != 0 && memory->parts[msg.oid.id].authProcPid != msg.pid) return -EACCES`

storage/imxrt-flash/flashsrv.c
nalajcie: Other ideas I've just thought of:
 - ditch the commandline param, abuse `O_EXCL` mode of `open` - if set - allow only this PID until `close` (maaaybe can even be used for meterfs (but not for every file open, just for the partition file?))
 - ditch the commandline param, use POSIX locking mechanism API (`fcntl` + `F_SETLK`) - probably harder :)

nalajcie (changes requested):

nalajcie:
:exclamation: please use sensible PR name, it will be used to prepare changelog for the next OS release :exclamation: 

Also, don't lie
image

storage/imxrt-flash/flashsrv.c
anglov: Hmm, right. This getattr is used for seek after opening file. Sorry, my mistake..

gorbanchovie: > Still, don't use `!` if not needed - makes it harder to understand, eg: `if (memory->parts[msg.oid.id].authProcPid != 0 && memory->parts[msg.oid.id].authProcPid != msg.pid) return -EACCES`

Hmm, my perspective was that it can be read as `!hasAccess` and it's easier to undarstand. Maybe now it's clear?

storage/imxrt-flash/flashsrv.c
gorbanchovie: > * ditch the commandline param, abuse `O_EXCL` mode of `open` - if set - allow only this PID until `close` (maaaybe can even be used for meterfs (but not for every file open, just for the partition file?))

I like this approach. I figured that maybe exclusive access could be granted only if `O_CREAT` flag is not set as behaviour with `O_CREAT | O_EXCL` is well defined and for `O_EXCL` is not.

storage/imxrt-flash/flashsrv.c
gorbanchovie: done

storage/imxrt-flash/flashsrv.c
gorbanchovie: outdated

anglov (commented):
Mainly stylystic comment. Functionally looks ok but I didn't tested.

I think you should update PR description

storage/imxrt-flash/flashsrv.c
```diff
@@ -479,59 +488,73 @@ static void flashsrv_rawCtl(flashsrv_memory_t *memory, msg_t *msg)
...
+			msg->o.lookup.fil.port = part->oid.port;
+			memcpy(&msg->o.lookup.dev, &msg->o.lookup.fil, sizeof(oid_t));
+			break;
+
+		case mtOpen:
+			/* Exclusive access mechanism enabled on O_EXCL except when O_CREAT is passed */
+			if ((msg->i.openclose.flags & (O_CREAT | O_EXCL)) == O_EXCL) {
+				part->authProcPid = msg->pid;
+			}
+			err = meterfs_open(msg->oid.id, (meterfs_ctx_t *)part->fsCtx);
```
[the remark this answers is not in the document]
gorbanchovie: You are right. I mean the end result is the same, but it's not how it should work. As there is no way to "open" meterfs partition I dropped the changes for meterfs partitions.

anglov (approved):

ziemleszcz (approved):

PR #514, opened 2024-09-13
imxrt-multi: add support for console over RTT

jmaksymowicz:
Allows for UART-like devices to be created which transmit data over RTT.

## Description
Console can be defined to use either UART (as before) or an RTT channel. For now it is not possible to have console through both UART and RTT.

Each RTT channel can be configured to work in non-blocking or blocking mode. In non-blocking mode data is discarded if there is no space left in the buffer. In blocking mode the code attempts to detect if there is a host receiving data over the RTT interface and if there is, the code waits a pre-defined period until there is space left in the buffer.

## Motivation and Context
With this change it is possible to access console (e.g. `psh`) over RTT interface.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt1170

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work: https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/592


multi/imxrt-multi/rttuart.h
```diff
@@ -0,0 +1,13 @@
+#include <stddef.h>
```
Darchiv, line 1: Copyright notice missing

multi/imxrt-multi/rttuart.h
```diff
@@ -0,0 +1,13 @@
...
+#include <sys/msg.h>
```
Darchiv, line 2: Include guard missing

multi/imxrt-multi/uart.c
```diff
@@ -772,7 +772,9 @@ static void uart_initPins(void)
 
 void uart_klogCblk(const char *data, size_t size)
 {
-	libtty_write(&uart_common.uarts[uartPos[UART_CONSOLE - id_uart1]].tty_common, data, size, 0);
+#ifdef UART_CONSOLE
+	libtty_write(&uart_common.uarts[uartPos[UART_CONSOLE - 1]].tty_common, data, size, 0);
```
Darchiv, line 776: Why change  `id_uart1` to `1`?

multi/imxrt-multi/rttuart.c
```diff
@@ -0,0 +1,252 @@
...
+#define RTT_RETRIES              (RTT_NO_PICKUP_TIMEOUT_MS / RTT_POLLING_RATE_MS)
+
+/* Doesn't need to be large, data will mostly be stored in RTT buffers */
+#define TTY_BUF_SIZE 64
+
+#define RTT_CHANNEL0_POS 0
+#define RTT_CHANNEL1_POS (RTT_CHANNEL0_POS + RTT_CHANNEL0)
+#define RTT_ACTIVE_CNT   (RTT_CHANNEL0 + RTT_CHANNEL1)
+
+typedef struct uart_s {
```
Darchiv, line 34: `uart_s` or `rttuart_s`?

multi/imxrt-multi/rttuart.c
```diff
@@ -0,0 +1,252 @@
...
+	int chn;
+	handle_t lock;
+	libtty_common_t tty_common;
+	volatile unsigned int diag_txSkipped; /* Accessed using a debugger */
+} rttuart_t;
+
+libtty_common_t *g_secondary_tty = NULL;
```
Darchiv, line 41: Unused global?

multi/imxrt-multi/uart.c
jmaksymowicz: `UART_CONSOLE` is the number of the UART being used for console (starting at 1), not an ID from `imxrt-multi.h`. It just happens that `id_uart1 == 1`, but if anyone moved these IDs around this code would stop working.

Darchiv: Valid point

multi/imxrt-multi/rttuart.h
jmaksymowicz: Added copyright notice

multi/imxrt-multi/rttuart.h
jmaksymowicz: Added header guard

multi/imxrt-multi/config.h
```diff
@@ -766,6 +766,54 @@
 #endif
 #endif
 
+#if defined(UART_CONSOLE)
+#if ISEMPTY(UART_CONSOLE)
+#error "UART_CONSOLE must not be empty"
+#elif UART_CONSOLE <= 0
+#error "Invalid value for UART_CONSOLE"
+#endif
+#endif
```
Darchiv, line 775: `UART_CONSOLE` is a bit flaky - it may be used somewhere else (e.g. kernel, plo, some usermode code...) where it is expected to have a specific value and at the same time it serves as a catch-all for ease of use (just one define to set up UART in plo, kernel, and multidrv). I think you can relax restrictions on `UART_CONSOLE` and rely on `UART_CONSOLE_USER` (used on a few targets) to make the process of using consoles/UARTs less painful.

```suggestion
#ifndef UART_CONSOLE_USER
#define UART_CONSOLE_USER UART_CONSOLE
#endif
```
This needs the condition above (`#if !defined(UART_CONSOLE) && !defined(RTT_CHANNEL_CONSOLE)`) to be reverted back to `#ifndef UART_CONSOLE`.

Also see the comment below

multi/imxrt-multi/rtt.c
```diff
@@ -0,0 +1,305 @@
...
+#include <sys/threads.h>
+#include <sys/file.h>
+
+#include <librtt.h>
+
+#include "common.h"
+#include "rtt.h"
+
+
+#define RTT_MAP_NAME             "rtt"
```
Darchiv, line 30: This could be `#ifndef`ed to make it possible to change the map name via e.g. board_config.h.

librtt/librtt.c
```diff
@@ -0,0 +1,432 @@
...
+	performCacheOp(librtt_common.invalFn, (void *)srcBuf, sz + 1, rd, wr);
+	while ((todo != 0) && (rd != wr)) {
+		*dstBuf++ = srcBuf[rd];
+		rd = (rd + 1) & sz;
+		todo--;
+	}
+
+	rtt->channel[ch].rd = rd;
+
+	dataMemoryBarrier();
```
Darchiv, line 160: This should actually be before the write to `rd` - if reorder happens and `rd` is set first, host/jlink may overwrite data before librtt has a chance to read it.

librtt/librtt.c
```diff
@@ -0,0 +1,432 @@
...
+			todo--;
+		}
+
+		rtt->channel[ch].rd = wr;
+	}
+
+	performCacheOp(librtt_common.cleanFn, (void *)dstBuf, sz + 1, rd, wr);
+	rtt->channel[ch].wr = wr;
+
+	dataMemoryBarrier();
```
Darchiv, line 202: This should actually be before the write to `wr` - if reorder happens and `wr` is set first, host/jlink may read data before librtt writes it.

librtt/librtt.c
```diff
@@ -0,0 +1,432 @@
...
+		if (librtt_tagReversed[n] != rtt->tag[librtt_tagLength - 1 - n]) {
+			break;
+		}
+	}
+
+	if (n != librtt_tagLength) {
+		return -EINVAL;
+	}
+
+	if ((rtt->txChannels > LIBRTT_TXCHANNELS) || (rtt->rxChannels > LIBRTT_RXCHANNELS)) {
```
Darchiv, line 331: This is tricky. We should not rely on CB being what we want it to be - if plo/kernel need additional channels, then we could reuse just the lower ones (as limited by `LIBRTT_TXCHANNELS` and `LIBRTT_TXCHANNELS`) instead of rejecting them.

jmaksymowicz: Fixed, now the control blocks can have more channels than the limit. The channels beyond the limit won't be usable and won't be modified.

multi/imxrt-multi/rtt.c
jmaksymowicz: Done, renamed to `RTT_SYSPAGE_MAP_NAME` to keep in sync with your changes in kernel.

multi/imxrt-multi/config.h
```diff
@@ -766,6 +766,64 @@
 #endif
 #endif
 
+#if defined(UART_CONSOLE)
```
Darchiv, line 769: Already defined above

multi/imxrt-multi/config.h
```diff
@@ -766,6 +766,64 @@
...
+#if ISEMPTY(UART_CONSOLE)
```
Darchiv, line 770: Actually, there are no restrictions for this - `UART_CONSOLE` may be empty, even though it may be handled differently in different modules (plo, kernel).

multi/imxrt-multi/imxrt-multi.c
```diff
@@ -168,7 +169,13 @@ static void uart_dispatchMsg(msg_t *msg)
 
 	switch (id) {
 		case id_console:
+#if !ISEMPTY(RTT_CHANNEL_CONSOLE)
+			rtt_handleMsg(msg, RTT_CHANNEL_CONSOLE + id_rtt0);
+#elif !ISEMPTY(UART_CONSOLE_USER)
 			uart_handleMsg(msg, UART_CONSOLE - 1 + id_uart1);
+#else
+			dummyuart_handleMsg(msg);
```
Darchiv, line 177: As discussed, there must be exactly one console enabled at the time (at least one of them must be enabled)

multi/imxrt-multi/config.h
jmaksymowicz: Fixed

multi/imxrt-multi/imxrt-multi.c
jmaksymowicz: Added check in `config.h` to ensure this condition doesn't happen.

multi/imxrt-multi/config.h
jmaksymowicz: OK, I removed this check.

multi/imxrt-multi/config.h
```diff
@@ -766,6 +766,63 @@
...
 
+#if !ISEMPTY(UART_CONSOLE)
+#if UART_CONSOLE <= 0
+#error "Invalid value for UART_CONSOLE"
+#endif
+#endif
+
+#if defined(UART_CONSOLE_USER)
+#if !ISEMPTY(UART_CONSOLE_USER)
+#if (UART_CONSOLE_USER <= 0)
```
Darchiv, line 777: These two ifs may be squashed into one

multi/imxrt-multi/config.h
```diff
@@ -766,6 +766,63 @@
...
+#endif
+#endif
+#else
+#define UART_CONSOLE_USER UART_CONSOLE
+#endif
+
+
+/* RTT */
+
+#ifndef RTT_CHANNEL0
```
Darchiv, line 788: Maybe let's use the current convention for now, i.e. `RTTx`, `RTTx_BLOCKING`, even though channels are 0-based. Users are familiar with it

multi/imxrt-multi/config.h
```diff
@@ -766,6 +766,63 @@
 #endif
 #endif
 
+#if !ISEMPTY(UART_CONSOLE)
+#if UART_CONSOLE <= 0
```
Darchiv, line 770: These two ifs may be squashed into one

multi/imxrt-multi/config.h
```diff
@@ -766,6 +766,63 @@
...
+#endif
+
+#ifndef RTT_CHANNEL1_BLOCKING
+#define RTT_CHANNEL1_BLOCKING 0
+#elif !ISBOOLEAN(RTT_CHANNEL1_BLOCKING)
+#error "RTT_CHANNEL1_BLOCKING must have a value of 0, 1, or be undefined"
+#endif
+
+
+#ifndef RTT_CHANNEL_CONSOLE
```
Darchiv, line 813: As above, I suggest using `RTT_CONSOLE_USER` to make board_config.h easier to read. There is no common `RTT_CONSOLE` define used, but maybe it's not needed for now, as an empty value is the default here

multi/imxrt-multi/config.h
jmaksymowicz: They can't. If `UART_CONSOLE` is empty, the expression `!ISEMPTY(UART_CONSOLE) && (UART_CONSOLE <= 0)` will result in an error `operator '<=' has no left operand`.

multi/imxrt-multi/config.h
jmaksymowicz: Done

multi/imxrt-multi/config.h
jmaksymowicz: Done

Darchiv: Is there a reason for using `RTT_0`, `RTT_0_BLOCKING` etc. notation instead of `RTT0` (as described previously: `RTTx`, `RTTx_BLOCKING`)? UARTs, SPIs, and other devices in config.h/board_config.h use this notation (e.g. `UART1`, `SPI2_SCK`)

jmaksymowicz: I overlooked this, now it's fixed.

Darchiv (approved):

PR #517, opened 2024-10-01
pc-ata: set IO port addresses based on pci BARs

adamgreloch:
Some hardware (tested on ThinkPad E540) use uncommon ATA IO port addresses instead of typical 0x1f0, 0x3f6, 0x1e8, 0x3ee. In such case, these addresses are often located in pci BAR0-BAR3 of the storage device. This change utilizes BARs via platformctl/pctl_get if nothing is found on common addresses

JIRA: RTOS-922

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

- [X] Tested by hand on: ia32 qemu and hw

## Checklist:

## Special treatment

### DIFF
--- a/storage/pc-ata/ata.c
+++ b/storage/pc-ata/ata.c
@@ -3,8 +3,8 @@
  *
  * Generic ATA controller driver
  *
- * Copyright 2012-2015, 2019, 2020 Phoenix Systems
- * Author: Marcin Stragowski, Kamil Amanowicz, Lukasz Kosinski
+ * Copyright 2012-2015, 2019, 2020, 2024 Phoenix Systems
+ * Author: Marcin Stragowski, Kamil Amanowicz, Lukasz Kosinski, Adam Greloch
  *
  * This file is part of Phoenix-RTOS.
  *
@@ -17,10 +17,14 @@
 #include <stdlib.h>
 #include <unistd.h>
 #include <string.h>
+#include <stdbool.h>
 
 #include <sys/io.h>
 #include <sys/list.h>
 #include <sys/threads.h>
+
+#include <sys/platform.h>
+#include <phoenix/arch/ia32/ia32.h>
 
 #include "ata.h"
 
@@ -31,6 +35,12 @@ ata_common_t ata_common;
 
 
 static ata_bus_t *buses;
+
+
+static const int pci_devClasses[2] = {
+	0x0101, /* IDE interface */
+	0x0105, /* ATA controller */
+};
 
 
 static uint32_t ata_readreg(void *base, uint8_t reg, uint8_t size)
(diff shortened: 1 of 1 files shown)


storage/pc-ata/ata.c
```diff
@@ -624,11 +638,72 @@ int ata_init(void)
...
+			}
+
+			if (ata_initbus((void *)(ata2_base | 0x1), (void *)(ata2_ctrl | 0x1), bus2) == EOK) {
+				bus2_found = 1;
+			}
+
+			if (buses != NULL) {
+				/* at least one bus found via pci config, success */
+				break;
+			}
```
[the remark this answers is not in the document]
adamgreloch: There seems to be some issue with clang-format-dff on CI. Both clang-format-15 and clang-format-18 output the same file, but the clang-format-diff-15 that is present in CI shows such error, while clang-format-diff-18 does not

storage/pc-ata/ata.c
```diff
@@ -624,11 +638,72 @@ int ata_init(void)
...
+
+	if (ata_initbus((void *)(ata1_base | 0x1), (void *)(ata1_ctrl | 0x1), bus1) == EOK) {
+		bus1_found = 1;
+	}
+
+	if (ata_initbus((void *)(ata2_base | 0x1), (void *)(ata2_ctrl | 0x1), bus2) == EOK) {
+		bus2_found = 1;
+	}
+
+	if (buses == NULL) {
```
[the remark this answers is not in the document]
adamgreloch: `ata_initbus` updates the `buses` if the bus initialization succeeds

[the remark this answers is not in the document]
agkaminski: My bad, it's global, right

storage/pc-ata/ata.c
```diff
@@ -624,11 +638,63 @@ int ata_init(void)
 		return -ENOMEM;
 	}
 
-	if (ata_initbus((void *)(ATA1_BASE | 0x1), (void *)(ATA1_CTRL | 0x1), bus1) < 0)
+	for (i = 0; i < sizeof(pci_devClasses) + 1; i++) {
+		if (i == 0) {
```
agkaminski, line 642: Way to remove this `if` (which kinda violates MISRA):
- init ata1Base etc before the loop
- break from loop on success
- if not success, then do the pctl stuff
- ???
- profit

adamgreloch: I think that organizing a loop in the prepare-then-init fashion is a bit more readable than the inverse, especially when there's has to be some way of skipping the init part if platformctl needs to be retried (i.e. when 0x0101 class device not present, but 0x0105 is). Currently this skip is realized by doing a `continue`, in inverse it may not be so clean

I wonder whether trying the initbuses for the common ports before the loop wasn't the best of all worlds :p

agkaminski: Ok, I see, this trick with continue is colliding with my idea. I think you're right that the original idea was better.

adamgreloch: Ok, I reverted to the original idea

agkaminski (approved):
LGTM

kemonats:
This PR has been merged but I have a small comment.
https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/5ba4288c9a011462be600893c32d17c3b654dd35/storage/pc-ata/ata.c#L657
`pci_devClasses` array has 2 elements not 8.
That is
```c
for (i = 0; i < sizeof(pci_devClasses) / sizeof(pci_devClasses[0]); i++) {
```

adamgreloch:
Uhh, thanks for catching this. I've issued a fix

PR #523, opened 2024-11-15
imxrt-multi: implement software controlled half-duplex UART

jmaksymowicz:
Disable receiver while transmission is in progress. Toggle GPIO in software when switching between transmit and receive mode.

## Description
This feature can be configured by adding a new line to `board_config.h` like this:
```
#define UART12_HALF_DUPLEX_GPIO .port = 4, .pin = 27, .onTx = 1
```
This results in signal `GPIO_MUX4_IO27` being manipulated, high state is output when transmitting and low when ready to receive. `.onTx = 0` would result in low state being output when transmitting and high when ready to receive.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: imxrt1170

## Checklist:

## Special treatment

### DIFF
--- a/multi/imxrt-multi/uart.c
+++ b/multi/imxrt-multi/uart.c
@@ -31,15 +31,15 @@
 #include "uart.h"
 
 
-#define UART1_POS 0
-#define UART2_POS (UART1_POS + UART1)
-#define UART3_POS (UART2_POS + UART2)
-#define UART4_POS (UART3_POS + UART3)
-#define UART5_POS (UART4_POS + UART4)
-#define UART6_POS (UART5_POS + UART5)
-#define UART7_POS (UART6_POS + UART6)
-#define UART8_POS (UART7_POS + UART7)
-#define UART9_POS (UART8_POS + UART8)
+#define UART1_POS  0
+#define UART2_POS  (UART1_POS + UART1)
+#define UART3_POS  (UART2_POS + UART2)
+#define UART4_POS  (UART3_POS + UART3)
+#define UART5_POS  (UART4_POS + UART4)
+#define UART6_POS  (UART5_POS + UART5)
+#define UART7_POS  (UART6_POS + UART6)
+#define UART8_POS  (UART7_POS + UART7)
+#define UART9_POS  (UART8_POS + UART8)
 #define UART10_POS (UART9_POS + UART9)
 #define UART11_POS (UART10_POS + UART10)
 #define UART12_POS (UART11_POS + UART11)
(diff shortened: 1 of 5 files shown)


anglov (commented):
(not a code review)
We check along with @mateuszkobak and solution works for us

multi/imxrt-multi/uart.c
```diff
@@ -818,6 +920,7 @@ int uart_init(void)
 		uart = &uart_common.uarts[i++];
 		uart->base = info[dev].base;
 		uart->dev_no = dev;
+		memcpy(&uart->halfDuplexAction, &uartHalfDuplex[dev], sizeof(uart->halfDuplexAction));
```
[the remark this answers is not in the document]
jmaksymowicz: Changed this to an assignment. Performance shouldn't be a concern, the struct is only 4 bytes.

multi/imxrt-multi/uart.c
```diff
@@ -107,13 +132,66 @@ static inline int uart_getTXcount(uart_t *uart)
...
+	msg.type = mtDevCtl;
+	msg.i.data = NULL;
+	msg.i.size = 0;
+	msg.o.data = NULL;
+	msg.o.size = 0;
+
+	imsg = (multi_i_t *)msg.i.raw;
+
+	imsg->gpio.type = gpio_set_dir;
+	imsg->gpio.dir.val = 1uL << action->pin;
```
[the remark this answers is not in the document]
jmaksymowicz: Fixed

multi/imxrt-multi/uart.c
```diff
@@ -107,13 +132,66 @@ static inline int uart_getTXcount(uart_t *uart)
 }
 
 
+static int gpio_setup(const uart_halfDuplexAction_t *action)
+{
+	msg_t msg;
```
nalajcie, line 137: always zero-fill message (either by basic/designated initializer or by memset).

Right now you're sending parts of your stack (possibly leaking secrets) to other process.

jmaksymowicz: In this code no messages are actually sent, it directly calls the `gpio_handleMsg` function.

Perhaps to avoid confusion I could make those functions of the GPIO driver part of its public interface.

multi/imxrt-multi/uart.c
```diff
@@ -107,13 +132,66 @@ static inline int uart_getTXcount(uart_t *uart)
 }
 
 
+static int gpio_setup(const uart_halfDuplexAction_t *action)
```
agkaminski, line 135: Name prefix does not match the module name - we use prefixes to avoid name collisions. Static will always win, but it's undesirable nevertheless.

multi/imxrt-multi/uart.c
agkaminski: @jmaksymowicz IMHO good idea, would be more efficient than packing/unpacking msg

multi/imxrt-multi/uart.c
```diff
@@ -92,6 +101,22 @@ static const int uartPos[] = { UART1_POS, UART2_POS, UART3_POS, UART4_POS, UART5
...
+static const uart_halfDuplexAction_t uartHalfDuplex[] = {
+	{ UART1_HALF_DUPLEX_GPIO },
+	{ UART2_HALF_DUPLEX_GPIO },
+	{ UART3_HALF_DUPLEX_GPIO },
+	{ UART4_HALF_DUPLEX_GPIO },
+	{ UART5_HALF_DUPLEX_GPIO },
+	{ UART6_HALF_DUPLEX_GPIO },
+	{ UART7_HALF_DUPLEX_GPIO },
+	{ UART8_HALF_DUPLEX_GPIO },
+	{ UART9_HALF_DUPLEX_GPIO },
```
ziemleszcz, line 113: UART9-12 are available only on rt1176.

multi/imxrt-multi/uart.c
```diff
@@ -142,21 +220,37 @@ static void uart_intrThread(void *arg)
...
 				}
 			}
 
-			*(uart->base + ctrlr) |= (1 << 27) | (1 << 26) | (1 << 25) | (1 << 21);
+			if ((uart->halfDuplexAction.port != 0) && (rxActive == 0)) {
+				uint32_t status = *(uart->base + statr);
+				if ((status & (1 << 22)) != 0u) {
+					/* Transmission finished, activate RX */
+					gpio_setPin(&uart->halfDuplexAction, 0);
+					*(uart->base + ctrlr) |= (1 << 18);
```
ziemleszcz, line 244: Is there some reason for not setting it through `intrs`?

jmaksymowicz: Because this bit is not interrupt related - I wanted `intrs` to specifically contain bits that turn on interrupts.

multi/imxrt-multi/uart.c
jmaksymowicz: Fixed

agkaminski:
LGTM

multi/imxrt-multi/uart.c
Darchiv: UART9-12 are disabled by config.h when not on rt117x, so these defines don't hurt

ziemleszcz: They consume memory.

Darchiv: As do `uartConfig` and `uartPos`. Either all three should be rewritten or none at all, for consistency.

I think `uartConfig`, `uartPos`, `uartHalfDuplex`, and `uart_init():info` can all be merged into one `uartPreConfig` array, just like in i2c.c. @jmaksymowicz What do you think?

ziemleszcz: I’m closing the topic since it doesn’t matter for the rt117x, but it might be worth adding a TODO note indicating that this could be optimized in the future.

Darchiv (approved):

PR #525, opened 2024-11-22
Adapt USB stack to work on ia32

adamgreloch:
JIRA: RTOS-937

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

## Checklist:

## Special treatment

- [X] This PR needs additional PRs to work (list the PRs, preferably in merge-order).
https://github.com/phoenix-rtos/phoenix-rtos-usb/pull/34


usb/ehci/ehci.c
```diff
@@ -34,32 +34,42 @@
...
 
 
 static inline void ehci_memDmb(void)
 {
-	asm volatile ("dmb" ::: "memory");
+#ifdef EHCI_IMX
+	asm volatile("dmb" ::: "memory");
+#else
+	__sync_synchronize();
+#endif
```
agkaminski, line 50: https://godbolt.org/z/1q4s7Gsbe
```suggestion
	__sync_synchronize();
```

agkaminski (commented):
General remark - might be nice to separate platform dependent parts in a per arch files (ehci-ia32-generic.c etc?)

usb/ehci/ehci.c
adamgreloch: GCC won't reduce `__sync_synchronize()` to `dmb` prior to 9.3.1, see e.g. 9.2.1 https://godbolt.org/z/orPnWevvn

adamgreloch:
> General remark - might be nice to separate platform dependent parts in a per arch files (ehci-ia32-generic.c etc?)

I agree and intend on doing it once I get the EHCI to work on the hardware, since for now it's uncertain which parts of the driver will end up being common. The proposed changes work on QEMU, but as it turned out during my recent tests on hardware (which weren't possible before, because I had to find the controllers' IRQs via ACPI interpretation first), EHCI emulation in QEMU was extremely permissive. I am therefore moving this PR back to draft until I fix the hardware

usb/ehci/phy-ia32-generic.c
```diff
@@ -0,0 +1,175 @@
...
+int hcd_getInfo(const hcd_info_t **info)
+{
+	platformctl_t pctl;
+	int i, err;
+
+	hcd_info_t *hcd_info = NULL;
+
+	pctl.action = pctl_get;
+	pctl.type = pctl_pci;
+	pctl.pci.id.cl = 0x0c03;
```
Darchiv, line 39: What is this cl/class?

usb/ehci/phy-ia32-generic.c
```diff
@@ -0,0 +1,175 @@
...
+
+	pctl.pcicfg.cfg = pci_cfg_interruptdisable;
+	pctl.pcicfg.enable = 0;
+
+	err = platformctl(&pctl);
+	if (err < 0) {
+		fprintf(stderr, "phy: enabling interrupts failed: %d\n", err);
+	}
+
+	eecp = (*(uint32_t *)((char *)hcd->base + 0x8) >> 8) & 0xFF;
```
Darchiv, line 134: Is `+ 0x8` a specific register's offset? If so, maybe add an enum for it?

usb/ehci/phy-ia32-generic.c
adamgreloch: Serial bus controller class. Added a proper define

usb/ehci/phy-ia32-generic.c
adamgreloch: It's an offset inside the HCCPARAMS register pointing to EECP. Added a local define, as this pointer is unlikely to be used anywhere else than this file

usb/ehci/ehci.c
Darchiv: For future reference: we are in the process of migrating to GCC 14, so @agkaminski's suggestion can be applied once we move completely and probably in a new PR (separating platform-dependent parts in per-arch files).

Darchiv (approved):

PR #535, opened 2025-02-27
tty: extend baudrate limit checked in set_baudrate()

gmaz-phoesys:
## Description

This commit extends allowable range of baudrates accepted by uart16550 driver from 115200 to 460800.

## Motivation and Context

This extension was needed for the purpose of Processor-in-the-loop implementation on zynq7000 platform. Maximum baudrate of 115200 caused significant delay in data frames transmission.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7a9-zynq7000-zturn.

## Checklist:

## Special treatment

### DIFF
--- a/tty/uart16550/uart16550.c
+++ b/tty/uart16550/uart16550.c
@@ -76,14 +76,18 @@ static void set_baudrate(void *_uart, speed_t baud)
 {
 	uint8_t reg;
 	uart_t *uart = (uart_t *)_uart;
-	int baud_rate = libtty_baudrate_to_int(baud);
+	int32_t baud_rate = libtty_baudrate_to_int(baud);
 
-	if (baud_rate <= 0 || baud_rate > 115200) {
+	if (baud_rate <= 0) {
 		return;
 	}
 
 	/* Baud divisor */
 	baud_rate = uart->clk / (16 * baud_rate);
+
+	if (baud_rate > UINT16_MAX) {
+		return;
+	}
 
 	reg = uarthw_read(uart->hwctx, REG_LCR);
 


niewim19:
For us it works, LGTM from me, however @Darchiv  this does not solve the problem. Maybe the check should be done on register value being set (if it fits, and only if not - reject the setting)

tty/uart16550/uart16550.c
```diff
@@ -78,7 +78,7 @@ static void set_baudrate(void *_uart, speed_t baud)
 	uart_t *uart = (uart_t *)_uart;
 	int baud_rate = libtty_baudrate_to_int(baud);
 
-	if (baud_rate <= 0 || baud_rate > 115200) {
+	if (baud_rate <= 0 || baud_rate > 460800) {
```
Darchiv, line 81: I agree with @niewim19 - the ceiling condition can be omitted and then `baud_rate` (after division) checked if it fits `uint16_t`.

tty/uart16550/uart16550.c
```diff
@@ -76,15 +76,15 @@ static void set_baudrate(void *_uart, speed_t baud)
...
-
-	if (baud_rate <= 0 || baud_rate > 115200) {
-		return;
-	}
+	int32_t baud_rate = libtty_baudrate_to_int(baud);
 
 	/* Baud divisor */
 	baud_rate = uart->clk / (16 * baud_rate);
 
+	if (baud_rate <= 0 || baud_rate > UINT16_MAX) {
```
agkaminski, line 84: Baudrate < 0 should be checked before regval calculation. 0 is a valid value to write (baudrate == 0 is not ofc)

agkaminski (approved):
LGTM

tty/uart16550/uart16550.c
niewim19: Yes, you`re right. To add more - `libtty_baudrate_to_int(B0)` returns 0 so it would crash the driver (I didn't expect B0 to be even possible)

PR #540, opened 2025-03-21
zynqmp: add CAN driver

dsabala:
## Description
Created a simple ZynqMP CAN driver. 

## Motivation and Context

Support of CAN peripheral in Zynq Ultrascale devices

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: ZCU104 evaluation board with manual test app included in this PR

## Checklist:

## Special treatment

- [] This PR needs additional PRs to work:

### DIFF
--- a/_targets/Makefile.aarch64a53-zynqmp
+++ b/_targets/Makefile.aarch64a53-zynqmp
@@ -6,4 +6,4 @@
 # Copyright 2024 Phoenix Systems
 #
 
-DEFAULT_COMPONENTS := zynq-uart zynq-flash
+DEFAULT_COMPONENTS := zynq-uart zynq-flash zynqmp-can
--- a/can/zynqmp-can/zynqmp-can-if.c
+++ b/can/zynqmp-can/zynqmp-can-if.c
@@ -1,0 +1,80 @@
+/*
+ * Phoenix-RTOS
+ *
+ * ZynqMP CAN driver interface library
+ *
+ * Copyright 2025 Phoenix Systems
+ * Author: Dariusz Sabala
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <errno.h>
+#include <stddef.h>
+#include <stdbool.h>
+#include <sys/types.h>
+#include <sys/msg.h>
+#include <string.h>
+
+#include "zynqmp-can-if.h"
+#include "zynqmp-can-priv.h"
+
+int zynqmp_canSend(oid_t *oid, const zynqmp_canFrame *frames, uint8_t count, bool block, uint32_t *framesSent)
+{
+	/* Allocate message on the stack */
+	msg_t msg = { 0 };
+	/* Use custom request type */
+	msg.type = mtDevCtl;
+	/* Pass extra information about the request in the "raw" block */
+	zynqmp_canDriverReq *req = (zynqmp_canDriverReq *)msg.i.raw;
+	req->operation = zynqmp_canOpSend;
+	req->block = (uint8_t)block;
+	/* Pass frames using the data/size pair */
+	msg.i.data = (const void *)frames;
+	msg.i.size = count * sizeof(zynqmp_canFrame);
+
+	/* Try flushing messages to the CAN TX FIFO */
+	int ret = msgSend(oid->port, &msg);
+	/* If msgSend responded with no error, then return the error code from the driver */
+	if (ret == 0) {
+		ret = msg.o.err;
+	}
+
+	/* Copy the number of sent frames */
+	if (framesSent != NULL) {
+		zynqmp_canDriverResp *response = (zynqmp_canDriverResp *)msg.o.raw;
+		*framesSent = response->framesReceivedSend;
+	}
+
+	return ret;
+}
+
+int zynqmp_canRecv(oid_t *oid, zynqmp_canFrame *buf, uint32_t bufLen, bool block, uint32_t timeoutUs, uint32_t *recvFrames)
+{
+	/* Allocate message on the stack */
+	msg_t msg = { 0 };
+	/* Use custom request type */
+	msg.type = mtDevCtl;
+	/* Pass extra information about the request in the "raw" block */
+	zynqmp_canDriverReq *req = (zynqmp_canDriverReq *)msg.i.raw;
+	req->operation = zynqmp_canOpRecv;
+	req->block = (uint8_t)block;
+	req->timeoutUs = timeoutUs;
+	/* Pass received frames using the data/size pair */
+	msg.o.data = (void *)buf;
+	msg.o.size = bufLen * sizeof(zynqmp_canFrame);
+
+	/* Try to read messages from the CAN RX FIFO */
+	int ret = msgSend(oid->port, &msg);
+	/* If msgSend responded with no error, then return the error code from the driver */
+	if (ret == 0) {
+		ret = msg.o.err;
+	}
+
+	/* Copy the number of received frames */
+	zynqmp_canDriverResp *response = (zynqmp_canDriverResp *)msg.o.raw;
+	*recvFrames = response->framesReceivedSend;
+
+	return ret;
+}
--- a/can/zynqmp-can/zynqmp-can-if.h
+++ b/can/zynqmp-can/zynqmp-can-if.h
@@ -1,0 +1,75 @@
+/*
+ * Phoenix-RTOS
+ *
+ * ZynqMP CAN driver interface library
+ *
+ * Copyright 2025 Phoenix Systems
+ * Author: Dariusz Sabala
+ *
+ * %LICENSE%
+ */
+
+#ifndef ZYNQMP_CAN_IF_H
+#define ZYNQMP_CAN_IF_H
+
+#include <stdint.h>
+#include <stdbool.h>
+
+#include <sys/types.h>
+
+/* Maximum possible CAN standard frame ID */
+#define ZYNQMP_CAN_ID_MAX (0x7ff)
+
+/* Structure used for passing CAN frames to and from the CAN driver */
+typedef struct {
+	uint16_t id;  /**< CAN 2.0 Standard ID [0x000 ... 0x7ff] */
+	uint16_t len; /**< CAN 2.0 Payload length [0 ... 8] */
+	union {
+		uint8_t bytes[8];  /**< Payload represented as an array of bytes */
+		uint32_t words[2]; /**< Payload represented as an array of words */
+	} payload;             /**< CAN 2.0 Payload (up to 8 bytes) */
+} zynqmp_canFrame;
+
+/**
+ * Send CAN frames
+ *
+ * Description: This function attempts to flush the provided CAN frames into the 64 frame wide
+ *              hardware TX FIFO buffer
+ *
+ * Parameters:
+ * - oid: Object identifier of the related device file (either /dev/can0 or /dev/can1)
+ * - frames: Pointer to an array of CAN frames to be sent
+ * - count: Number of CAN frames in the provided array (accepts up to 64 frames)
+ * - block: Determines whether the function will block if there is insufficient space in the TX FIFO
+ * - framesSent: Pointer to store the number of frames sent (useful for non-blocking calls; can be NULL for blocking calls)
+ *
+ * Return values:
+ * - 0: Operation successful
+ * - -EINVAL: Invalid argument
+ * - -EAGAIN: Blocking is disabled, and there is no space in the TX buffer
+ */
+int zynqmp_canSend(oid_t *oid, const zynqmp_canFrame *frames, uint8_t count, bool block, uint32_t *framesSent);
+
+/**
+ * Receive CAN frames
+ *
+ * Description: This function attempts to retrieve received CAN frames from the 64 frame wide
+ *              hardware RX FIFO buffer. If the buffer is empty, it can block for the specified timeout,
+ * 				or return immediately
+ *
+ * Parameters:
+ * - oid: Object identifier of the related device file (either /dev/can0 or /dev/can1)
+ * - buf: Pointer to the buffer where received frames will be stored
+ * - bufLen: Length of the buffer (maximum number of CAN frames it can hold)
+ * - block: Determines whether the function will block if there are no frames in RX buffer
+ * - timeoutUs: Timeout in microseconds to wait for frames
+ * - recvFrames: Pointer to store the number of frames received
+ *
+ * Return values:
+ * - 0: Operation successful
+ * - -ETIME: Timeout expired
+ * - -EINVAL: Invalid argument
+ */
+int zynqmp_canRecv(oid_t *oid, zynqmp_canFrame *buf, uint32_t bufLen, bool block, uint32_t timeoutUs, uint32_t *recvFrames);
+
+#endif /* ZYNQMP_CAN_IF_H */
--- a/can/zynqmp-can/zynqmp-can-test.c
+++ b/can/zynqmp-can/zynqmp-can-test.c
@@ -1,0 +1,432 @@
+/*
+ * Phoenix-RTOS
+ *
+ * ZynqMP CAN driver test application
+ *
+ * Copyright 2025 Phoenix Systems
+ * Author: Dariusz Sabala
+ *
+ * %LICENSE%
+ */
+
+#include <stdio.h>
+#include <errno.h>
+#include <string.h>
+#include <stdlib.h>
+#include <termios.h>
+#include <stddef.h>
+#include <sys/types.h>
+#include <sys/threads.h>
+#include <sys/msg.h>
+#include <signal.h>
+#include <unistd.h>
+#include <sys/time.h>
+
+#include <zynqmp-can-if.h>
+
+/* Structure to store statistics for received frames */
+typedef struct {
+	uint32_t receivedFramesCount; /**< Number of received frames */
+	zynqmp_canFrame lastFrame;    /**< Last received frame */
+} frameStats;
+
+/* Application data */
+static struct {
+	uint8_t stack[_PAGE_SIZE] __attribute__((aligned(16)));
+	int run;                                   /**< flag to control the main loop */
+	handle_t mutex_stats_access;               /**< mutex for accessing frame statistics */
+	frameStats frame_stats[ZYNQMP_CAN_ID_MAX]; /**< Array to store statistics for each CAN ID */
+} common;
+
+/**
+ * Waits for a command input for a specified number of seconds.
+ */
+static int zynqmpCanTest_waitCmd(unsigned int secs)
+{
+	struct timeval left;
+	fd_set fds;
+
+	left.tv_sec = (time_t)secs;
+	left.tv_usec = 0;
+	fflush(stdout);
+	FD_ZERO(&fds);
+	FD_SET(STDIN_FILENO, &fds);
+	select(STDIN_FILENO + 1, &fds, NULL, NULL, &left);
+
+	if (FD_ISSET(STDIN_FILENO, &fds)) {
+		return getchar();
+	}
+
+	return 0;
+}
+
+/**
+ * Decodes and prints the contents of a CAN frame based on its ID.
+ */
+static void zynqmpCanTest_decodeAndPrintFrame(zynqmp_canFrame *frame)
+{
+	switch (frame->id) {
+		/* Placeholder for decoding logic */
+		default: {
+			/* Unknown frame content */
+			printf("\n");
+		} break;
+	}
+}
+
+/**
+ * Thread function to print CAN frame statistics periodically.
+ * Displays the received frame count and content for each CAN ID.
+ */
+static void zynqmpCanTest_statsPrintThread(void *arg)
+{
+	unsigned int delay = 1;
+
+	/* Clear terminal */
+	printf("\033[2J");
+
+	/* Disable cursor */
+	printf("\033[?25l");
+
+	while (common.run) {
+		/* Move cursor to the beginning */
+		printf("\033[f");
+		printf("\033[K");
+		printf("\033[2J");
+
+		mutexLock(common.mutex_stats_access);
+
+		printf("   ID | Count | Frame content \n");
+		for (uint32_t i = 0; i < ZYNQMP_CAN_ID_MAX; i++) {
+			if (common.frame_stats[i].receivedFramesCount != 0) {
+				printf("0x%3X |  %4i | ", i, common.frame_stats[i].receivedFramesCount);
+				zynqmpCanTest_decodeAndPrintFrame(&common.frame_stats[i].lastFrame);
+			}
+			common.frame_stats[i].receivedFramesCount = 0;
+		}
+
+		mutexUnlock(common.mutex_stats_access);
+
+		int cmd = zynqmpCanTest_waitCmd(delay);
+		if (cmd == 'q') {
+			common.run = 0;
+		}
+	}
+
+	endthread();
+}
+
+/**
+ * Sends two example CAN frames.
+ */
+static int zynqmpCanTest_SendTwoFrames(oid_t *oid)
+{
+	static const zynqmp_canFrame frames[] = {
+		{ .id = 0x100,
+			.len = 2,
+			.payload.bytes = { 0x21, 0x37 } },
+		{ .id = 0x200,
+			.len = 8,
+			.payload.bytes = { 0x11, 0x22, 0x33, 0x44, 0x55, 0x66, 0x77, 0x88 } },
+	};
+
+	/* Perform non-blocking send */
+	uint32_t framesSent = 0;
+	int ret = zynqmp_canSend(oid, frames, (sizeof(frames) / sizeof(frames[0])), false, &framesSent);
+
+	if (ret == 0) {
+		printf("zynqmp-can-test: Successfully sent %u CAN frames\n", framesSent);
+		return EXIT_SUCCESS;
+	}
+	else {
+		printf("zynqmp-can-test: Failed to send CAN frames, error %i\n", ret);
+		return EXIT_FAILURE;
+	}
+}
+
+/**
+ * Sends CAN frames with blocking enabled.
+ */
+static int zynqmpCanTest_SendBlocking(oid_t *oid)
+{
+	/* Generate example data: "enumerated" frames */
+	zynqmp_canFrame frames[64] = { 0 };
+	for (uint32_t i = 0; i < (sizeof(frames) / sizeof(frames[0])); i++) {
+		frames[i].id = 0x100;
+		frames[i].len = 1;
+		frames[i].payload.bytes[0] = i;
+	}
+
+	/* Send first 32 frames without problems and wait */
+	uint32_t framesSent = 0;
+	int ret = zynqmp_canSend(oid, frames, 32, true, &framesSent);
+	if (ret == 0) {
+		printf("zynqmp-can-test: Successfully sent %u CAN frames\n", framesSent);
+	}
+	else {
+		printf("zynqmp-can-test: Failed to send CAN frames, error %i\n", ret);
+		return EXIT_FAILURE;
+	}
+
+	/* Send frames again using the whole buffer; it should block but succeed */
+	ret = zynqmp_canSend(oid, frames, (sizeof(frames) / sizeof(frames[0])), true, &framesSent);
+	if (ret == 0) {
+		printf("zynqmp-can-test: Successfully sent %u CAN frames\n", framesSent);
+		return EXIT_SUCCESS;
+	}
+	else {
+		printf("zynqmp-can-test: Failed to send CAN frames, error %i\n", ret);
+		return EXIT_FAILURE;
+	}
+}
+
+/**
+ * Sends CAN frames without blocking.
+ */
+static int zynqmpCanTest_SendNonBlocking(oid_t *oid)
+{
+	/* Generate example data: "enumerated" frames */
+	zynqmp_canFrame frames[64] = { 0 };
+	for (uint32_t i = 0; i < (sizeof(frames) / sizeof(frames[0])); i++) {
+		frames[i].id = 0x100;
+		frames[i].len = 1;
+		frames[i].payload.bytes[0] = i;
+	}
+
+	/* Send first 32 frames without problems and without waiting */
+	uint32_t framesSent = 0;
+	int ret = zynqmp_canSend(oid, frames, 32, false, &framesSent);
+	if (ret == 0) {
+		printf("zynqmp-can-test: Successfully sent %u CAN frames\n", framesSent);
+	}
+	else {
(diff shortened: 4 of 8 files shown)


can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+			timeCfg[timeCfgN].baudrate_kbps,
+			timeCfg[timeCfgN].sample);
+
+	return 0;
+}
+
+/**
+ * Swaps the endianness of a 32-bit value.
+ */
+uint32_t swap_endianness(uint32_t value)
```
jmaksymowicz, line 324: You can use `ntohl` from `arpa/inet.h` or `be32toh` from `endian.h`.

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+
+	left.tv_sec = (time_t)secs;
+	left.tv_usec = 0;
+	fflush(stdout);
+	FD_ZERO(&fds);
+	FD_SET(STDIN_FILENO, &fds);
+	select(STDIN_FILENO + 1, &fds, NULL, NULL, &left);
+
+	if (FD_ISSET(STDIN_FILENO, &fds))
+		return getchar();
```
agkaminski, line 60:
```suggestion
	if (FD_ISSET(STDIN_FILENO, &fds)) {
		return getchar();
	}
```

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+
+	return 0;
+}
+
+/**
+ * Decodes and prints the contents of a CAN frame based on its ID.
+ */
+static void decode_and_print_frame(zynqmp_canFrame *frame)
```
agkaminski, line 68: Use camelCase, perhaps file prefix? Not that important I guess for a test.
```suggestion
static void zynqmpCanTest_decodeAndPrintFrame(zynqmp_canFrame *frame)
```

can/zynqmp-can/zynqmp-can-if.c
```diff
@@ -0,0 +1,104 @@
...
+#include <stdbool.h>
+#include <sys/types.h>
+#include <phoenix/msg.h>
+#include <sys/msg.h>
+#include <string.h>
+
+#include "zynqmp-can-if.h"
+#include "zynqmp-can-priv.h"
+
+int zynqmp_can_send(oid_t *oid, zynqmp_canFrame *frames, uint32_t count, bool block, uint32_t *sendFrames)
```
agkaminski, line 24:
```suggestion
int zynqmp_can_send(oid_t *oid, const zynqmp_canFrame *frames, uint32_t count, bool block, uint32_t *sendFrames)
```

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+
+	endthread();
+}
+
+/**
+ * Sends two example CAN frames.
+ */
+static int send_two_frames(oid_t *oid)
+{
+	zynqmp_canFrame frames[] = {
```
agkaminski, line 126:
```suggestion
	static const zynqmp_canFrame frames[] = {
```

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+	tcsetattr(STDIN_FILENO, TCSANOW, &state);
+}
+
+/**
+ * Receives CAN frames in a loop with a timeout.
+ * Displays received frame statistics in real time.
+ */
+static int recv_in_loop(oid_t *oid)
+{
+	switchmode(0);
```
agkaminski, line 304: Standard idiom is to rather store initial termios state in a safe place and restore it at the end

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+
+	if (mutexCreate(&mutex_stats_access) != 0) {
+		printf("cantest: Failed to create mutex\n");
+		return -1;
```
agkaminski, line 308: Exiting with terminal not restored

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+	}
+
+	/* Allocate buffer for received frames */
+	static zynqmp_canFrame buf[64] = { 0 };
+	uint32_t recv_frames = 0;
+
+	/* Start thread to print results each second */
+	handle_t thread_hdl;
+	uint8_t stack[4096];
```
agkaminski, line 317: That's a lot, consider malloc or static

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+	}
+
+	/* Check if the correct number of arguments was passed */
+	if (argc != 2) {
+		printf("usage: zynqmp-can-test <operation>\n");
+		return -1;
+	}
+
+	/* Perform action based on command */
+	if (0 == strcmp("send", argv[1])) {
```
agkaminski, line 376: I get the idea behind reversed order in comparisons, but:
a) It's incoherent with our code base,
b) it's incoherent with this file (even the function)
c) gcc emits a warning on `=` instead of `==` for a few years now

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+
+#include <posix/utils.h>
+
+#include <phoenix/arch/aarch64/zynqmp/zynqmp.h>
+
+#include "zynqmp-can-priv.h"
+#include "zynqmp-can-if.h"
+
+/* Macro to calculate the length of an array */
+#define ARRAY_LENGTH(array) (sizeof((array)) / sizeof((array)[0]))
```
agkaminski, line 38: Please replace with explicit `sizeof`s

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+ * Setting this to low values (e.g., 1) reduces the blocking time for the sending function
+ * but increases the frequency of interrupts.
+ */
+#define CAN_TX_FIFO_WATERMARK 4
+
+/* Default CAN baud rate in kbps */
+#define CAN_BAUDRATE_KBPS_DEFAULT (1000)
+
+/* Base addresses for ZynqMP CAN peripherals */
+#define ZYNQMP_CAN0_BASE (0x00FF060000)
```
agkaminski, line 51: Style nitpick - use lowecase hex
```suggestion
#define ZYNQMP_CAN0_BASE (0x00ff060000)
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+#define ZYNQMP_CAN1_BASE (0x00FF070000)
+
+/* CAN peripheral instances */
+enum { can_0 = 0,
+	can_1 = 1,
+	canInstancesCount = 2 };
```
agkaminski, line 57: Used only for comparing with a number - number of instances define would suffice

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+
+/* Structure representing the CAN peripheral registers */
+typedef struct {
+	uint32_t SRR;             /* Software Reset and Enable, Address offset:  0x0 */
```
agkaminski, line 61: All-uppercase identifiers are reserved for macros only

agkaminski, line 61: All fields should be volatile
```suggestion
	volatile uint32_t srr;             /* Software Reset and Enable, Address offset:  0x0 */
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+} can_periph_t;
+
+/* Data structure used by the driver */
+static struct {
+	volatile can_periph_t *base; /**< Pointer to the peripheral base address */
+	oid_t oid;                   /**< Object identifier for the related device file */
+	handle_t cond;               /**< Conditional variable for synchronizing interrupts */
+	handle_t inth;               /**< Interrupt handler object */
+	handle_t lock;               /**< Mutex used with the conditional variable for synchronization */
+} can = { 0 };
```
agkaminski, line 103: Initialized is redundant
```suggestion
} can;
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+	{ .baudrate_kbps = 50, .sample = "87.5", .brpr = 0x7C, .btr = 0x05 },
+	{ .baudrate_kbps = 100, .sample = "90.0", .brpr = 0x31, .btr = 0x07 },
+	{ .baudrate_kbps = 125, .sample = "87.5", .brpr = 0x31, .btr = 0x05 },
+	{ .baudrate_kbps = 250, .sample = "87.5", .brpr = 0x18, .btr = 0x05 },
+	{ .baudrate_kbps = 500, .sample = "90.0", .brpr = 0x09, .btr = 0x07 },
+	{ .baudrate_kbps = 1000, .sample = "90.0", .brpr = 0x04, .btr = 0x07 },
+};
+
+/* Function to display command-line help */
+static void cli_help(const char *progname)
```
agkaminski, line 134: All functions, even static ones, shall have relevant prefix. Use camelCase
```suggestion
static void zynqmpCan_cliHelp(const char *progname)
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+	if (init_clock(can_id) < 0) {
+		debug("zynqmp-can: cannot initialize clocks\n");
+		return EXIT_FAILURE;
+	}
+
+	/* Configure MIO pins */
+	if (config_mio_pin(info[can_id].rxPin) < 0 || config_mio_pin(info[can_id].txPin) < 0) {
+		debug("zynqmp-can: failed to initialize MIO pins\n");
+		return -EINVAL;
+	}
```
agkaminski, line 267: Perhaps move it to a later position? There's a lot that can go wrong after this and there's no pins uninit anywhere

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+	/* Reset the CAN peripheral */
+	if (reset_can_peripheral(can_id) < 0) {
+		debug("zynqmp-can: failed to reset CAN peripheral\n");
+		return -EINVAL;
+	}
+
+	/* Wait for configuration mode */
+	do {
+		reg_val = can.base->SR & (1 << 0);
+	} while (reg_val != 1);
```
agkaminski, line 298: Less CPU hogging 
```suggestion
	while ((can.base->SR & (1 << 0)) != 1) {
		usleep(100);
	}
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+				can.base->IER &= ~(1 << 13);
+				can.base->ICR = (1 << 13);
+				can.base->IER |= (1 << 13);
+
+				/* Wait for "Transmit FIFO Watermark Empty" interrupt */
+				condWait(can.cond, can.lock, 0);
+			}
+			else {
+				/* Not enough space in the buffer */
+				return -ENOMEM;
```
agkaminski, line 355: More suitable for non-block would-block situation.
```suggestion
				return -EAGAIN;
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+		/* Check if TX FIFO is not full */
+		if (0U != (can.base->SR & (1 << 10))) {
+			if (block) {
+				/* Disable, clear, and enable "Transmit FIFO Watermark Empty" interrupt */
+				can.base->IER &= ~(1 << 13);
+				can.base->ICR = (1 << 13);
+				can.base->IER |= (1 << 13);
+
+				/* Wait for "Transmit FIFO Watermark Empty" interrupt */
+				condWait(can.cond, can.lock, 0);
```
agkaminski, line 351: We do not check if wakeup is spurious - if so we will put data in a full FIFO

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+{
+	*recvFrames = 0;
+
+	/* Disable, clear, and enable "Receive FIFO Not Empty" interrupt */
+	can.base->IER &= ~(1 << 7);
+	can.base->ICR = (1 << 7);
+	can.base->IER |= (1 << 7);
+
+	/* Wait for data in the RX FIFO (signaled by "Receive FIFO Not Empty" IRQ) */
+	int condRet = condWait(can.cond, can.lock, timeoutUs);
```
agkaminski, line 384: What if FIFO is already full? Shouldn't we check that before going to sleep? Do not depend on the stickiness of the condVar, it might've been cleared already.

condWait (to be used correctly) **has** to be used in a following way

`while (!condition) { condWait(cond, lock, timeout); }`

otherwise you can cause random deadlocks

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+	if (condRet == -ETIME) {
+		/* Disable "Receive FIFO Not Empty" interrupt */
+		can.base->IER &= ~(1 << 7);
```
agkaminski, line 387: Why are you disabling interrupts? We should still collect incoming data even if no thread is receiving it

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+					msg.o.err = send_frames(frames, count, req->blockTxOperation, &resp->framesReceivedSend);
+				}
+				else if (req->operation == zynqmp_can_opRecv) {
+					zynqmp_canFrame *buf = (zynqmp_canFrame *)msg.o.data;
+					uint32_t bufLen = msg.o.size / sizeof(zynqmp_canFrame);
+					msg.o.err = recv_frames(buf, bufLen, req->timeoutUs, &resp->framesReceivedSend);
+				}
+				else {
+					printf("zynqmp-can: unknown opcode\n");
+					msg.o.err = -ENOSYS;
```
agkaminski, line 459:
```suggestion
					msg.o.err = -EINVAL;
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,552 @@
...
+					for (uint32_t i = 0; i < ARRAY_LENGTH(timeCfg); i++) {
+						if (timeCfg[i].baudrate_kbps == baudrate) {
+							timeCfgFound = true;
+						}
+					}
+					if (!timeCfgFound) {
+						debug("zynqmp-can: failed to configure desired baud rate\n");
+						return EXIT_FAILURE;
+					}
+				} break;
```
agkaminski, line 508: Redundant, already checked in can_init

```suggestion
				case 'b':
					baudrate = atoi(optarg);
					break;
```

agkaminski (dismissed):

can/zynqmp-can/zynqmp-can.c
dsabala: Ok, I put the condWait calls in a while loop checking the condition and added another check criteria to now wait if RX FIFO is already non-empty

can/zynqmp-can/zynqmp-can.c
dsabala: I think this question was answered in another topic here

dsabala (commented):
Answered all questions, fixed code, tested on ZCU104 with manual test app.

can/zynqmp-can/zynqmp-can-if.c
```diff
@@ -0,0 +1,104 @@
...
+{
+	/* Validate arguments */
```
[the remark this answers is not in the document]
agkaminski: It is perfectly fine to crash the user app when the function is incorrectly used, but it is not fine at all to be able remotely crash the server - thus the verification should be performed on server side.

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,398 @@
...
+#include <phoenix/msg.h>
+#include <sys/threads.h>
+#include <sys/msg.h>
+#include <signal.h>
+#include <unistd.h>
+
+#include <zynqmp-can-if.h>
+
+/* Macro to calculate the length of an array */
+#define ARRAY_LENGTH(array) (sizeof((array)) / sizeof((array)[0]))
```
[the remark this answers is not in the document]
agkaminski: It never did, as it always has been forbidden :smile: Just a company policy, rationale: makes more difficult to make a classic error:
```
#define NLEMS(x) sizeof(x)/sizeof(*x)
(...)
void foo(int bar[]) {
(...)
size_t len = NELEMS(bar); /* Oopsie */
```

can/zynqmp-can/zynqmp-can.c
agkaminski: Ack, approach seems rational to me

_targets/Makefile.aarch64a53-zynqmp
```diff
@@ -6,4 +6,4 @@
 # Copyright 2024 Phoenix Systems
 #
 
-DEFAULT_COMPONENTS := zynq-uart zynq-flash
+DEFAULT_COMPONENTS := zynq-uart zynq-flash zynqmp-can zynqmp-can-test
```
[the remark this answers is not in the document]
dsabala: ok, thank you :)

can/zynqmp-can/zynqmp-can-if.c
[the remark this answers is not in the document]
dsabala: fixed, moved those checks into the server

can/zynqmp-can/zynqmp-can.c
```diff
@@ -322,6 +322,20 @@ static int zynqmpCan_init(unsigned int can_id, int baudrate)
  */
 static int zynqmpCan_sendFrames(zynqmp_canFrame *frames, uint32_t count, bool block, uint32_t *sendFrames)
 {
+	/* Validate arguments */
+	if (frames == NULL) {
+		fprintf(stderr, "zynqmp-can: null pointer argument\n");
```
agkaminski, line 327: I would like to propose not printing these errors - because of these user can maliciously spam error messages to the dmesg

can/zynqmp-can/zynqmp-can-if.h
```diff
@@ -0,0 +1,74 @@
...
+ * - bufLen: Length of the buffer (maximum number of CAN frames it can hold)
+ * - timeoutUs: Timeout in microseconds to wait for frames
+ * - recvFrames: Pointer to store the number of frames received
+ *
+ * Return values:
+ * - EOK: Operation successful
+ * - -ETIME: Timeout expired
+ * - -EINVAL: Invalid argument
+ */
+int zynqmp_canRecv(oid_t *oid, zynqmp_canFrame *buf, uint32_t bufLen, time_t timeoutUs, uint32_t *recvFrames);
```
Darchiv, line 72: Maybe add a non-blocking variant of this function, as is the case with send? Be it via `bool block` parameter of a separate `zynqmp_canRecvNb()` function

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,401 @@
...
+	termiosCfgNew.c_cc[VMIN] = 1;
+	if (tcsetattr(STDIN_FILENO, TCSANOW, &termiosCfgNew) < 0) {
+		fprintf(stderr, "cantest: failed to set termios mode\n");
+		return -EXIT_FAILURE;
+	}
+
+	if (mutexCreate(&common.mutex_stats_access) != 0) {
+		fprintf(stderr, "cantest: Failed to create mutex\n");
+		ret = -EXIT_FAILURE;
+		goto restore_termios;
```
agkaminski, line 307: `goto` is forbidded both by MISRA and our coding convention: https://docs.phoenix-rtos.com/latest/coding/index.html

can/zynqmp-can/zynqmp-can-test.c
```diff
@@ -0,0 +1,401 @@
...
+
+/* Application data */
+static struct {
+	uint8_t stack[_PAGE_SIZE] __attribute__((aligned(16)));
+	int run;                                   /**< flag to control the main loop */
+	handle_t mutex_stats_access;               /**< mutex for accessing frame statistics */
+	frameStats frame_stats[ZYNQMP_CAN_ID_MAX]; /**< Array to store statistics for each CAN ID */
+} common = {
+	.run = 1,
+};
```
agkaminski, line 41: Not super important, but this one, small initialized causes whore struct to be placed in .data, along with the zeroed-out stack. It means it is a part of the ELF and has to be copied on process start ;) How about placing `common.run = 1;` in `main()`?

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,565 @@
...
+		}
+	}
+
+	/* Read all received frames */
+	for (uint32_t i = 0; i < bufLen; i++) {
+		/* Try clearing the interrupt (it will only clear if there is no data) */
+		can.base->icr = (1 << 7);
+
+		/* Check if there is still data in the RX FIFO (by reading pending IRQ status) */
+		if (0 == (can.base->isr & (1 << 7))) {
```
agkaminski, line 428: One comparison survived

can/zynqmp-can/zynqmp-can-if.h
```diff
@@ -0,0 +1,74 @@
...
+#define ZYNQMP_CAN_IF_H
+
+#include <stddef.h>
+#include <stdbool.h>
+#include <sys/types.h>
+#include <sys/time.h>
+#include <phoenix/msg.h>
+
+/* Maximum possible CAN standard frame ID */
+#define ZYNQMP_CAN_ID_MAX (0x7FF)
```
agkaminski, line 22:
```suggestion
#define ZYNQMP_CAN_ID_MAX (0x7ff)
```

can/zynqmp-can/zynqmp-can-if.h
```diff
@@ -0,0 +1,74 @@
...
+
+/* Structure used for passing CAN frames to and from the CAN driver */
+typedef struct {
+	uint16_t id;  /**< CAN 2.0 Standard ID [0x000 ... 0x7FF] */
```
agkaminski, line 26:
```suggestion
	uint16_t id;  /**< CAN 2.0 Standard ID [0x000 ... 0x7ff] */
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -322,6 +322,20 @@ static int zynqmpCan_init(unsigned int can_id, int baudrate)
...
+	/* Validate arguments */
+	if (frames == NULL) {
+		fprintf(stderr, "zynqmp-can: null pointer argument\n");
+		return -EINVAL;
+	}
+	if (count == 0) {
+		fprintf(stderr, "zynqmp-can: zero messages to send\n");
+		return -EINVAL;
+	}
+	if (count > 64) {
```
Darchiv, line 334: If `count` is limited to 64, then why not use a more limited type, e.g. `uint8_t`? `uint32_t` suggests that the data range is large, and yet we will fail with even a value such as 2^16.

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,565 @@
...
+
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <endian.h>
+
+#include <sys/debug.h>
```
agkaminski, line 20:
```suggestion
```

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,565 @@
...
+	/* Parse command-line arguments */
+	if (argc > 1) {
+		while ((c = getopt(argc, argv, "n:b:h")) != -1) {
+			switch (c) {
+				case 'b': {
+					baudrate = atoi(optarg);
+				} break;
+
+				case 'n': {
+					can_id = atoi(optarg);
```
agkaminski, line 524: According to MISRA - avoid casting signed -> enum.

IMHO `can_0`, `can_1` enums serve little purpose, plain int suffices (we store the peripheral number in variable called `can_id`, it's plenty descriptive)

Darchiv (commented):
I have left a few nitpicks.

Did you add separate commits with review fixes to circumvent Github's flaky "see last changes" mechanism which includes other changes on master done in the meantime? Separate commits make it harder to review new code and track old comments, i.e. they are not shown.

Also, please use a verb in PR's title and commits' messages, as can be seen in commit history. For example "zynqmp: Add CAN driver"

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,565 @@
...
+					if ((can_id < can_0) || (can_id > can_1)) {
+						fprintf(stderr, "zynqmp-can: invalid CAN peripheral ID\n");
+						return -EXIT_FAILURE;
```
Darchiv, line 527: You don't have to make `EXIT_FAILURE` negative. In fact, in Linux exit codes are 0-255

can/zynqmp-can/zynqmp-can.c
dsabala: Sure, removed all trivial error messages

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,547 @@
...
+					uint32_t count = msg.i.size / sizeof(zynqmp_canFrame);
+					msg.o.err = zynqmpCan_framesSent(frames, count, req->block, &resp->framesReceivedSend);
+				}
+				else if (req->operation == zynqmp_canOpRecv) {
+					zynqmp_canFrame *buf = (zynqmp_canFrame *)msg.o.data;
+					uint32_t bufLen = msg.o.size / sizeof(zynqmp_canFrame);
+					msg.o.err = zynqmpCan_recvFrames(buf, bufLen, req->block, req->timeoutUs, &resp->framesReceivedSend);
+				}
+				else {
+					fprintf(stderr, "zynqmp-can: unknown opcode\n");
```
agkaminski, line 461:
```suggestion
```

can/zynqmp-can/zynqmp-can.c
agkaminski: What @Darchiv said references to the process exit status only

can/zynqmp-can/zynqmp-can.c
dsabala: fixed

can/zynqmp-can/zynqmp-can.c
dsabala: Ok this is clear explanation, thank you. Removed "-" sign before all `EXIT_FAILURE` references and used `EXIT_FAILURE` in all process exit places

agkaminski (dismissed):
LGTM

jmaksymowicz (approved):
A few nitpicks but otherwise LGTM

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,548 @@
...
+		munmap((void *)can.base, _PAGE_SIZE);
+		resourceDestroy(can.cond);
+		return -ENOENT;
+	}
+
+	/* Reset the CAN peripheral */
+	if (zynqmpCan_resetPeriphal(can_id) < 0) {
+		fprintf(stderr, "zynqmp-can: failed to reset CAN peripheral\n");
+		return -EINVAL;
+	}
```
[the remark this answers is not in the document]
dsabala: sure, fixed

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,548 @@
...
+
+	/* Program the bit sampling clock */
+	can.base->brpr = timeCfg[timeCfgN].brpr;
+	can.base->btr = timeCfg[timeCfgN].btr;
+
+	/* Configure the watermark for TX FIFO */
+	can.base->wir = (CAN_TX_FIFO_WATERMARK << 8);
+
+	/* Register the interrupt handler */
+	interrupt(info[can_id].irq, zynqmpCan_irqHandler, &can, can.cond, &can.inth);
```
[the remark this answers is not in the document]
dsabala: fixed

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,560 @@
...
+	uint32_t timeCfgN = 0;
+	for (uint32_t i = 0; i < (sizeof(timeCfg) / sizeof(timeCfg[0])); i++) {
+		if (timeCfg[i].baudrate_kbps == baudrate) {
+			timeCfgFound = true;
+			timeCfgN = i;
+		}
+	}
+	if (!timeCfgFound) {
+		fprintf(stderr, "zynqmp-can: failed to configure desired baud rate\n");
+		return EXIT_FAILURE;
```
agkaminski, line 243: Why change return values? `EXIT_FAILURE == 1`, this won't be recognized as error in `main()`

dsabala: I wanted to stay true to the convention that the process returns positive numbers as an error indicator, I changed the comparison operator, now it should be ok

agkaminski: No, the convention is to return negative values on error. The only exception is a exit status of a process, that per POSIX has to be a value from 0 to 255 inclusively

agkaminski (dismissed):

can/zynqmp-can/zynqmp-can.c
dsabala: ok, reverted error return values to previous version

Darchiv:
> Also, please use a verb in PR's title and commits' messages, as can be seen in commit history. For example "zynqmp: Add CAN driver"

Please also update PR's title

agkaminski (commented):
We'll, return values hasn't been restored, these changed.

can/zynqmp-can/zynqmp-can.c
```diff
@@ -0,0 +1,560 @@
...
+	uint32_t timeCfgN = 0;
+	for (uint32_t i = 0; i < (sizeof(timeCfg) / sizeof(timeCfg[0])); i++) {
+		if (timeCfg[i].baudrate_kbps == baudrate) {
+			timeCfgFound = true;
+			timeCfgN = i;
+		}
+	}
+	if (!timeCfgFound) {
+		fprintf(stderr, "zynqmp-can: failed to configure desired baud rate\n");
+		return -EINVAL;
```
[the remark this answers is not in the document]
dsabala: Sure, replaced all error returns to `return -1` in this function

PR #545, opened 2025-04-07
sensors: add project specific sensors linking

niewim19:
## Description

Adding possibility to link sensors implementations from outside of phoenix-rtos-devices. If user wishes to do this, he must double-build the phoenix-rtos-devices: once as default, and later again, after additional sensor building, with correct `DEVICE_SENSORS` variable name.

As there are now two headers installed from libsensors, header names refactor was implemented, and all API of libsensors, no matter if libsensor user or sensor implementation are now available as `<libsensors/client.h>` or `<libsensors/sensor.h>`.

This is of course breaking change, and all the names of headers must be now change, this commit takes care of all usages within `phoenix-rtos-devices` as a separate commit.

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment


sensors/Makefile
```diff
@@ -22,7 +22,8 @@ SENSORS_SIM := imu/imu_sim.c mag/mag_sim.c baro/baro_sim.c gps/gps_sim.c gps/gps
 SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(SENSORS_ALL), $(DEVICE_SENSORS)))
 
 NAME := sensors
-LOCAL_HEADERS_DIR := nothing
+
+LIBS := $(SENSORS_PROJECT_DEPS) # Project-specific sensors .a files and their dependencies
```
nalajcie, line 26: not related to this change, but:
 - `SENSORS_SIM` files are included twice in `LOCAL_SRCS` .
 - IMHO we should use wildcards here instead of hand-listing every c file

sensors/Makefile
```diff
@@ -11,7 +11,7 @@ include $(static-lib.mk)
 # Sensors SPI interface
 NAME := libsensors-spi
 LOCAL_HEADERS := sensors-spi.h
-LOCAL_HEADERS_DIR := nothing  # do not reinstall libsensors.h
+LOCAL_HEADERS_DIR := nothing  # do not reinstall libsensors/client.h
```
nalajcie, line 14: also `sensors.h` - so the comment is still incorrect.

not related to this PR but still - `libsensors-spi` should be a separate component with different HAL implementations (or drivers should use common SPI API if it would ever be available), not a c file withing the same dir with hard-coded zynq API.

sensors/gps/common.h
```diff
@@ -22,7 +22,7 @@
 #include <stdlib.h>
 
 #include "nmea.h"
-#include "../sensors.h"
+#include <libsensors/sensor.h>
```
nalajcie, line 25: include before the local header?

sensors/baro/lps25xx.c
```diff
@@ -20,7 +20,7 @@
 #include <spi.h>
 #include <sensors-spi.h>
 
-#include "../sensors.h"
+#include <libsensors/sensor.h>
```
nalajcie, line 23: probably we should update the header guards in `sensor.h` and `client.h` (include `LIBSENSORS` prefix?)

nalajcie (changes requested):
- use imperative mood in commit message body also
- IMHO this should be a single commit - you're splitting atomic change into 2 commits (or maybe better: do a header file renames + `include` updates in one commit; add external sensors linking in separate commit - those are separate, unrelated changes)
- commits with breaking change need to have `!` as a first char in commit message title

sensors/Makefile
niewim19: Fixed the comment on reinstalling.

About the libsensors.spi - yeah I am aware of it. As for now I have created the issue:
 - https://github.com/phoenix-rtos/phoenix-rtos-project/issues/1289

But I do not plan on solving this within this PR for now.

sensors/baro/lps25xx.c
niewim19: Yes, added correct naming in guards.

sensors/Makefile
niewim19: - `SENSORS_SIM` fix included in one of commits as it is a quick fix
 - added commits that use wildcards for sensor selection

sensors/Makefile
```diff
@@ -22,7 +22,8 @@ SENSORS_SIM := imu/imu_sim.c mag/mag_sim.c baro/baro_sim.c gps/gps_sim.c gps/gps
 SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(SENSORS_ALL), $(DEVICE_SENSORS)))
 
 NAME := sensors
-LOCAL_HEADERS_DIR := nothing
```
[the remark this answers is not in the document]
niewim19: My mistake, yes this should remain. Brought back, thanks!

niewim19:
>  use imperative mood in commit message body 
I forget that this rule applies also to body - thanks

> IMHO this should be a single commit - you're splitting atomic change into 2 commits (or maybe better: do a header file renames + include updates in one commit; add external sensors linking in separate commit - those are separate, unrelated 
Good idea - I split changes in the proposed fashion:
 - commit with header renames + updates across this repository
 - commit with external sensor linking

>  commits with breaking change need to have ! as a first char in commit message title
 Applied!

sensors/Makefile
```diff
@@ -11,19 +11,25 @@ include $(static-lib.mk)
...
-SENSORS_SIM := imu/imu_sim.c mag/mag_sim.c baro/baro_sim.c gps/gps_sim.c gps/gps_sim.c
-SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(SENSORS_ALL), $(DEVICE_SENSORS)))
+
+#Find all local sensor files
+SENSOR_DIRS := imu/ mag/ baro/ gps/
+SENSOR_SRCS := $(patsubst sensors/%, %, $(wildcard $(addsuffix *.c, $(addprefix sensors/, $(SENSOR_DIRS)))))
+
+SENSORS_SIM := $(filter %sim.c, $(SENSOR_SRCS))
+SENSORS_HW := $(filter-out %sim.c, $(SENSOR_SRCS))
+SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(patsubst %.c, %, $(SENSORS_HW)), $(DEVICE_SENSORS)))
```
nalajcie, line 27: after some thinking:
- we have 2 separate mechanisms to select the drivers:
  - one by *removing* some `.c` files from the default codebase 
  - one by *adding* extra static libs at a linking stage
- maybe we should use one mechanism
  - compile current drivers into static libs
  - provide default static libs list which might be overriden on the project basis
  - (that way you're removing the driver from the binary by just not-including it in the libraries list)

Maybe we should do it now - as we're already introducing a breaking change?

agkaminski: +1

sensors/Makefile
```diff
@@ -11,19 +11,25 @@ include $(static-lib.mk)
...
 # Sensors Server
-SENSORS_ALL := imu/lsm9dsxx imu/mpu6000 mag/lsm9dsxx mag/lis2mdl baro/lps25xx baro/ms5611 gps/pa6h gps/ubx
-SENSORS_SIM := imu/imu_sim.c mag/mag_sim.c baro/baro_sim.c gps/gps_sim.c gps/gps_sim.c
-SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(SENSORS_ALL), $(DEVICE_SENSORS)))
+
+#Find all local sensor files
+SENSOR_DIRS := imu/ mag/ baro/ gps/
+SENSOR_SRCS := $(patsubst sensors/%, %, $(wildcard $(addsuffix *.c, $(addprefix sensors/, $(SENSOR_DIRS)))))
+
+SENSORS_SIM := $(filter %sim.c, $(SENSOR_SRCS))
```
dsabala, line 25: @nalajcie @agkaminski @niewim19 

This might not be the best idea, but perhaps we could consider removing the entire "sensors" directory from the open-source repository.

How much of the code tightly coupled with libsensors is already in the closed-source part of the project, and how much remains in the open-source repository? For instance, the libsensc module, which appears to work in a "client-server" pair with libsensors, is part of the closed-source codebase and is already quite closely integrated with the autopilot.

From the perspective of a former external user of Phoenix-RTOS (a scenario I can easily relate to, as in my previous role, I had a ticket in the backlog to create a proof of concept for Phoenix-RTOS), I would likely be very eager to explore and experiment with the system. However, I would probably choose to bypass libsensors entirely and develop my own abstraction to achieve similar functionality. This approach would allow me to avoid the cognitive overhead of understanding how libsensors works - IMO its abstractions seem somewhat complex. Instead, I would focus on gaining hands-on experience with Phoenix-RTOS, particularly its message-passing mechanisms, which would be both more enjoyable and significantly more interesting to explore.

In other words: in my opinion the greatest strength of this module is the ability to use it integrated with other high abstraction modules, and this is not available for open source right now.

niewim19: Thanks for the thoughtful feedback! You're absolutely right that the full value of libsensors and sensorhub comes with high-level components integration — like the ones you mentioned which are currently closed-source. However, we’d still prefer to keep sensorhub and the broader sensors directory in the open-source tree for a few reasons:

    Customization potential: Users like yourself (in a previous role) might choose to build their own abstractions, but having a reference implementation helps guide those efforts.

    Future extensibility: We're exploring options to make more of the system open over time. Keeping sensorhub open now lowers the barrier to eventual integration when that happens.

    Keeping it open-source forces us to keep it universal - allows for fast integration in new projects (no matter open or closed source) if sensors will be necessary (it is a forseen future)

That said, your point about cognitive overhead is well taken. Perhaps we can improve the experience by adding clear documentation or diagrams showing how the pieces fit together.

niewim19:
@nalajcie @agkaminski 
I have proposed a novel approach to sensor selection as you proposed: with static libraries per sensor.

This has some drawbacks, which mainly come from the fact that we have no dependency tree. Thus dependencies of a dependency are not structured and have to be kept in a single layer list, and stored for the final `sensors` binary build. This is applicable to sensors from `phoenix-rtos-devices`.

The other factor is that no matter how I specify the **general sensors list** (and their dependencies if applicable) and **project specific sensor list**, they are somewhat not the same: one lists element from current repo, and another lists dependencies from external repositories. This forces us to redefine the `DEVICE_SENSORS` variable for the rebuilding of `phoenix-rtos-devices` after project sensors are build. So single definition of  `DEVICE_SENSORS` for `phoenix-rtos-devices` with external `phoenix-<project_name>-devices` is (probably?) not possible with low/moderate code effort.. During the first building of the `sensors` it is not possible to distinguish between them.

Lastly - for `build.project` file the following structure is planned:
```
export DEVICE_SENSORS="mpu6000 lis2mdl"
b_build_project () {
	b_log "Building phoenix-<project_name>-devices"
	make -C phoenix-<project_name>-devices all install

	b_log "Rebuilding sensors (with <project_name> sensors)"
	make -C phoenix-rtos-devices sensors install DEVICE_SENSORS+="project_specific_sensor-name"
}
```

sensors/Makefile
nalajcie: @dsabala thanks for the input - in the scope of the current PR we won't take any harsh decisions and the refactoring of sensor components is much overdue regardless of the final place the code should be developed further.

The original idea for `SensorHub` was to:
 - provide a generic, robust event-based sensor API (as opposed to common polling-based approach) - think about `uORB` from PX4 (albeit not so generic) or `evdev` for HID devices from linux kernel - not necessarily limited to the sensors commonly used in UAVs
 - standardize the sensor APIs
 - (due to the above - provide alternative ways for re-using the same sensors in simpler environment eg. requiring only polling API)

IMHO having the above as an OSS component would be beneficial for multitude of projects.

I agree that the current state of `SensorHub` digress from the original idea - mostly because of being driven by a single, specific use case. Maybe instead of *hiding the ugly/convoluted code* we should *refactor/rewrite it* into more generic/better designed interface?

dsabala: Ok thanks for the explanation, now I understand the approach we have here. I agree that this module deserves a refactor someday. I hope we can come up with a sensible solution tailored to the needs of the Phoenix RTOS project.

sensors/Makefile
```diff
@@ -4,26 +4,49 @@
 # Copyright 2022 Phoenix Systems
 #
 
+LOCAL_PATH := $(call my-dir)
+
 # Sensors Server API Library
 NAME := libsensors
 include $(static-lib.mk)
 
+# FIXME: should be stored and compiled outside
```
nalajcie, line 13: whitespace at the end

not sure it should be stored / compiled outside -  should provide different implementations with common API

sensors/Makefile
```diff
@@ -4,26 +4,49 @@
...
-# Sensors Server
-SENSORS_ALL := imu/lsm9dsxx imu/mpu6000 mag/lsm9dsxx mag/lis2mdl baro/lps25xx baro/ms5611 gps/pa6h gps/ubx
-SENSORS_SIM := imu/imu_sim.c mag/mag_sim.c baro/baro_sim.c gps/gps_sim.c gps/gps_sim.c
-SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(SENSORS_ALL), $(DEVICE_SENSORS)))
+
+# Simulated sesnsors common library
+NAME := libsimsensors_common
+SRCS := $(wildcard $(LOCAL_PATH)simsensor_common/*.c)
+LOCAL_HEADERS_DIR := nothing  # do not reinstall libsensors/*.h
+DEP_LIBS := libsensors
```
nalajcie, line 27: `DEP_LIBS` for `static-lib.mk` is invalid and does nothing, there was a typo in safety guard (https://github.com/phoenix-rtos/phoenix-rtos-build/pull/229), but still - you should have found out that it doesn't work :P

This error is common across the code.

sensors/Makefile
```diff
@@ -4,26 +4,49 @@
...
 DEPS := libzynq7000-gpio-msg libspi-msg
 include $(static-lib.mk)
 
-# Sensors Server
-SENSORS_ALL := imu/lsm9dsxx imu/mpu6000 mag/lsm9dsxx mag/lis2mdl baro/lps25xx baro/ms5611 gps/pa6h gps/ubx
-SENSORS_SIM := imu/imu_sim.c mag/mag_sim.c baro/baro_sim.c gps/gps_sim.c gps/gps_sim.c
-SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(SENSORS_ALL), $(DEVICE_SENSORS)))
+
+# Simulated sesnsors common library
+NAME := libsimsensors_common
```
nalajcie, line 24: hmm, mayybe all of the common parts (simsensor common, gps common, etc.) should just be a part of `libsensors` - unused functions just will be removed while linking. Right now libsensors is a `dummy` lib providing only header files - so probably shouldn't be a library at all?

sensors/Makefile
```diff
@@ -4,26 +4,49 @@
...
+# FUNCTION WRAPPER FOR DEPENDENCIES INJECTION
+# Should be used in sensors part-Makefiles instead for sensor libraries "DEP_LIBS" setting
+# Adds dependencies for both local sensor static library building and for "sensors" binary building
+define SENSORS_ADD_DEPS
+DEP_LIBS := $(1)
+SENSORS_DEPS := $(sort $(SENSORS_DEPS) $(1))
+endef
+
+# spi dependencies bundle for spi driven sensors
+SENSORS_SPI_DEPS := libsensors-spi libzynq7000-gpio-msg libspi-msg
```
nalajcie, line 40: - I would propose not to include any automatic library dependency resolution - this should be done in a global way in whole build system
- using a macro is not a good solution due to external components (if external component would depend on internal lib - we can't use the macro and still need to state the dependency explicitly)
- instead - always link against common / transport (eg. `libsensors-spi`) libs - they would be removed if no driver needing them is included.


I would propose:
```make
SENSORHUB_DRIVERS_ALL :=

ALL_SENSOR_MAKES := $(shell find . -mindepth 2 -name sensors.mk -not -path '*/.*')
include $(ALL_SENSOR_MAKES)

SENSORHUB_DRIVERS_ENABLED ?= default sensor list

$(info available sensors: $(SENSORHUB_DRIVERS_ALL))
$(info enabled sensors: $(SENSORHUB_DRIVERS_ENABLED))

SENSORHUB_DRIVERS_INTERNAL := $(filter $(SENSORHUB_DRIVERS_ENABLED), $(SENSORHUB_DRIVERS_ALL))
SENSORHUB_DRIVERS_EXTERNAL := $(filter-out $(SENSORHUB_DRIVERS_INTERNAL), $(SENSORHUB_DRIVERS_ENABLED))

$(info internal drivers: $(SENSORHUB_DRIVERS_INTERNAL))
$(info external drivers: $(SENSORHUB_DRIVERS_EXTERNAL))

NAME := sensors
LOCAL_HEADERS_DIR := nothing # do not reinstall libsensors/*.h
SRCS := $(LOCAL_PATH)sensors.c
DEP_LIBS := libsensors libsensors-spi libsimsensors_common  # FIXME: move *common* to libsensors?
DEP_LIBS += $(SENSORHUB_DRIVERS_INTERNAL)
LIBS := $(SENSORHUB_DRIVERS_EXTERNAL)
include $(binary.mk)
```


the `$(info XXX)` part is just to show what is happening, but I'm not completely opposed to printing some info in the final solution

nalajcie (dismissed):
I've put my ideas into inline comments. As we don't have dependency resolution for now - creating dedicated `sensors.mk` to force makefile parsing order is IMHO a good solution.

Please note that some files still have invalid includes:
image

niewim19:
Closed by accident by pushing empty commit tree. Repoened

sensors/Makefile
niewim19: - Applied very similar to proposed structure.
 - Removed macros.
 - Reduced logging to only print out what sensors are compiled.
 - merged common sensor code into `libsensors` with clear include structure.
 - default sensors compiled are only the simulated ones

sensors/Makefile
niewim19: Merged common sensor code into `libsensors` with clear include structure.

sensors/Makefile
niewim19: Well... yeah that`s on my, my mistake. Corrected and tested for correct building of all sensors

sensors/Makefile
niewim19: Now compiled into `libsensors`.

niewim19:
Just putting this PR back onto your radar, @nalajcie. Would appreciate your feedback when you have a moment. Thanks!

nalajcie (commented):
I much more like the current version. 

Added some comments which might be used to polish the solution a little bit, some of them might be fixed later (but I insist on adding relevant `FIXME`s in this PR before the merge).

sensors/Makefile
nalajcie: IMHO we shouldn't merge target-dependant component into the common code, it should be solved by our buildsystem

sensors/Makefile
```diff
@@ -4,26 +4,29 @@
...
-SENSORS_SIM := imu/imu_sim.c mag/mag_sim.c baro/baro_sim.c gps/gps_sim.c gps/gps_sim.c
-SENSORS_LOCAL := $(SENSORS_SIM) $(addsuffix .c, $(filter $(SENSORS_ALL), $(DEVICE_SENSORS)))
+# Sensors application
+SENSORHUB_COMMON_DEPS := libzynq7000-gpio-msg libspi-msg libsensors
+SENSORHUB_DRIVERS_ALL :=
+
+ALL_SENSOR_MAKES := $(shell find . -mindepth 2 -name sensors.mk -not -path '*/.*')
+include $(ALL_SENSOR_MAKES)
+
+SENSORHUB_DRIVERS_ENABLED ?= baro_sim imu_sim mag_sim gps_sim
```
nalajcie, line 22: hmm, maybe add comment that this is the default sensors list and the intent is to change it externally on a per-project basis?

sensors/Makefile
```diff
@@ -4,26 +4,29 @@
...
+SENSORHUB_DRIVERS_INTERNAL := $(filter $(SENSORHUB_DRIVERS_ENABLED), $(SENSORHUB_DRIVERS_ALL))
+SENSORHUB_DRIVERS_EXTERNAL := $(filter-out $(SENSORHUB_DRIVERS_INTERNAL), $(SENSORHUB_DRIVERS_ENABLED))
+$(info sensors enabled (internal [external]): $(SENSORHUB_DRIVERS_INTERNAL) [$(SENSORHUB_DRIVERS_EXTERNAL)])
 
 NAME := sensors
-LOCAL_HEADERS_DIR := nothing
-LOCAL_SRCS += sensors.c $(SENSORS_LOCAL) $(SENSORS_SIM)
-LOCAL_SRCS += gps/nmea.c gps/common.c simsensor_common/event_queue.c simsensor_common/simsensor_reader.c simsensor_common/simsensor_generic.c
-DEP_LIBS := libsensors libsensors-spi libzynq7000-gpio-msg libspi-msg
+LOCAL_HEADERS_DIR := nothing # do not reinstall libsensors/*.h
```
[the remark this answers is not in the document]
niewim19: The name-changing way was chosen. It allows for a lot of `LOCAL_HEADERS_DIR := nothing` code removal. The comment above was changed to highlight that both functionalities are in it (server and client side)

sensors/Makefile
niewim19: Added!

PR #550, opened 2025-04-17
mcxn94x-multi: Add cpu (mailbox) driver

Darchiv:
JIRA: RTOS-1044

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: mcxn947-frdm.

## Checklist:

## Special treatment

      https://github.com/phoenix-rtos/phoenix-rtos-kernel/pull/648

### DIFF
--- a/multi/mcxn94x-multi/cpu.c
+++ b/multi/mcxn94x-multi/cpu.c
@@ -1,0 +1,260 @@
+/*
+ * Phoenix-RTOS
+ *
+ * MCX N94x M33 CPU core driver
+ *
+ * Copyright 2025 Phoenix Systems
+ * Author: Daniel Sawka
+ *
+ * %LICENSE%
+ */
+
+#include <errno.h>
+#include <fcntl.h>
+#include <poll.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/file.h>
+#include <sys/interrupt.h>
+#include <sys/minmax.h>
+#include <sys/msg.h>
+#include <sys/platform.h>
+#include <sys/threads.h>
+
+#include <libtty-lf-fifo.h>
+
+#include "common.h"
+#include "dev.h"
+
+#define REG_IRQ_OFFSET 4
+/* clang-format off */
+enum {
+	reg_irq = 0, reg_irq_set, reg_irq_clr, mutex = 62
+};
+/* clang-format on */
+
+#define REG_IRQ(CPUID, REG) (*(common.mailbox + (REG_IRQ_OFFSET * (CPUID)) + (REG)))
+
+
+static struct {
+	uint8_t rxBuf[64];
+	lf_fifo_t rxFifo;
+	handle_t rxLock;
+	handle_t txLock;
+	handle_t irqLock;
+	handle_t rxCond;
+
+	volatile uint32_t *mailbox;
+	unsigned int major;
+	uint8_t coreidSelf;
+	uint8_t coreidRemote;
+} common;
+
+
+static int mailbox_irqHandler(unsigned int n, void *arg)
+{
+	size_t consumed = 0;
+	uint32_t value = REG_IRQ(common.coreidSelf, reg_irq);
+
+	if (value != 0) {
+		REG_IRQ(common.coreidSelf, reg_irq_clr) = value;
+
+		size_t rxLen = (value >> 24) & 0x3;
+
+		for (; consumed < rxLen; consumed++) {
+			uint8_t byte = (value >> (8 * consumed)) & 0xff;
+
+			/* TODO: Data is overwritten if rx fifo is full. Consider disabling
+			the interrupt somehow and leaving the value in the register until
+			the application reads from rx fifo. Also, fifo needs to be more
+			generic and at least have function to get its current length. */
+			lf_fifo_push(&common.rxFifo, byte);
+		}
+	}
+
+	return (consumed > 0) ? 1 : -1;
+}
+
+
+static ssize_t fifoRead(unsigned char *buf, size_t size, bool blocking)
+{
+	size_t bytesRead = 0;
+
+	if (buf == NULL || size == 0) {
+		return -EINVAL;
+	}
+
+	/* Synchronize multiple readers */
+	mutexLock(common.rxLock);
+
+	while (bytesRead < size) {
+		mutexLock(common.irqLock);
+		while (lf_fifo_empty(&common.rxFifo) != 0) {
+			if (!blocking) {
+				mutexUnlock(common.irqLock);
+				mutexUnlock(common.rxLock);
+				return (bytesRead > 0) ? (ssize_t)bytesRead : -EAGAIN;
+			}
+			condWait(common.rxCond, common.irqLock, 0);
+		}
+		mutexUnlock(common.irqLock);
+
+		while (bytesRead < size) {
+			if (lf_fifo_pop(&common.rxFifo, &buf[bytesRead]) != 0) {
+				bytesRead += 1;
+			}
+			else {
+				break;
+			}
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+	mutexLock(common.tx_lock);
+
+	while (bytes_written < len) {
+		if (blocking == 0 && REG_IRQ(common.coreid_remote, reg_irq) != 0u) {
+			mutexUnlock(common.tx_lock);
+			return (bytes_written > 0) ? bytes_written : -EAGAIN;
+		}
+
+		while (blocking != 0 && REG_IRQ(common.coreid_remote, reg_irq) != 0u) {
+			; /* Wait until the other CPU reads IRQx register */
```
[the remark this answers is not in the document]
Darchiv: The remote core is expected to fetch the Mailbox.IRQ register in a fast manner (using interrupts), so this is not really an issue. Maybe it will be worse when FIFOs used will support non-overwriting mode and thus produce backpressure upon the sender, because the receiver will not clear the register until there is space in RX FIFO. Handling such case will require a more sophisticated protocol over Mailbox, with acknowledgements etc. as there is no interrupt line for indicating that the register is empty.

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+
+			/* TODO: Data is overwritten if rx fifo is full. Consider disabling
+			the interrupt somehow and leaving the value in the register until
+			the application reads from rx fifo. Also, fifo needs to be more
+			generic and at least have function to get its current length. */
+			lf_fifo_push(&common.rx_fifo, byte);
+		}
+	}
+
+	return (consumed > 0) ? 1 : -1;
```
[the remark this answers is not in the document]
Darchiv: The return value is used by the kernel to decide whether or not to signal the conditional variable associated with this interrupt handling routine (as declared by `interrupt()` syscall).

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+
+	if (dev_allocMajor(&common.major) < 0) {
+		/* TODO - exit the driver? trigger the reset? report the error? */
+		return;
+	}
+
+	dev_register(cpu_handleMsg, common.major);
+
+	unsigned int minor = 0;
+	char fname[] = "cpuM33";
```
agkaminski, line 270: Can be simply passes as an argument to the `dev_registerFile`, I guess you want to be super explicit with function usage, consider below to avoid copying string to the stack
```suggestion
	static const char *fname = "cpuM33";
```

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+	common.mailbox = (void *)0x400b2000;
+
+	lf_fifo_init(&common.rx_fifo, common.rx_buf, sizeof(common.rx_buf));
+
+	mutexCreate(&common.rx_lock);
+	mutexCreate(&common.tx_lock);
+	mutexCreate(&common.irq_lock);
+	condCreate(&common.rx_cond);
+
+	interrupt(mailbox0_irq, mailbox_irqHandler, NULL, common.rx_cond, NULL);
```
agkaminski, line 245: In interrupt we're using `common.coreid_self`. It's dangerous to enable the interrupt before this value is set

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	switch (msg->type) {
+		case mtOpen:
+		case mtClose:
+			msg->o.err = 0;
+			break;
+
+		case mtWrite:
+			blocking = ((msg->i.io.mode & O_NONBLOCK) != 0u) ? 0u : 1u;
```
agkaminski, line 207: Using bool for `blocking` would make code somewhat shorter in places

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+			return (bytes_written > 0) ? bytes_written : -EAGAIN;
+		}
+
+		while (blocking != 0 && REG_IRQ(common.coreid_remote, reg_irq) != 0u) {
+			; /* Wait until the other CPU reads IRQx register */
+		}
+
+		/* Message structure: MSB[ 6b reserved | 2b len(data) | 24b data ]LSB */
+		uint32_t value = 0;
+		to_write = min(len - (size_t)bytes_written, 3);
```
agkaminski, line 156: BTW I would like to suggest consistent use of camelCase ;) 
```suggestion
		size_t to_write = min(len - (size_t)bytes_written, 3);
```

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+	}
+
+	/* Synchronize multiple readers */
+	mutexLock(common.rx_lock);
+
+	if (blocking != 0) {
+		while (bytes_read < size) {
+			mutexLock(common.irq_lock);
+			while (lf_fifo_empty(&common.rx_fifo) != 0) {
+				condWait(common.rx_cond, common.irq_lock, 0);
```
agkaminski, line 96: I don't think we need separate blocks for blocking/non-blocking
```suggestion
				if (!blocking) {
					mutexLock(common.irq_lock);
					return bytes_read ? bytes_read : -EGAIN;
				}
				condWait(common.rx_cond, common.irq_lock, 0);
```

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,275 @@
...
+				else {
+					break;
+				}
+			}
+		}
+	}
+	else {
+		if (lf_fifo_empty(&common.rx_fifo) != 0) {
+			mutexUnlock(common.rx_lock);
+			return -EWOULDBLOCK;
```
agkaminski, line 113: `EWOULDBLOCK` is for sockets, I think

Darchiv: Right, fixed

multi/mcxn94x-multi/cpu.c
Darchiv: Done

multi/mcxn94x-multi/cpu.c
Darchiv: Done

multi/mcxn94x-multi/cpu.c
Darchiv: This was a leftover from version which sets up multiple channels. Passing the literal directly now

multi/mcxn94x-multi/cpu.c
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	/* Synchronize multiple readers */
+	mutexLock(common.rxLock);
+
+	while (bytesRead < size) {
+		mutexLock(common.irqLock);
+		while (lf_fifo_empty(&common.rxFifo) != 0) {
+			if (!blocking) {
+				mutexUnlock(common.irqLock);
```
agkaminski, line 97: Sorry for a buggy suggestion 
```suggestion
				mutexUnlock(common.irqLock);
				mutexUnlock(common.rxLock);
```

multi/mcxn94x-multi/cpu.c
Darchiv: Done, camelCased everything

multi/mcxn94x-multi/cpu.c
Darchiv: I wanted to avoid locking `irqLock` unnecessarily in non-blocking call. Your suggestion makes code cleaner, to opted for it

multi/mcxn94x-multi/cpu.c
Darchiv: Fixed

PR #551, opened 2025-04-30
zynqmp: add i2c driver

dsabala:
Generalized zynq7000 i2c driver to use it in zynqmp
JIRA: PP-289

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: Trenz TEBF0808 carrier board

## Checklist:

## Special treatment

### DIFF
--- a/i2c/zynq/libzynq-i2c.c
+++ b/i2c/zynq/libzynq-i2c.c
@@ -1,7 +1,7 @@
 /*
  * Phoenix-RTOS
  *
- * Zynq-7000 I2C driver
+ * Zynq7000 / ZynqMP I2C driver
  *
  * Copyright 2022 Phoenix Systems
  * Author: Hubert Buczynski
@@ -23,7 +23,14 @@
 #include <errno.h>
 #include <i2c.h>
 
+#if defined(__CPU_ZYNQ7000)
 #include <phoenix/arch/armv7a/zynq7000/zynq7000.h>
+#elif defined(__CPU_ZYNQMP)
+#include <phoenix/arch/aarch64/zynqmp/zynqmp.h>
+#else
+#error "Unsupported platform"
+#endif
+
 #include <board_config.h>
 
 /* I2C registers */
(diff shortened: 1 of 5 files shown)


i2c/zynq/libzynq-i2c.c
```diff
@@ -43,58 +50,89 @@
...
 		int pin;          /* MIO pin */
 		unsigned int cfg; /* MIO pin configuration */
 	} pins[2];            /* I2C pins, CLK, SDA */
 } i2c_info_t;
 
+/* clang-format off */
 
-/* device configuration based on board_config */
+/* Device configuration based on board_config */
+#if defined(__CPU_ZYNQ7000)
```
agkaminski, line 76: Now we have a common structure, why not capitalize on this? How about hiding differences under e.g. defines, like in the `MIO_SDA` case?

i2c/zynq/libzynq-i2c.c
```diff
@@ -451,6 +471,91 @@ static int i2c_reset(void)
...
+	reg |= (3 << 14) | (49 << 8);
+
+	/* clear FIFO, - 7bit addressing */
+	reg |= (1 << 6) | (1 << 2);
+
+	*(i2c.base + I2C_CR) = reg;
+
+	/* initialize timeout; wait 255 SCL cycles when the SCL is held Low, before generating a timeout interrupt. */
+	*(i2c.base + I2C_TIMEOUT) = 255;
+}
```
agkaminski, line 490: Isn't it exactly the same as on zynq7000?

i2c/zynq/libzynq-i2c.c
```diff
@@ -427,12 +445,14 @@ static int i2c_initClk(void)
...
+	const i2c_info_t *info = &devsInfo[i2c.devID];
+
 	platformctl_t pctl;
 	int err;
 
 	pctl.action = pctl_get;
 	pctl.type = pctl_devreset;
-	pctl.devreset.dev = pctl_ctrl_i2c_rst;
+	pctl.devreset.dev = info->rst;
 	err = platformctl(&pctl);
```
agkaminski, line 456: Pls fix while editing this
```suggestion
	return platformctl(&pctl);
```

i2c/zynq/zynq-i2c.c
```diff
@@ -127,7 +127,7 @@ int main(int argc, char **argv)
 	}
 
 	printf("i2c: initialized\n");
-	thread((void *)port);
+	thread((void *)&port);
```
agkaminski, line 130: This is a bad pattern in case of beginthread instead of direct function call. I suggest keeping the old style in case some day we decide to spawn a thread instead

agkaminski:
Sorry for the late review! It slipped through the cracks

i2c/zynq/libzynq-i2c.c
dsabala: Maybe we can go the other way and remove MIO_SDA to reduce code complexity (its one "abstraction" layer less for an engineer who read code).

i2c/zynq/libzynq-i2c.c
dsabala: fixed

i2c/zynq/libzynq-i2c.c
dsabala: Prescaler value is different. I put entire i2c_initCtrl function into single ifdef block, however we could use ifdef block limited to those two lines
```
/* (20 000 000 Hz / (2.2 * (1 + 1)) * (49 + 1)) = 90,9 kHz */
reg |= (3 << 14) | (49 << 8);
```

i2c/zynq/zynq-i2c.c
dsabala: I did it to lint `warning: cast from pointer to integer of different size`

agkaminski: You can cast through `intptr_t` to solve this instead
```
int port = (int)(intptr_t)arg;
```

i2c/zynq/libzynq-i2c.c
agkaminski: Pls check, it is as it was

dsabala: I guess the comment was attached to the wrong line
obraz

i2c/zynq/zynq-i2c.c
dsabala: Maybe we should give up on prematurely preparing the "thread" function for its use as an argument to the "beginthread" function and just use strict, fitting datatype?

agkaminski: Well, I is a valid consideration. In this exact case I can see a value and reason to extend the driver with multiple threads - there is more than one I2C instance handled here and these can be handled in parallel, independently. So IMHO it might be beneficial to consider adding a thread (to be able to handle each and every I2C instance at the same time at least) instead of removing these preparations (not in this PR ofc)

agkaminski:
@niewim19  General remark - the approach used to share the driver between zynq7000 and zynqmp will suffice for now (as there are some more important priorities), but we should examine more pedantic approach in the future and implement it as soon, as possible. This applies to other drivers of zynq* as well.

What I mean is to remove inline `ifdefs`, that are in various places and to prepare a translation layer (pretty much HAL) to hide the differences between platforms. This "HAL" might be shared across various drivers for this family too.

Other consideration are parameters - pins and clocks configuration. It would be ideal to use something like DTB for these, but we might at least move them to the `board_config.h` for the time being, while DTB support is not available. E.g. clock prescaler should be calculated based on the input and target clock, either dynamically (better, as DTB parsing is done in runtime) or statically.

This PR does not regress the code regarding above point, so threat this as a general remark. This one is not a merge blocker in any way.

i2c/zynq/zynq-i2c.c
dsabala: should I use the "intptr_t" solution?

agkaminski: Yes, if it in fact works and does not give warning 😄

niewim19:
> @niewim19 General remark - the approach used to share the driver between zynq7000 and zynqmp will suffice for now (as there are some more important priorities), but we should examine more pedantic approach in the future and implement it as soon, as possible. This applies to other drivers of zynq* as well.
> 
> What I mean is to remove inline `ifdefs`, that are in various places and to prepare a translation layer (pretty much HAL) to hide the differences between platforms. This "HAL" might be shared across various drivers for this family too.
> 
> Other consideration are parameters - pins and clocks configuration. It would be ideal to use something like DTB for these, but we might at least move them to the `board_config.h` for the time being, while DTB support is not available. E.g. clock prescaler should be calculated based on the input and target clock, either dynamically (better, as DTB parsing is done in runtime) or statically.
> 
> This PR does not regress the code regarding above point, so treat this as a general remark. This one is not a merge blocker in any way.

Yes you`re right. Dedicated `hal`s will be implemented after PCI driver.

PR #567, opened 2025-07-15
tty/uart16550: add TDA4VM target

jmaksymowicz:
## Description
Added hardware abstraction files for TDA4VM target.

To make the `uart16550` driver compatible with the UART on TDA4VM some changes had to be made concerning interrupt handling. On that platform the UART interrupt is level-triggered and cannot be reconfigured to be edge-triggered. Additionally, trying to mask the data ready interrupt in the `IMR` register has no effect after the interrupt has been asserted. To solve this, the hardware FIFO is read into a buffer until the data ready interrupt flag is down.

Code has been tested on `ia32-generic-qemu` and `riscv64-generic-qemu` and found to work. It also achieves better performance because it reads FIFO fully before re-enabling the interrupt (thus avoiding interrupt handling overhead).

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv7r5f-tda4vm-som, ia32-generic-qemu, riscv64-generic-qemu

## Checklist:

## Special treatment

### DIFF
--- a/tty/uart16550/uart16550.c
+++ b/tty/uart16550/uart16550.c
@@ -13,6 +13,7 @@
  */
 
 #include <errno.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -36,8 +37,12 @@
 
 #define KMSG_CTRL_ID 100
 
+#define SW_BUF_SIZE 64
+
 typedef struct {
 	uint8_t hwctx[64];
+	uint8_t buf[SW_BUF_SIZE];
+	volatile unsigned int buf_i;
 
 	unsigned int init;
 	unsigned int clk;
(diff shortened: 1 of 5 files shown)


tty/uart16550/uarthw-tda4vm.c
```diff
@@ -0,0 +1,284 @@
...
+{
+	int muxSetting, isTx;
+	platformctl_t pctl;
+	pctl.action = pctl_set;
+	pctl.type = pctl_pinconfig;
+
+	for (unsigned i = 0; i < MAX_PINS_PER_UART; i++) {
+		if (info->pins[i].pin < 0) {
+			return -EINVAL;
+		}
```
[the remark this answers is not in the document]
jmaksymowicz: The sentinel value is only necessary if not all elements in the array is used. No out-of-bounds reads are possible.

tty/uart16550/uarthw-tda4vm.c
```diff
@@ -0,0 +1,284 @@
...
+
+		if (info->pins[i].pin == pin) {
+			muxSetting = info->pins[i].muxSetting;
+			isTx = info->pins[i].isTx;
+			pctl.pin_config.pin_num = pin;
+			pctl.pin_config.debounce_idx = 0;
+			pctl.pin_config.mux = muxSetting & 0xff;
```
Darchiv, line 189: `info->pins[i].muxSetting` is defined as `uint8_t` - why store it in a temporary `int` variable if it can be assigned directly to `pctl.pin_config.mux` without a mask?

tty/uart16550/uarthw-tda4vm.c
```diff
@@ -0,0 +1,284 @@
...
+};
+
+
+typedef struct {
+	volatile uint32_t *base;
+	unsigned int irq;
+} uarthw_ctx_t;
+
+
+uint8_t uarthw_read(void *hwctx, unsigned int reg)
```
Darchiv, line 143: `uarthw_ctx_t` could have been forward-declared in uarthw.h to avoid silly casts from `void*`. For future reference, out of scope of this PR

tty/uart16550/uarthw-tda4vm.c
```diff
@@ -0,0 +1,284 @@
...
+			{ pin_wkup_gpio0_12, 0, 1 },
+			{ pin_mcu_ospi1_d1, 4, 0 },
+			{ pin_wkup_gpio0_11, 2, 0 },
+			{ pin_wkup_gpio0_13, 0, 0 },
+		},
+		.pins_selected = {
+			.tx = UART0_TX,
+			.rx = UART0_RX,
+		},
+	},
```
Darchiv, line 133: Purposefully only config for UART0 is provided?

tty/uart16550/uart16550.c
```diff
@@ -129,46 +134,99 @@ static void set_cflag(void *_uart, tcflag_t *cflag)
...
-		while ((iir = uarthw_read(uart->hwctx, REG_IIR)) & IIR_IRQPEND) {
-			condWait(uart->intcond, uart->mutex, 0);
+		uarthw_write(uart->hwctx, REG_IMR, target_imr);
+		condWait(uart->intcond, uart->mutex, 0);
+		/* For the following part the interrupt needs to be masked */
+		uarthw_write(uart->hwctx, REG_IMR, 0);
+
+		/* Empty received buffer */
+		unsigned int buf_i = uart->buf_i;
+		for (unsigned int i = 0; i < buf_i; i++) {
```
Darchiv, line 206: I assume that the compiler will optimize-out this loop on RISC-V, as `uart->buf_i` is always 0, so it is not worth to `#ifdef` this loop?

tty/uart16550/uarthw-tda4vm.c
jmaksymowicz: Yes, originally I expected we could use other UARTs on this platform but now we are required to only use peripherals in the MCU domain.

PR #590, opened 2025-09-25
libsensors: operands in sensors bus usage

niewim19:
## Description

 - adds operands for generic bus (spi, i2c) control in sensors.h
   - implementation-defined access to buses (spi or i2c) should be defined by application,
   - access function should provided into sensor_bus_t structure before `alloc` function being called.
   - application must define its own structure of spi/i2c context
 - adds local sensor reads for all but `ms5611` sensor
   - `ms5611` will be addressed later, as it is not poll-structured (at least two calls needed for one read)
 - simplifies localsensor implementation
   - from now on `locsensors` module only auto-registers compiled sensor libraries and allows driver retrieval
   - change was necessary as to not artificially split sensor allocation into init (with `sensor_bus_t` setup) and alloc (where sensors are allocated)
 - removal of zynq7000 dependencies

## Motivation and Context

There are multiple commits in this PR as well-defined changes were implemented. They are going to be squashed into one.

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: zynq7000.

## Checklist:

## Special treatment


sensors/server-include/libsensors/sensor.h
```diff
@@ -18,18 +18,68 @@
 
 #include <sys/rb.h>
 #include <posix/idtree.h>
+#include <spi-msg.h>
 
 
 #define THREAD_PRIORITY_MSGSRV 3 /* priority of messaging server of sensorhub */
 #define THREAD_PRIORITY_SENSOR 3 /* priority of sensor driver threads */
 
+enum sensor_bus_type { bus_none = 0,
```
agkaminski, line 27: What's use case of `bus_none`?

sensors/server-include/libsensors/sensor.h
```diff
@@ -18,18 +18,68 @@
...
+		 * For I2C these are the name, or virtual path to i2c device (e.g "/dev/i2c1"), and subdev is unused.
+		 */
+		int (*bus_open)(struct _sensor_bus_t *bus, const char *dev, const char *subdev);
+	} ops;
+	union {
+		struct {
+			spimsg_ctx_t ctx; /* spi context, not to be used by device */
+			oid_t ss;         /* spi ss oid, not to be used by device */
+		} spi;
+	} spec;
```
agkaminski, line 72: We can hide details of the bus context by using a pointer and forward declaration instead. Then thanks to that each bus driver might use its specific context implementation.

```
struct bus_ctx;
(...)
struct bus_ctx *ctx;
```

or (former preferred imho)

```
struct bus_ctx;
typedef struct bus_ctx bus_ctx_t;
(...)
bus_ctx_t *ctx;
```

Implementation is then provided by the lib user and bus provider.

sensors/imu/mpu6000.c
```diff
@@ -141,66 +138,99 @@ static int mpu6000_whoamiCheck(mpu6000_ctx_t *ctx)
...
+	if (bus->bustype != bus_spi) {
+		return -ENODEV;
+	}
+
+	/* initialize SPI device communication */
+	if (bus->ops.bus_open(bus, spiDev, ss) < 0) {
+		return -1;
+	}
+
+	if (bus->ops.bus_cfg == NULL) {
```
agkaminski, line 208: `NULL` check is redundant

niewim19: removed

sensors/server-include/libsensors/sensor.h
niewim19: Implemented in the first form proposed

sensors/server-include/libsensors/sensor.h
```diff
@@ -15,9 +15,11 @@
 #define _LIBSENSORS_SENSOR_H_
 
 #include <libsensors/client.h>
+#include <libsensors/bus.h>
 
 #include <sys/rb.h>
 #include <posix/idtree.h>
+#include <spi-msg.h>
```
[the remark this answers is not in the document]
niewim19: Removed

PR #612, opened 2025-12-17
imx6ull-uart: enable hardware flow control

jmaksymowicz:
## Description
The UART driver already had a configuration option `use_rts_cts`, but it did nothing except configure MUX-es. This PR fully implements hardware flow control that prevents overruns in all situations. To achieve this, FIFO implementation from phoenix-rtos-corelibs was used, as it offers the ability to push without possibility of overwrites.

In half-duplex mode the CTS pin is used instead to switch between transmit and receive mode. CTS pin high indicates that the UART is transmitting and CTS low indicates it is receiving - this signalling is compatible with typical RS-485 transceivers, which this change is intended for.

TODO: for now the half-duplex mode switches the `UCR2_RXEN` bit regardless of whether the UART is opened for reading - this would need to be fixed. Also, an `ioctl` would need to be added to switch modes between no flow control / hardware flow control / half-duplex mode at runtime.

## Motivation and Context
Allow hardware flow control for full RS-232 support and half-duplex mode for RS-485 support.
JIRA: SEN-40

## Types of changes

## How Has This Been Tested?

## Checklist:

## Special treatment

### DIFF
--- a/tty/imx6ull-uart/imx6ull-uart.c
+++ b/tty/imx6ull-uart/imx6ull-uart.c
@@ -30,8 +30,8 @@
 #include <fcntl.h>
 #include <syslog.h>
 
+#include <lf-fifo.h>
 #include <libtty.h>
-#include <libtty-lf-fifo.h>
 #include <libklog.h>
 
 #include <phoenix/arch/armv7a/imx6ull/imx6ull.h>
@@ -65,8 +65,9 @@
 #define USR1_TRDY      (1 << 13)
 #define USR1_FRAMERR   (1 << 10)
 #define USR1_PARITYERR (1 << 15)
-#define USR2_RDR       (1 << 0)
-#define USR2_ORE       (1 << 1)
+#define USR2_RDR       (1 << 0) /* Receive data ready */
+#define USR2_ORE       (1 << 1) /* RX overrun */
+#define USR2_TXDC      (1 << 3) /* Transmitter complete */
 #define UTS_TXFULL     (1 << 4)
 #define UTS_SOFTRST    (1 << 0)
 #define UCR1_UARTEN    (1 << 0)
@@ -76,11 +77,18 @@
 #define UCR2_RXEN      (1 << 1)
 #define UCR2_TXEN      (1 << 2)
 #define UCR2_WS        (1 << 5)
+#define UCR2_CTS       (1 << 12) /* CTS pin output (inverted) */
+#define UCR2_CTSC      (1 << 13)
 #define UCR2_IRTS      (1 << 14)
 #define UCR3_RXDMUXSEL (1 << 2)
 #define UCR3_RI        (1 << 8)
 #define UCR3_DCD       (1 << 9)
 #define UCR3_DSR       (1 << 10)
+#define UCR4_TCEN      (1 << 3) /* Transmitter complete interrupt enable */
+
+#define RTS_CTS_NONE       0
+#define RTS_CTS_HWFLOW     1
+#define RTS_CTS_HALFDUPLEX 2
 
 #define XCAT2(a, b) a##b
 #define PCTL(x)     XCAT2(pctl_, x)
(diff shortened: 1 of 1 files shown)


tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -361,12 +376,27 @@ static int uart_intr(unsigned int intr, void *data)
 	if ((*(uart.base + ucr1) & UCR1_RRDYEN) != 0) {
 		while ((*(uart.base + usr2) & USR2_RDR) != 0) {
 			/* NOTE: lock-free push */
-			if (lf_fifo_push(&uart.rx_sw_fifo, (*(uart.base + urxd)) & 0xff) == 0) {
-				uart.counters.sw_overrun++;
+			if (uart.rts_cts_mode == RTS_CTS_HWFLOW) {
+				if (lf_fifo_full(&uart.rx_sw_fifo)) {
+					*(uart.base + ucr1) &= ~UCR1_RRDYEN;
```
ziemleszcz, line 403: Missing `break`? Infinite while loop when LF FIFO is full (byte is not read from HW FIFO).

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -614,12 +707,15 @@ char __attribute__((aligned(8))) stack0[2048];
...
+	printf("Usage: %s [mode device speed parity rts_cts_mode [-t] [-e] [-s]]\n", progname);
 	printf("\tmode: 0 - raw, 1 - cooked (default cooked)\n");
 	printf("\tdevice: 1 to 8 (default 1)\n");
 	printf("\tspeed: baud_rate (default 115200)\n");
 	printf("\tparity: 0 - none, 1 - odd, 2 - even (default none)\n");
-	printf("\tuse_rts_cts: 0 - no hardware flow control, 1 - use hardware flow control (default no hardware flow control)\n");
+	printf("\trts_cts_mode (default 0):\n"
+		   "\t\t0 - RTS/CTS not used\n"
+		   "\t\t1 - RTS/CTS used for hardware flow control\n"
+		   "\t\t2 - CTS used for RS485 nDE signal, RTS unused\n");
```
ziemleszcz: Why nDE? Driver enable is active high.

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -203,6 +212,12 @@ typedef struct {
 static uart_t uart = { 0 };
 
 
+static inline bool uart_uses_rts_cts(uart_t *uart)
+{
+	return uart->rts_cts_mode != RTS_CTS_NONE;
```
ziemleszcz: In `RTS_CTS_HALFDUPLEX` mode, RTS is unused. It should not be muxed or configured. For example on MIDASH this is a GPIO pin used to control LED.

tty/imx6ull-uart/imx6ull-uart.c
jmaksymowicz: Yes, changed the description to how the pin works.

tty/imx6ull-uart/imx6ull-uart.c
jmaksymowicz: I changed the setup so that RTS pin will not be configured in half-duplex mode.

tty/imx6ull-uart/imx6ull-uart.c
jmaksymowicz: True, fixed the loop now.

jmaksymowicz:
When writing this code I noticed that there's an undocumented hardware "feature" - the CTS pin idles high when either the receiver is disabled or UART is disabled. Because we want to use CTS as the "driver enable" signal for RS485 transceiver, we must keep CTS low whenever we're not transmitting - this means that in half-duplex mode data reception has to be enabled even when the UART device is closed or only open for writing.

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -254,7 +271,17 @@ static int uart_open(int flags)
 		}
 
 		/* start RX */
-		*(uart.base + ucr2) |= UCR2_RXEN;
+		if (uart.rts_cts_mode != RTS_CTS_HALFDUPLEX) {
+			*(uart.base + ucr2) |= UCR2_RXEN;
+		}
+		else {
+			/* In half-duplex mode we need to keep RXEN at 1 the whole time so that CTS pin idles low.
```
ziemleszcz: Use our multi-line comment style.

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -254,7 +271,17 @@ static int uart_open(int flags)
...
+			 * Due to this, FIFO may be full of junk data when opening UART - so here we empty it. */
+			while ((*(uart.base + usr2) & USR2_RDR) != 0) {
```
ziemleszcz, line 282: Also flush SW FIFO.

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -778,6 +918,9 @@ int main(int argc, char **argv)
 	/* enable UART */
 	*(uart.base + ucr1) |= UCR1_UARTEN;
 
+	/* Set up pin muxes */
+	set_mux(&uart);
```
ziemleszcz, line 937: Set pin muxes before UART setup?

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -361,12 +390,28 @@ static int uart_intr(unsigned int intr, void *data)
...
+				if (lf_fifo_full(&uart.rx_sw_fifo)) {
+					*(uart.base + ucr1) &= ~UCR1_RRDYEN;
+					break;
+				}
+				else {
+					lf_fifo_push(&uart.rx_sw_fifo, (*(uart.base + urxd)) & 0xff);
+				}
+			}
+			else {
+				/* TODO: here we should increment uart.counters.sw_overrun if necessary,
```
ziemleszcz: Use our multi-line comment style.

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -603,23 +705,56 @@ static void set_cflag(void *_uart, tcflag_t *cflag)
 }
 
 
+static void uart_mode_setup(uart_t *uart, int rts_cts_mode)
+{
+	uart->rts_cts_mode = rts_cts_mode;
+	uint32_t tmp = *(uart->base + ucr2) & ~(UCR2_IRTS | UCR2_CTSC | UCR2_CTS | UCR2_RXEN | UCR2_TXEN);
+	/* 8-bit transmit, no reset */
+	tmp |= UCR2_WS | UCR2_SRST;
+	/* UNDOCUMENTED BEHAVIOR WARNING: The UART ignores UCR2_CTS bit if either:
```
ziemleszcz: Use our multi-line comment style.

tty/imx6ull-uart/imx6ull-uart.c
jmaksymowicz: No, I intentionally set up muxes after enabling the UART - otherwise in half-duplex mode the CTS pin will output a short "glitch" high state at startup.

ziemleszcz: Might be worth adding a comment explaining the reason, so the order is not reversed in the future.

tty/imx6ull-uart/imx6ull-uart.c
jmaksymowicz: I don't think it's necessary. While UART is closed, `UCR1_RRDYEN` gets set to 0, so we don't get RX ready interrupts - nothing gets put into the SW FIFO.

tty/imx6ull-uart/imx6ull-uart.c
jmaksymowicz: Added comment.

tty/imx6ull-uart/imx6ull-uart.c
ziemleszcz: What about UART reopen? Could stale data remain in the driver SW FIFO or TTY SW FIFO from the previous open? I think we had a similar issue on DATARO.

tty/imx6ull-uart/imx6ull-uart.c
ziemleszcz: The multiline comment should look like this:

/*
 *  LINE 1
 * LINE 2
 */

First and last line has no text (GitHub will probably mess it up).

tty/imx6ull-uart/imx6ull-uart.c
jmaksymowicz: This could be a problem in all modes - between issuing an UART close and clearing RRDYEN there could be extra characters that will then get "spit out" when the UART is opened next time. I could add a SW FIFO and TTY FIFO flush before the UART is opened.

ziemleszcz (approved):

PR #656, opened 2026-04-28
stm32l4-multi/pwm_n6: implement bidirectional DShot communication

jmaksymowicz:
Implement bidirectional DShot communication and optimize the sending and receiving of DShot values.

## Description

In order to implement bidirectional DShot it was necessary to add new peripheral types (timer capture) to the DMA API. To do it efficiently, lookup tables with peripheral setups were reorganized to reduce redundancies. Also, a bug in the DMA driver was fixed that would cause memory buffer length to be incorrectly calculated in certain usage scenarios. Said bug could, at worst, lead to buffer over-runs when writing.

Many optimizations and improvements were made to and around the `pwm_set` function:

1. The number of HW register accesses was reduced - accessing them is very slow, so unnecessary or redundant operations were removed, consolidated or moved to `pwm_configure`.
1. After forcing timer update, waiting was changed from IRQ-driven to polling - the wait is extremely short (~50 ns) so the previous method had massive overhead of > 8 μs.
1. Argument verification can now be skipped using `pwm_setInternal` function.
1. Option to change PWM signal polarity (active-low vs active-high) was added.

Some other improvements were made to the timers, such as changing DMA mutex-ing to be per-timer, replacing some hard-coded values with macros for better readability and caching timer base frequency to speed up timeout calculations.

The procedure for sending DShot was optimized by using the `DMAR` register to "distribute" timer values across `CCRx` registers instead of using multiple DMA channels to perform the transaction. This reduces the required number of DMA channels from 4 to 1, freeing them for other uses. It does reduce flexibility somewhat (e.g. it's not possible to leave a channel at previous value, only set it to idle value), but for our use case this is acceptable.

In the final version the send/receive buffer is stored in cached memory. In testing it turned out that while rendering TX timer counts to uncached memory was marginally faster (about 1-2 μs faster), processing RX timer counts was significantly slower (about 6 μs) - so I made the decision to return to using cached memory.

To switch the timer from send to receive mode, a callback is attached to the DMA transaction. The callback runs in ISR context as soon as the timer starts outputting idle values. In this callback, the timer is reconfigured to start receiving. The reconfiguration takes about 1.6 μs - well within the protocol's time limit of 30 μs. The timer's channels are reconfigured for capturing input signal transition timestamps and the timer itself to wait until data reception should end. The timer's update ISR disables data capture and signals userspace thread to wake up and process the captured data.

Captured data is processed with a simple algorithm - the time difference between two captures is divided by expected bit length to get the number of bits that should be shifted in. Perhaps by examining all lengths bit length could be recovered more precisely - this would be useful if clock drift turns out to be a major issue, but so far no such problems were found. 

The driver outputs received data into an array of `int`s given by the `data` pointer in the message. Values >= 0 indicate successful reception (lower 16 bits contain the data), `-ENODATA` that no decodable data was received and `-EILSEQ` that the received data could not be decoded.

Ultimately, the time for preparing the timer for sending (rendering values, clearing cache, DMA setup) is 16 μs and for receiving (stopping DMA, decoding values from 4 channels, cleanup) is 11 μs. CPU usage by multidriver when transacting 1000x a second is 5.2% (incl. OS message processing overhead).

Suggestions are welcome for improvements to the API.

## Motivation and Context
JIRA: RTOS-1291

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: armv8m55-stm32n6

## Checklist:

## Special treatment

### DIFF
--- a/multi/stm32l4-multi/pwm_n6.c
+++ b/multi/stm32l4-multi/pwm_n6.c
@@ -28,8 +28,8 @@
 #include <stdlib.h>
 #include <time.h>
 
-#define USE_DSHOT_DMA 1
-#define MAX_DSHOT_DMA 32
+#define USE_DSHOT_DMA  1
+#define DMA_ALLOC_SIZE (18 * PWM_CHN_NUM * sizeof(uint16_t))
 
 #define TIM_DIER_UIE   (1 << 0)  /* Update IRQ enable */
 #define TIM_DIER_CC1IE (1 << 1)  /* Capture/compare 1 IRQ enable */
@@ -104,11 +104,15 @@ typedef struct {
 	bool initialized;
 #if USE_DSHOT_DMA
 	handle_t dmaMutex;
+	void *dmaMem;
 	const struct libdma_per *dma_per[PWM_CHN_NUM];
 	const struct libdma_per *dma_multiChannel;
 	const struct libdma_per *dma_capture[PWM_CHN_NUM];
 #endif
 } pwm_tim_data_t;
+
+/* Idle value made available as a static const, so that it can be read by DMA */
+static const uint16_t pwm_idleValue = 0;
 
 static const struct {
 	pwm_tim_setup_t timer[pwm_tim_count];
@@ -238,6 +242,12 @@ static struct {
 	pwm_tim_data_t timer[pwm_tim_count];
 	uint32_t baseFreq; /* Timer base frequency, 32-bit range is acceptable on this platform */
 } pwm_common;
+
+/* Get log2(size) for sizes of 4, 2 or 1. */
+static inline uint8_t sizeLog(size_t size)
+{
+	return (size == 4) ? 2 : ((size == 2) ? 1 : 0);
+}
 
 
 static int pwm_validateTimer(pwm_tim_id_t timer, bool checkInitialized, uint8_t checkDMAFlags)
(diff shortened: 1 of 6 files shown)


multi/stm32l4-multi/pwm_n6.c
```diff
@@ -436,174 +692,96 @@ int pwm_configure(pwm_tim_id_t timer, uint16_t prescaler, uint32_t top)
...
-		*(base + tim_ccer) &= ~(1 << PWM_CCER_CCNE_OFF(chn));
+		tmpccer &= ~(1 << PWM_CCER_CCNE_OFF(chn));
 	}
 
-	dataBarier();
-
-	tmpcr2 = *(base + tim_cr2);
-	tmpccmr = *(base + PWM_CCMR_REG(chn));
-	tmpccer = *(base + tim_ccer);
+	*(base + tim_ccer) = tmpccer;
```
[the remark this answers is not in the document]
jmaksymowicz: I know, tbh if we are worried about concurrent accesses we should mutex all operations (except for updating PWM duty cycle) per-timer. This was also a problem previously.

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -339,20 +569,25 @@ int pwm_disableChannel(pwm_tim_id_t timer, pwm_ch_id_t chn)
...
+	if (((prevChannelOn >> chn) & 1) == 0) {
+		/* Already disabled */
+		return 0;
+	}
+
 	/* Disable the channel. Clear CCxE in CCER */
 	*(pwm_setup.timer[timer].base + tim_ccer) &= ~(1 << PWM_CCER_CCE_OFF(chn));
 	if (pwm_channelHasComplement(timer, chn)) {
 		*(pwm_setup.timer[timer].base + tim_ccer) &= ~(1 << PWM_CCER_CCNE_OFF(chn));
 	}
```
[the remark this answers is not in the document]
jmaksymowicz: see above

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -883,49 +997,473 @@ int pwm_setBitSequence(pwm_tim_id_t timer, pwm_ch_id_t chn, void *data, uint32_t
...
+		*(arg->base + tim_ccmr1) = arg->regs.ccmr[0];
+		*(arg->base + tim_ccmr2) = arg->regs.ccmr[1];
+		*(arg->base + tim_ccer) = arg->regs.ccer;
+		*(arg->base + tim_arr) = arg->regs.arrRx;
+		/* Clear update event flags */
+		*(arg->base + tim_sr) = ~TIM_SR_UIF;
+		*(arg->base + tim_egr) = TIM_EGR_UG; /* Force an update event */
+		while ((*(arg->base + tim_sr) & TIM_SR_UIF) == 0) {
+			/* Wait for timer to have an update event, it's very short (~50 ns) */
+		}
```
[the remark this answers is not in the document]
jmaksymowicz: I don't think this is a real problem. When we force update through `TIM_EGR_UG` the flag gets raised very quickly even if the timer counting is disabled. Ofc we could make up scenarios like "what if someone switches off clock to timer peripheral" or "what if the timer hardware is faulty" but then the drone will fall out of the sky anyway because we can't transmit to ESCs.

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -361,9 +596,24 @@ uint64_t pwm_getBaseFrequency(pwm_tim_id_t timer)
...
+/* Generate update event (UG bit in EGR) to update prescaler, autoreload value and clear counters */
+static void pwm_forceUpdateEvent(pwm_tim_id_t timer)
+{
+	volatile uint32_t *base = pwm_setup.timer[timer].base;
+	*(base + tim_sr) = ~TIM_SR_UIF; /* Clear update event flag */
+	/* Force an update event */
+	*(base + tim_egr) = TIM_EGR_UG;
+	while ((*(base + tim_sr) & TIM_SR_UIF) == 0) {
+		/* Wait for timer to have an update event, it should be very short (~50 ns) */
+	}
```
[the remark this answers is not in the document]
jmaksymowicz: see above

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -59,141 +136,296 @@ typedef struct {
...
+		uint32_t new_bits = (~val & 1) << nBits;
+		val <<= nBits;
+		val |= (new_bits != 0) ? (new_bits - 1) : 0;
+	}
+
+	/* Decode modified GCR into regular GCR. Note that because bit 21 is set, bit 20 will also be set,
+	 * but it's okay because we only care about bits 0..19. */
+	uint32_t gcr = val ^ (val >> 1);
+
+	static int8_t gcrDecodeLookup[32] = {
```
agkaminski:
```suggestion
	static const int8_t gcrDecodeLookup[32] = {
```

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -298,39 +531,47 @@ static int pwm_updateEventIrq(unsigned int n, void *arg)
...
+		return 0;
 	}
-	/* Disable all enabled channels */
-	for (pwm_ch_id_t chn = 0; chn < PWM_CHN_NUM; chn++) {
-		if (((pwm_common.timer[timer].channelOn >> chn) & 1) != 0) {
-			pwm_disableChannel(timer, chn);
-		}
+
+	if (doMutex) {
+		mutexLock(pwm_common.timer[timer].mutex);
```
agkaminski: Might be better to `don't` mutex ever and just do it manually when needed instead or perhaps `_pwm_disableChannelInternal` implementation?

I'm not insisting, it's just not a pattern present in the code base up to date really

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -383,25 +652,44 @@ int pwm_configure(pwm_tim_id_t timer, uint16_t prescaler, uint32_t top)
...
 	}
 
 	/* Fail if one of the channels hasn't been disabled */
-	if (pwm_common.timer[timer].channelOn != 0) {
+	if (atomic_load(&pwm_common.timer[timer].channelOn) != 0) {
 		return -EPERM;
 	}
 
+	mutexLock(pwm_common.timer[timer].mutex);
+	dataBarier();
```
agkaminski: `mutexLock` should be enough

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -436,174 +722,100 @@ int pwm_configure(pwm_tim_id_t timer, uint16_t prescaler, uint32_t top)
...
+	if ((((PWM_TIM_BREAK1_MODE | PWM_TIM_BREAK2_MODE) >> timer) & 1) != 0) {
+		/* Enable main output (MOE) */
+		*(base + tim_bdtr) = (1 << 15);
 	}
-	irq->flag_uevreceived = 0;
-	mutexUnlock(irq->uevlock);
 
+	pwm_forceUpdateEvent(timer);
+
+	dataBarier();
```
agkaminski: `mutexUnlock` should be enough

jmaksymowicz: Yes, I accidentally left some old code in.

multi/stm32l4-multi/pwm_n6.c
jmaksymowicz: Yes, I accidentally left some old code in.

multi/stm32l4-multi/pwm_n6.c
jmaksymowicz: Done

multi/stm32l4-multi/pwm_n6.c
jmaksymowicz: In `pwm_disableChannelInternal` I decided to mutex in the external function, it doesn't matter much because the function isn't used very often. I kept the `doMutex` pattern in `pwm_setInternal` - I understand the concern, but I don't see a way to do it better right now.

multi/stm32l4-multi/stm32l4-multi.h
```diff
@@ -223,6 +223,12 @@ typedef struct {
 } __attribute__((packed)) pwmdischn_t;
 
 
+#define PWM_BITSEQ4_CHN_INVALID (1UL << 15) /* This channel in the array will remain idle */
+#define PWM_BITSEQ4_CHN_ACT_LOW (1UL << 14) /* This channel in the array will be active low */
```
niewim19: Not a breaking change, but dshot nomenclature rarely mentions active line state as polarity determinant.. Most often the idle state is used as differentiation, so IDLE_LOW is Dshot, and IDLE_HIGH is Bi-Dshot. Can we align the nomenclature?

jmaksymowicz: OK, changed name of this flag to `PWM_BITSEQ4_CHN_IDLE_HIGH`.

multi/stm32l4-multi/pwm_n6.c
```diff
@@ -861,71 +1004,542 @@ int pwm_setBitSequence(pwm_tim_id_t timer, pwm_ch_id_t chn, void *data, uint32_t
...
+	if (doRx != 0) {
+		res = pwm_bitSequenceTimerReconfig(timer, doRx, gcrBitTime, &arg);
+		if (res < 0) {
+			mutexUnlock(t->mutex);
+			return res;
+		}
+
+		t->irq.flag_dshot_uev = DSHOT_UEV_DMA_RX_END;
+	}
+
```
agkaminski, line 1448: Missing `mutexUnlock(t->mutex)` below this point

agkaminski (dismissed):

multi/stm32l4-multi/pwm_n6.c
jmaksymowicz: Unlocking mutex is done in `pwm_bitSequenceCleanup`

agkaminski: Can't say I love it, but it is indeed. Thx

agkaminski (approved):

PR #679, opened 2026-07-24
tty/uart16550: fixes and performance improvements

jmaksymowicz:
Multiple fixes to the 16550 driver to improve performance and compatibility with different platforms.

## Description
The 16550 UART driver had multiple problems that were addressed:
1. The AXI UART 16550 v2.0 implementation that was used on Zynq platforms requires that `FCR.FIFOEN` bit be set to 1 first, in a separate write, before FIFO parameters in the same register can be configured. This is not explicitly mentioned in the documentation, only that `FIFOEN` bit must be 1. Now FIFO threshold is set correctly.
2. Compile-time FIFO options were added to allow FIFO threshold to be set per-device (incl. option to turn off FIFO completely). This allows user to configure the UART for a particular use case.
3. Hardware FIFO overrun conditions are now read and accumulated in software and (in debug configuration) signalled through syslog.
4. Previously data within ISR was read into a bespoke buffer implementation. This was replaced with a FIFO implementation from `lf-fifo.h` library. This is in line with other UART drivers and is more reliable - the library was designed to be lock-free using atomic operations.
5. Pushing characters into `libtty` FIFO was reworked so that as many characters as possible are pushed within one lock/unlock pair. This avoids acquiring and releasing the lock for each character separately, which would be bad for performance.
6. Reading data within ISR was extended to read as many characters as are available. This improves performance and helps avoid overruns.
7. Stack size was increased, as the previous stack size was too small for reliable operation on AArch64 platforms.
8. Starting of interrupt handler thread (`uart_intthr`) and registering of interrupt was moved until after hardware initialization is finished. This fixes a race condition that could lead to interrupt mask not being set properly.

In testing on `aarch64-zynqmp-zcu104` these changes eliminate problems with overruns and reduce CPU usage to 37% when UART is used to its maximum configuration (baud rate 460800, receiving characters continuously, 14-character FIFO threshold).

## Motivation and Context
YT: RTOS-1394

## Types of changes

## How Has This Been Tested?

- [x] Tested by hand on: aarch64-zynqmp-som

## Checklist:

## Special treatment

### DIFF
--- a/tty/uart16550/uart16550.c
+++ b/tty/uart16550/uart16550.c
@@ -20,6 +20,9 @@
 #include <unistd.h>
 #include <paths.h>
 #include <string.h>
+#include <stdatomic.h>
+#include <stddef.h>
+#include <lf-fifo.h>
 
 #include <sys/file.h>
 #include <sys/interrupt.h>
@@ -39,13 +42,50 @@
 
 #define SW_BUF_SIZE 64
 
+#define N_UARTS 4
+
+#define THREAD_STACK_SIZE 2048
+
+/*
+ * Note: different 16550 implementations have different meanings for FIFO threshold settings.
+ * Typically 0x2 is 8 characters threshold in a FIFO of 16 characters.
+ */
+#define UART16550_FIFO_DEFAULT 0x2
+
+#ifndef UART16550_FIFO0
+#define UART16550_FIFO0 UART16550_FIFO_DEFAULT
+#endif
+
+#ifndef UART16550_FIFO1
+#define UART16550_FIFO1 UART16550_FIFO_DEFAULT
+#endif
+
+#ifndef UART16550_FIFO2
+#define UART16550_FIFO2 UART16550_FIFO_DEFAULT
+#endif
+
+#ifndef UART16550_FIFO3
+#define UART16550_FIFO3 UART16550_FIFO_DEFAULT
+#endif
+
+
+static const int8_t uart_defaultFifo[N_UARTS] = {
+	UART16550_FIFO0,
+	UART16550_FIFO1,
+	UART16550_FIFO2,
+	UART16550_FIFO3,
+};
+
+
 typedef struct {
 	uint8_t hwctx[64];
-	uint8_t buf[SW_BUF_SIZE];
-	volatile unsigned int buf_i;
+	lf_fifo_t rxSwFifo;
+	uint8_t rxSwFifoData[SW_BUF_SIZE];
+	atomic_uint_fast8_t lineStatus;
 
 	unsigned int init;
 	unsigned int clk;
+	volatile unsigned int hwOverruns; /* Intended to be read by debugger */
 
 	handle_t mutex;
 	handle_t intcond;
@@ -54,13 +94,13 @@ typedef struct {
 	oid_t oid;
 	libtty_common_t tty;
 
-	char stack[1024] __attribute__((aligned(8)));
+	char stack[THREAD_STACK_SIZE] __attribute__((aligned(8)));
 } uart_t;
 
 
 static struct {
-	uart_t uarts[4];
-	char stack[1024] __attribute__((aligned(8)));
+	uart_t uarts[N_UARTS];
+	char stack[THREAD_STACK_SIZE] __attribute__((aligned(8)));
 } uart_common;
 
 
@@ -68,7 +108,7 @@ static uart_t *uart_get(oid_t *oid)
 {
 	unsigned int i;
 
-	for (i = 0; i < sizeof(uart_common.uarts) / sizeof(uart_common.uarts[0]); i++) {
+	for (i = 0; i < N_UARTS; i++) {
 		if ((uart_common.uarts[i].oid.port == oid->port) && (uart_common.uarts[i].oid.id == oid->id)) {
 			return uart_common.uarts + i;
 		}
@@ -136,6 +176,20 @@ static void signal_txready(void *arg)
 	condSignal(uart->intcond);
 }
 
+
+/*
+ * This function is necessary because Line Status Register contains error bits that are clear-on-read
+ * (like overrun bit), but also hardware-cleared bits like data ready bit.
+ * We need to store and accumulate error bits until we are ready to handle them.
+ */
+static uint8_t uart_readUpdateLineStatus(uart_t *uart)
+{
+	uint8_t lineStatus = uarthw_read(uart->hwctx, REG_LSR);
+	atomic_fetch_or(&uart->lineStatus, lineStatus);
+	return lineStatus;
+}
+
+
 #ifdef __TARGET_RISCV64
 __attribute__((section(".interrupt"))) static int uart_interrupt(unsigned int n, void *arg)
 {
@@ -144,8 +198,7 @@ __attribute__((section(".interrupt"))) static int uart_interrupt(unsigned int n,
 	 * Fortunately the UART IRQ on this platform is edge-triggered so we can just
 	 * exit interrupt, signal uart->intcond and main thread will handle the rest.
 	 */
-	uart_t *uart = (uart_t *)arg;
-	return uart->intcond;
+	return 1;
 }
 #else
 static int uart_interrupt(unsigned int n, void *arg)
@@ -162,18 +215,17 @@ static int uart_interrupt(unsigned int n, void *arg)
 	 */
 	uint8_t iir = uarthw_read(uart->hwctx, REG_IIR);
 	uarthw_write(uart->hwctx, REG_IMR, 0);
-	unsigned int i = uart->buf_i;
 	do {
 		uint8_t intr_type = (iir >> 1) & 0x7;
 		if ((intr_type == IIR_CODE_DR) || (intr_type == IIR_CODE_RTO)) {
-			uint8_t c = uarthw_read(uart->hwctx, REG_RBR);
-			if (i < SW_BUF_SIZE) {
-				uart->buf[i] = c;
-				i++;
+			uint8_t status = uart_readUpdateLineStatus(uart);
+			while ((status & LSR_DR) != 0) {
+				lf_fifo_ow_push(&uart->rxSwFifo, uarthw_read(uart->hwctx, REG_RBR));
+				status = uart_readUpdateLineStatus(uart);
 			}
 		}
 		else if (intr_type == IIR_CODE_LS) {
-			uarthw_read(uart->hwctx, REG_LSR);
+			uart_readUpdateLineStatus(uart);
 		}
 		else if (intr_type == IIR_CODE_MS) {
 			uarthw_read(uart->hwctx, REG_MSR);
@@ -182,40 +234,61 @@ static int uart_interrupt(unsigned int n, void *arg)
 		iir = uarthw_read(uart->hwctx, REG_IIR);
 	} while ((iir & IIR_IRQPEND) == 0);
 
-	uart->buf_i = i;
-	return uart->intcond;
+	return 1;
 }
 #endif
+
+
+static void uart_handleErrors(uart_t *uart)
+{
+	(void)uart_readUpdateLineStatus(uart);
+	uint8_t lineStatus = atomic_exchange(&uart->lineStatus, 0);
+	if (lineStatus & LSR_OE) {
+		uart->hwOverruns++;
+	}
+}
 
 
 static void uart_intthr(void *arg)
 {
 	uart_t *uart = (uart_t *)arg;
-	uint8_t target_imr = IMR_DR;
+	uint8_t target_imr = IMR_DR | IMR_LS;
 
 	mutexLock(uart->mutex);
 	for (;;) {
 		uarthw_write(uart->hwctx, REG_IMR, target_imr);
+		uart_handleErrors(uart);
 		condWait(uart->intcond, uart->mutex, 0);
 		/* For the following part the interrupt needs to be masked */
 		uarthw_write(uart->hwctx, REG_IMR, 0);
 
-		/* Empty received buffer */
-		unsigned int buf_i = uart->buf_i;
-		for (unsigned int i = 0; i < buf_i; i++) {
-			libtty_putchar(&uart->tty, uart->buf[i], NULL);
-		}
+		if (!lf_fifo_empty(&uart->rxSwFifo) || (uart_readUpdateLineStatus(uart) & LSR_DR) != 0) {
+			int wake = 0, wakeHelper = 0;
+			libtty_putchar_lock(&uart->tty);
+			/* Empty received buffer */
+			uint8_t c;
+			while (lf_fifo_ow_pop(&uart->rxSwFifo, &c) != 0) {
+				libtty_putchar_unlocked(&uart->tty, c, &wakeHelper);
+				wake |= wakeHelper;
+			}
 
-		uart->buf_i = 0;
-		/* Depending on implementation we may have more characters in hardware FIFO */
-		while ((uarthw_read(uart->hwctx, REG_LSR) & LSR_DR) != 0) {
-			libtty_putchar(&uart->tty, uarthw_read(uart->hwctx, REG_RBR), NULL);
+			/* Depending on implementation we may have more characters in hardware FIFO */
+			while ((uart_readUpdateLineStatus(uart) & LSR_DR) != 0) {
+				c = uarthw_read(uart->hwctx, REG_RBR);
+				libtty_putchar_unlocked(&uart->tty, c, &wakeHelper);
+				wake |= wakeHelper;
+			}
+
+			libtty_putchar_unlock(&uart->tty);
+			if (wake != 0) {
+				libtty_wake_reader(&uart->tty);
+			}
 		}
 
 		/* Check for transmit */
 		while (1) {
 			if (libtty_txready(&uart->tty) != 0) {
-				if ((uarthw_read(uart->hwctx, REG_LSR) & LSR_THRE) != 0) {
+				if ((uart_readUpdateLineStatus(uart) & LSR_THRE) != 0) {
 					uarthw_write(uart->hwctx, REG_THR, libtty_getchar(&uart->tty, NULL));
 				}
 				else {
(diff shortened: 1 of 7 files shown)


tty/uart16550/uart16550.c
```diff
@@ -162,18 +212,17 @@ static int uart_interrupt(unsigned int n, void *arg)
...
-			uint8_t c = uarthw_read(uart->hwctx, REG_RBR);
-			if (i < SW_BUF_SIZE) {
-				uart->buf[i] = c;
-				i++;
-			}
+			int debug_i = 0;
+			do {
+				lf_fifo_ow_push(&uart->rxSwFifo, uarthw_read(uart->hwctx, REG_RBR));
+				debug_i++;
+			} while ((uart_readUpdateLineStatus(uart) & LSR_DR) != 0);
```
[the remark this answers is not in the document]
jmaksymowicz: Fixed

tty/uart16550/uart16550.c
```diff
@@ -162,18 +212,15 @@ static int uart_interrupt(unsigned int n, void *arg)
...
 		uint8_t intr_type = (iir >> 1) & 0x7;
 		if ((intr_type == IIR_CODE_DR) || (intr_type == IIR_CODE_RTO)) {
-			uint8_t c = uarthw_read(uart->hwctx, REG_RBR);
-			if (i < SW_BUF_SIZE) {
-				uart->buf[i] = c;
-				i++;
-			}
+			do {
+				lf_fifo_ow_push(&uart->rxSwFifo, uarthw_read(uart->hwctx, REG_RBR));
+			} while ((uart_readUpdateLineStatus(uart) & LSR_DR) != 0);
```
julianuziemblo: In UART 16550 IP Datasheet (https://caro.su/msx/ocm_de1/16550.pdf) it reads that:
> The [LSR] register informs the user about the status of the transmitter and the receiver. In order to get information about a received character, LSR must be read before reading that received character from RHR. 

We should read the LSR first.

Additionally, error and condition bits (Parity Error Framing Error, Break Interrupt Condition) seem to be queued per-character (however, not the overrun error):

> [The Parity Error] bit is queued in the receiver’s FIFO, so it is associated to the particular character that had the error. Therefore, LSR must be read before RHR: each time a character is read from RHR the next character passes to the top of the FIFO and LSR is loaded with the queued error flags corresponding to this top-of-the-FIFO character.

This part is for more the future though, if we wanted to handle the per-character errors (e.g. with `PARMRK` termios flag) - just a thing I noticed while reading the RM.

tty/uart16550/uart16550.c
```diff
@@ -162,18 +212,15 @@ static int uart_interrupt(unsigned int n, void *arg)
...
 	do {
 		uint8_t intr_type = (iir >> 1) & 0x7;
 		if ((intr_type == IIR_CODE_DR) || (intr_type == IIR_CODE_RTO)) {
-			uint8_t c = uarthw_read(uart->hwctx, REG_RBR);
-			if (i < SW_BUF_SIZE) {
-				uart->buf[i] = c;
-				i++;
-			}
+			do {
+				lf_fifo_ow_push(&uart->rxSwFifo, uarthw_read(uart->hwctx, REG_RBR));
```
julianuziemblo, line 230: Why don't we want to signal the SW Fifo overrun here like the HW overrun and we just overwrite?

tty/uart16550/uart16550.c
```diff
@@ -399,30 +469,39 @@ static int _uart_init(uart_t *uart, unsigned int uartn, unsigned int speed)
...
-	uarthw_write(uart->hwctx, REG_FCR, 0xa7);
-
 	/* Enable hardware interrupts */
 	uarthw_write(uart->hwctx, REG_MCR, MCR_OUT2);
 
 	/* Set interrupt mask */
 	uarthw_write(uart->hwctx, REG_IMR, 0);
 
+	beginthread(uart_intthr, 1, uart->stack, sizeof(uart->stack), uart);
+	interrupt(uarthw_irq(uart->hwctx), uart_interrupt, uart, uart->intcond, &uart->inth);
```
julianuziemblo: maybe we can handle errors from these? if so, we could also probably turn the hardware off somehow (zero-out the control/interrupt mask registers?) on error

tty/uart16550/uart16550.c
```diff
@@ -443,8 +522,8 @@ int main(int argc, char **argv)
 
 	portCreate(&port);
```
julianuziemblo: maybe we can handle errors from this?

tty/uart16550/uart16550.c
```diff
@@ -399,30 +469,39 @@ static int _uart_init(uart_t *uart, unsigned int uartn, unsigned int speed)
 		return err;
 	}
 
-	uart->buf_i = 0;
+	atomic_store(&uart->lineStatus, 0);
+	lf_fifo_init(&uart->rxSwFifo, uart->rxSwFifoData, sizeof(uart->rxSwFifoData));
 	condCreate(&uart->intcond);
 	mutexCreate(&uart->mutex);
```
julianuziemblo: maybe we can handle errors from these?

tty/uart16550/uart16550.c
jmaksymowicz: `lf_fifo_ow_push` doesn't signal overruns. I should probably open an issue about it, I had the same problem in imx6ull-uart: https://github.com/phoenix-rtos/phoenix-rtos-devices/blob/a5b3fe3d83d1d6c147917347b14cfbbb62f80999/tty/imx6ull-uart/imx6ull-uart.c#L411-L415

tty/uart16550/uart16550.c
jmaksymowicz: Maybe you're right, if we want to handle errors like parity per-character then we need to read LSR first.

I did it like this because just the fact we have a data ready/receive timeout interrupt means there's a character in the FIFO, so we don't need to read LSR to find out about it:
> These two interrupt sources refer to the same fact: There is available data to be read from the receiver’s FIFO (or 1-word RHR).

tty/uart16550/uart16550.c
```diff
@@ -182,40 +229,63 @@ static int uart_interrupt(unsigned int n, void *arg)
...
 #endif
 
 
+static void uart_handleErrors(uart_t *uart)
+{
+	(void)uart_readUpdateLineStatus(uart);
+	uint8_t lineStatus = atomic_exchange(&uart->lineStatus, 0);
+	if (lineStatus & LSR_OE) {
+#ifndef NDEBUG
+		syslog(LOG_WARNING, "UART16550: overrun detected");
```
ziemleszcz: Ugly (syslog under mutex). Overruns are reported only by debug build. consider counting overruns instead and reporting them periodically (maybe as a TODO).

tty/uart16550/uart16550.c
```diff
@@ -399,30 +469,39 @@ static int _uart_init(uart_t *uart, unsigned int uartn, unsigned int speed)
...
 
-	beginthread(uart_intthr, 1, uart->stack, sizeof(uart->stack), uart);
-	interrupt(uarthw_irq(uart->hwctx), uart_interrupt, uart, uart->intcond, &uart->inth);
-
 	/* Set speed (MOD) */
 	uarthw_write(uart->hwctx, REG_LCR, LCR_DLAB);
 	uarthw_write(uart->hwctx, REG_LSB, divisor);
 	uarthw_write(uart->hwctx, REG_MSB, divisor >> 8);
 
+	/* Enable and configure FIFOs */
```
ziemleszcz, line 546: Move it after `Set data fromat` to avoid trash data in RX FIFO?

tty/uart16550/uart16550.c
```diff
@@ -399,30 +469,39 @@ static int _uart_init(uart_t *uart, unsigned int uartn, unsigned int speed)
...
+	if (fifo < 0) {
+		uarthw_write(uart->hwctx, REG_FCR, 0);
```
ziemleszcz, line 548: Reset RX/TX FIFO first?

tty/uart16550/uart16550.c
```diff
@@ -399,30 +469,39 @@ static int _uart_init(uart_t *uart, unsigned int uartn, unsigned int speed)
...
+	}
+	else {
+		/* On some implementations we need to set FIFOEN separately before any other write can take effect */
+		uarthw_write(uart->hwctx, REG_FCR, 0x01);
+		uint8_t fifoSetting = (((uint8_t)fifo & 0x3) << 6);
+		uarthw_write(uart->hwctx, REG_FCR, fifoSetting | 0x27);
```
ziemleszcz: Could you describe 0x27 bits in a comment?

tty/uart16550/uart16550.c
ziemleszcz: Run `interrupt()` first before enabling IRQ line in `uart_intthr()`?

jmaksymowicz: A sane interrupt controller should mark the interrupt line as pending until we unmask it within the `interrupt` syscall. But considering how many platforms this driver runs on maybe it's better to register the interrupt first.

tty/uart16550/uart16550.c
jmaksymowicz: Changing FIFOEN from 1 to 0 should reset FIFO according to docs. If FIFOEN was already 0 then there's nothing to reset.

tty/uart16550/uart16550.c
jmaksymowicz: In imx6ull driver errors are also reported while under mutex - also, it's just a mutex for waiting on conditional, so it doesn't block much from happening.

For now I changed it so that errors are only counted to a volatile variable, intended to be read with debugger. Otherwise I'd probably need to spin up another thread to report them periodically. I can't report every overrun in a production build, because the CPU usage from printing messages will likely make the overrun even worse.

tty/uart16550/uart16550.c
jmaksymowicz: Done

tty/uart16550/uart16550.c
jmaksymowicz: Done

tty/uart16550/uart16550.c
jmaksymowicz: Done

tty/uart16550/uart16550.c
jmaksymowicz: Done

tty/uart16550/uart16550.c
jmaksymowicz: Unfortunately one of the bits is implementation-dependent (FCR bit 5) and I added in a comment my best guess why it was set.

I managed to trace the value back to this commit: https://github.com/phoenix-rtos/phoenix-rtos-devices/commit/4407579c24548c0d50c8b6d225cbe9b1058911e7

tty/uart16550/uart16550.c
ziemleszcz: Accepting this solution for now, but consider adding a TODO. We should have a way to detect overruns in the logs, as they can break applications. There is no need to report every overrun individually. Periodic stats are sufficient. We had exactly this issue in DATARO.

tty/uart16550/uart16550.c
```diff
@@ -31,6 +33,7 @@
 #include <libklog.h>
 #include <board_config.h>
 #include <posix/utils.h>
+#include <syslog.h>
```
ziemleszcz, line 37: Unused.

tty/uart16550/uart16550.c
```diff
@@ -415,15 +495,42 @@ static int _uart_init(uart_t *uart, unsigned int uartn, unsigned int speed)
...
+	if (err < 0) {
+		resourceDestroy(uart->inth);
+		uarthw_write(uart->hwctx, REG_MCR, 0);
+		resourceDestroy(uart->mutex);
+		resourceDestroy(uart->intcond);
+		libtty_destroy(&uart->tty);
+		return err;
+	}
+
+	return 0;
```
ziemleszcz, line 585: Nitpick: return EOK.

tty/uart16550/uart16550.c
ziemleszcz: Page allocated by `uarthw_init()` is not freed in error paths below. Consider adding `uarthw_deinit()`.

tty/uart16550/uart16550.c
```diff
@@ -182,40 +232,61 @@ static int uart_interrupt(unsigned int n, void *arg)
...
+				libtty_wake_reader(&uart->tty);
+			}
 		}
 
 		/* Check for transmit */
 		while (1) {
 			if (libtty_txready(&uart->tty) != 0) {
-				if ((uarthw_read(uart->hwctx, REG_LSR) & LSR_THRE) != 0) {
+				if ((uart_readUpdateLineStatus(uart) & LSR_THRE) != 0) {
 					uarthw_write(uart->hwctx, REG_THR, libtty_getchar(&uart->tty, NULL));
```
ziemleszcz: Batch writer wake up similar to the reader?

tty/uart16550/uart16550.c
jmaksymowicz: I thought we were removing EOK from userspace (it should remain in kernel only): https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/641#discussion_r2977612427

tty/uart16550/uart16550.c
ziemleszcz: I'm not sure this comment and bit name is correct anyway. According to TL16C750 datasheet:
`Bit 5: When this bit is set 64-byte mode of operation is selected. When cleared, the 16-byte mode is
selected. A write to FCR bit 5 is protected by setting the line control register (LCR) bit 7 = 1. LCR bit 7 needs
to cleared for normal operation.`

However, `LCR_DLAB` (bit 7) is cleared a few lines earlier (REG_LCR = LCR_D8N1), so this does not enable the extended FIFO mode on the 16750.

On TDA4VM, this bit has a different meaning. Bits 5-4 configure the TX FIFO trigger level (see the TDA4VM Technical Reference Manual). This might be what the original code was trying to configure.

tty/uart16550/uart16550.c
ziemleszcz: I wasn't aware of this. Fine by me.

tty/uart16550/uart16550.c
jmaksymowicz: > This might be what the original code was trying to configure.

Definitely not for TDA4VM. I ported the driver for this platform and I didn't touch the code configuring FIFOs until now. This register value far pre-dates the TDA4VM port.

I concluded it was for enabling a 64-character FIFO mode based on https://www.nxp.com/docs/en/data-sheet/SC16C750B.pdf In this implementation bit 5 is meaningful in regardless of LCR_DLAB or any other configuration mode.

ziemleszcz: OK. Is it fine to to set FIFO trigger level to this value on TDA4VM (not sure if it's RX or TX level)? Shouldn't we add some comment about?

tty/uart16550/uart16550.c
jmaksymowicz: Done

tty/uart16550/uart16550.c
jmaksymowicz: Now used.

tty/uart16550/uart16550.c
jmaksymowicz: I know, but even in the rare case that UART initialization does go wrong, the resource leak here is negligible. We are not allocating a page of actual RAM, just adding a mapping to our page table.

tty/uart16550/uart16550.c
jmaksymowicz: In the end I did add an option to have a thread reporting errors to syslog.

tty/uart16550/uart16550.c
jmaksymowicz: I changed HW initialization so that it reports what value should be in the reserved bits of FCR.

tty/uart16550/uarthw.h
```diff
@@ -16,7 +16,13 @@
 #ifndef _DEV_UART16550_UARTHW_H_
 #define _DEV_UART16550_UARTHW_H_
 
-#include <stdio.h>
+#include <stdint.h>
```
ziemleszcz, line 19: Nitpick: add <stddef.h> for `size_t`.

tty/uart16550/uart16550.c
```diff
@@ -182,51 +240,87 @@ static int uart_interrupt(unsigned int n, void *arg)
...
+	}
+}
+
+
+static bool uart_handleTx(uart_t *uart)
+{
+	bool waitForTx = false;
+	int wake = 0;
+	while (libtty_txready(&uart->tty) != 0) {
+		if ((uart_readUpdateLineStatus(uart) & LSR_THRE) != 0) {
```
ziemleszcz, line 264: This could be optimized further. THRE is set when TX FIFO is empty and is cleared after writing just one character. Knowing TX FIFO size, you could safely write multiple characters without checking THRE after each one.

The FIFO size could be added to `uarthw_info_t`.

tty/uart16550/uart16550.c
```diff
@@ -387,64 +506,133 @@ static int _uart_init(uart_t *uart, unsigned int uartn, unsigned int speed)
...
+
+			default:
+				printf("UART16550 driver. Usage:\n");
+				printf("%s [-n] [-e]\n", argv[0]);
+				printf("\t-n - Skip creating %s device (default: no)\n", _PATH_CONSOLE);
+				printf("\t-e - Report data errors to syslog (default: no)\n");
+				return -1;
+		}
+	}
+
```
ziemleszcz, line 620: Print usage and return -1 on `optind != argc` to reject fake arguments.

tty/uart16550/uart16550.c
jmaksymowicz: FIFO size depends on the exact implementation of 16550 - on a fixed platform like TDA4VM, or when using AXI UART 16550 we know exactly what it is, but e.g. on ia32 PCs it could be different on each machine and require auto-detection.

Can you file a separate issue for this? This PR is already quite far into scope creep.

ziemleszcz: RTOS-1413

tty/uart16550/uart16550.c
jmaksymowicz: Done

tty/uart16550/uarthw.h
jmaksymowicz: Done

ziemleszcz (approved):

PR #693, opened 2026-09-02
!libtty: fix races and refactor

julianuziemblo:
YT: RTOS-1355

## Description
Consolidate the TTY TX and RX locks into one lock that can be provided
externally by the initializer (e.g. UART driver). This lock is used for
all operations on the RX and TX queue and for waiting on RX and TX
condvars. If no lock is provided, libtty creates its own lock. The TTY
lock can be locked and unlocked by added libtty_lock and libtty_unlock
internal (HW) interface functions.

Make driver-facing functions unsynchronized (they have to be called
under the provided in initialization TTY lock) - their names were
changed to start with `_`: _libtty_putchar, _libtty_popchar and
_libtty_getchar.

Add unsynchronized versions of libtty_read, libtty_read_nonblock and
libtty_write (same names starting with `_`) to use when the driver
already locks the provided lock (e.g. in PTY).

## Motivation and Context
Wanting to add further features to libtty, I wanted to minimize the possibilities for error.

Depends-On: phoenix-rtos-posixsrv:julianuziemblo/RTOS-1355

## Types of changes

- [x] Chore (refactoring, style fixes, git/CI config, submodule management, no code logic changes)

## How Has This Been Tested?

- [x] Tested by hand on: `ia32-generic-qemu`

## Checklist:

## Special treatment

- [x] This PR needs additional PRs to work:
    - https://github.com/phoenix-rtos/phoenix-rtos-build/pull/292,
    - https://github.com/phoenix-rtos/phoenix-rtos-posixsrv/pull/34
    - https://github.com/phoenix-rtos/phoenix-rtos-utils/pull/265
    - https://github.com/phoenix-rtos/libphoenix/pull/527

### DIFF
--- a/tty/libtty/libtty.c
+++ b/tty/libtty/libtty.c
@@ -43,11 +43,11 @@
 #define LOG_TAG "libtty: "
 
 /* clang-format off */
-#define log_debug(fmt, ...)     do { if (0) printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
-#define log_ioctl(fmt, ...)     do { if (0) printf(COL_CYAN LOG_TAG "IOCTL: " fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
-#define log_info(fmt, ...)      do { if (0) printf(COL_CYAN LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
-#define log_warn(fmt, ...)      do { if (1) printf(COL_YELLOW LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
-#define log_error(fmt, ...)     do { if (1) printf(COL_RED  LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_debug(fmt, ...) do { if (0) printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
+#define log_ioctl(fmt, ...) do { if (0) printf(COL_CYAN LOG_TAG "IOCTL: " fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_info(fmt, ...)  do { if (0) printf(COL_CYAN LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_warn(fmt, ...)  do { if (1) printf(COL_YELLOW LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_error(fmt, ...) do { if (1) printf(COL_RED  LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
 /* clang-format on */
 
 /* } DEBUG */
@@ -74,48 +74,10 @@
 
 #define TX_FIFO_NOTFULL_WATERMARK 16 /* amount of free space in fifo before we will wake up the writer */
 
-static void termios_optimize(libtty_common_t *tty)
-{
-	/* check break characters list */
-	tty->breakchars[0] = CNL;
-	int n = 1;
 
-	if (tty->term.c_cc[VEOF] != _POSIX_VDISABLE)
-		tty->breakchars[n++] = tty->term.c_cc[VEOF];
-
-	if (tty->term.c_cc[VEOL] != _POSIX_VDISABLE)
-		tty->breakchars[n++] = tty->term.c_cc[VEOL];
-
-	tty->breakchars[n] = '\0';
-
-	/* check if we have break char in the RX FIFO */
-	tty->t_flags &= ~TF_HAVEBREAK;
-	if (CMP_FLAG(l, ICANON)) {
-		if (libttydisc_rx_have_breakchar(tty))
-			tty->t_flags |= TF_HAVEBREAK;
-	}
-}
-
-static void termios_init(struct termios *term, speed_t speed)
-{
-	memset(term, 0, sizeof(*term));
-
-	term->c_iflag = TTYDEF_IFLAG & TTYSUP_IFLAG;
-	term->c_oflag = TTYDEF_OFLAG & TTYSUP_OFLAG;
-	term->c_lflag = TTYDEF_LFLAG & TTYSUP_LFLAG;
-	term->c_cflag = TTYDEF_CFLAG;
-
-	term->c_ispeed = speed;
-	term->c_ospeed = speed;
-
-	memcpy(term->c_cc, ttydefchars, sizeof(ttydefchars));
-}
-
+#if 0
 static void termios_print_flags(const struct termios *termios_p)
 {
-
-	return;
-
 	log_info("TERMIOS :");
 	log_info("input flags: (0x%x)", termios_p->c_iflag);
 	log_info("	INLCR  = %u // Translate NL to CR on input.", (termios_p->c_iflag & INLCR) != 0);
@@ -145,34 +107,73 @@ static void termios_print_flags(const struct termios *termios_p)
 			log_info("	[%2u]    = %u", i, termios_p->c_cc[i]);
 #endif
 }
+#else
+#define termios_print_flags(termios_p)
+#endif
+
+
+static void termios_optimize(libtty_common_t *tty)
+{
+	/* check break characters list */
+	tty->breakchars[0] = CNL;
+	int n = 1;
+
+	if (tty->term.c_cc[VEOF] != _POSIX_VDISABLE) {
+		tty->breakchars[n++] = tty->term.c_cc[VEOF];
+	}
+
+	if (tty->term.c_cc[VEOL] != _POSIX_VDISABLE) {
+		tty->breakchars[n++] = tty->term.c_cc[VEOL];
+	}
+
+	tty->breakchars[n] = '\0';
+
+	/* check if we have break char in the RX FIFO */
+	tty->t_flags &= ~TF_HAVEBREAK;
+	if (CMP_FLAG(l, ICANON)) {
+		if (libttydisc_rx_have_breakchar(tty)) {
+			tty->t_flags |= TF_HAVEBREAK;
+		}
+	}
+}
+
+
+static void termios_init(struct termios *term, speed_t speed)
+{
+	memset(term, 0, sizeof(*term));
+
+	term->c_iflag = TTYDEF_IFLAG & TTYSUP_IFLAG;
+	term->c_oflag = TTYDEF_OFLAG & TTYSUP_OFLAG;
+	term->c_lflag = TTYDEF_LFLAG & TTYSUP_LFLAG;
+	term->c_cflag = TTYDEF_CFLAG;
+
+	term->c_ispeed = speed;
+	term->c_ospeed = speed;
+
+	memcpy(term->c_cc, ttydefchars, sizeof(ttydefchars));
+}
 
 
 ssize_t libtty_read(libtty_common_t *tty, char *data, size_t size, unsigned mode)
 {
-	ssize_t ret = 0;
-
-	if (tty->t_flags & TF_CLOSING)
-		return -EBADF;
-
-	if (CMP_FLAG(l, ICANON))
-		ret = libttydisc_read_canonical(tty, data, size, mode, NULL);
-	else
-		ret = libttydisc_read_raw(tty, data, size, mode, NULL);
-
-	return ret;
+	return libtty_read_nonblock(tty, data, size, mode, NULL);
 }
+
 
 ssize_t libtty_read_nonblock(libtty_common_t *tty, char *data, size_t size, unsigned mode, libtty_read_state_t *st)
 {
 	ssize_t ret = 0;
 
-	if (tty->t_flags & TF_CLOSING)
+	if ((tty->t_flags & TF_CLOSING) != 0) {
 		return -EBADF;
+	}
 
-	if (CMP_FLAG(l, ICANON))
+	if (CMP_FLAG(l, ICANON)) {
 		ret = libttydisc_read_canonical(tty, data, size, mode, st);
-	else
+	}
+	else {
 		ret = libttydisc_read_raw(tty, data, size, mode, st);
+	}
 
 	return ret;
 }
@@ -181,7 +182,7 @@ ssize_t libtty_read_nonblock(libtty_common_t *tty, char *data, size_t size, unsi
 /* writer wake up is done outside of libtty if wake_writer is not NULL */
 unsigned char libtty_getchar(libtty_common_t *tty, int *wake_writer)
 {
-	unsigned char c = fifo_pop_back(tty->tx_fifo);
+	unsigned char c = libtty_popchar(tty);
 
 	if (wake_writer != NULL) {
 		*wake_writer = fifo_freespace(tty->tx_fifo) >= TX_FIFO_NOTFULL_WATERMARK;
@@ -230,17 +231,35 @@ int libtty_init(libtty_common_t *tty, libtty_callbacks_t *callbacks, unsigned in
 		return -1;
 	}
 
-	if (condCreate(&tty->tx_waitq) != EOK)
+	if (condCreate(&tty->tx_waitq) != EOK) {
+		free(tty->tx_fifo);
+		free(tty->rx_fifo);
 		return -1;
+	}
 
-	if (condCreate(&tty->rx_waitq) != EOK)
+	if (condCreate(&tty->rx_waitq) != EOK) {
+		resourceDestroy(tty->tx_waitq);
+		free(tty->tx_fifo);
+		free(tty->rx_fifo);
 		return -1;
+	}
 
-	if (mutexCreate(&tty->tx_mutex) != EOK)
+	if (mutexCreate(&tty->tx_mutex) != EOK) {
+		resourceDestroy(tty->tx_waitq);
+		resourceDestroy(tty->rx_waitq);
+		free(tty->tx_fifo);
+		free(tty->rx_fifo);
 		return -1;
+	}
 
-	if (mutexCreate(&tty->rx_mutex) != EOK)
+	if (mutexCreate(&tty->rx_mutex) != EOK) {
+		resourceDestroy(tty->tx_waitq);
+		resourceDestroy(tty->rx_waitq);
+		resourceDestroy(tty->tx_mutex);
+		free(tty->tx_fifo);
+		free(tty->rx_fifo);
 		return -1;
+	}
 
 	fifo_init(tty->tx_fifo, bufsize);
 	fifo_init(tty->rx_fifo, bufsize);
@@ -291,12 +310,15 @@ ssize_t libtty_write(libtty_common_t *tty, const char *data, size_t size, unsign
 	ssize_t len = 0;
 
 	/* short path */
-	if (tty->t_flags & TF_CLOSING)
+	if ((tty->t_flags & TF_CLOSING) != 0) {
 		return -EPIPE;
-	else if (fifo_is_full(tty->tx_fifo) && (mode & O_NONBLOCK))
+	}
+	else if (fifo_is_full(tty->tx_fifo) && (mode & O_NONBLOCK) != 0) {
 		return -EWOULDBLOCK;
-	else if (size == 0)
+	}
+	else if (size == 0) {
 		return 0;
+	}
 
 	mutexLock(tty->tx_mutex);
 
@@ -305,11 +327,13 @@ ssize_t libtty_write(libtty_common_t *tty, const char *data, size_t size, unsign
 	/* write contents of the buffer */
 	while (len < size) {
 		while (fifo_freespace(tty->tx_fifo) < fifo_freespace_for_single_char) {
-			if (tty->t_flags & TF_CLOSING)
+			if (tty->t_flags & TF_CLOSING) {
 				goto exit;
+			}
 
-			if (mode & O_NONBLOCK)
+			if (mode & O_NONBLOCK) {
 				goto exit;
+			}
 
 			CALLBACK(signal_txready);
 			condWait(tty->tx_waitq, tty->tx_mutex, 0);
@@ -335,16 +359,18 @@ ssize_t libtty_write(libtty_common_t *tty, const char *data, size_t size, unsign
 #endif
 
 exit:
-
-	if (tty->t_flags & TF_CLOSING)
+	if ((tty->t_flags & TF_CLOSING) != 0) {
 		len = -EPIPE;
-	else if ((len == 0) && (mode & O_NONBLOCK))
+	}
+	else if ((len == 0) && (mode & O_NONBLOCK) != 0) {
 		len = -EWOULDBLOCK;
+	}
 
 	mutexUnlock(tty->tx_mutex);
 
 	return len;
 }
+
 
 int libtty_txready(libtty_common_t *tty)
 {
@@ -359,15 +385,18 @@ int libtty_txready(libtty_common_t *tty)
 	return !fifo_is_empty(tty->tx_fifo);
 }
 
+
 int libtty_txfull(libtty_common_t *tty)
 {
 	return fifo_is_full(tty->tx_fifo);
 }
 
+
 int libtty_rxready(libtty_common_t *tty)
 {
 	return !fifo_is_empty(tty->rx_fifo);
 }
+
 
 int libtty_poll_status(libtty_common_t *tty)
 {
@@ -375,21 +404,27 @@ int libtty_poll_status(libtty_common_t *tty)
 
 	/* poll in ICANON mode should return POLLIN only if breakchar is present */
 	if (!CMP_FLAG(l, ICANON)) {
-		if (libtty_rxready(tty))
+		if (libtty_rxready(tty)) {
 			revents |= POLLIN | POLLRDNORM;
+		}
 	}
 	else {
-		if (tty->t_flags & TF_HAVEBREAK)
+		if ((tty->t_flags & TF_HAVEBREAK) != 0) {
 			revents |= POLLIN | POLLRDNORM;
+		}
 	}
 
-	if (!libtty_txfull(tty))
+	if (!libtty_txfull(tty)) {
 		revents |= POLLOUT | POLLWRNORM;
-	if (tty->t_flags & TF_CLOSING)
+	}
+
+	if ((tty->t_flags & TF_CLOSING) != 0) {
 		revents |= POLLHUP;
+	}
 
 	return revents;
 }
+
 
 void libtty_signal_pgrp(libtty_common_t *tty, int signal)
 {
@@ -399,15 +434,18 @@ void libtty_signal_pgrp(libtty_common_t *tty, int signal)
 	}
 }
 
-void libtty_drain(libtty_common_t *tty)
+
+static void libtty_drain(libtty_common_t *tty)
 {
 	mutexLock(tty->tx_mutex);
-	while (!fifo_is_empty(tty->tx_fifo))
+	while (!fifo_is_empty(tty->tx_fifo)) {
 		condWait(tty->tx_waitq, tty->tx_mutex, 0);
-
+	}
 	mutexUnlock(tty->tx_mutex);
 }
-void libtty_flush(libtty_common_t *tty, int type)
+
+
+static void libtty_flush(libtty_common_t *tty, int type)
 {
 	if (type == TCIFLUSH || type == TCIOFLUSH) {
 		mutexLock(tty->rx_mutex);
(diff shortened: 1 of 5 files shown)


tty/libtty/libtty.c
```diff
@@ -419,14 +462,17 @@ void libtty_flush(libtty_common_t *tty, int type)
 		/* leaving one char in TX fifo should allow us to avoid */
 		/* undefined behaviour if writer is in the middle of operation */
 		mutexLock(tty->tx_mutex);
-		fifo_remove_all_but_one(tty->tx_fifo);
+		/* TODO: transfer to lf-fifo.h: lf_fifo_clear_leave_last */
+		unsigned int head = atomic_load_explicit(&tty->tx_fifo.head, memory_order_relaxed);
+		atomic_store_explicit(&tty->tx_fifo.tail, head == 0 ? tty->tx_fifo.mask : head - 1, memory_order_relaxed);
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist you're mistaking `lf_fifo_t` from `libtty-lf-fifo.h` with the one from `lf-fifo.h` from `phoenix-rtos-corelibs`. I agree the former should be transitioned to the latter, but this is a case for another PR.

tty/libtty/libtty.c
```diff
@@ -181,10 +182,10 @@ ssize_t libtty_read_nonblock(libtty_common_t *tty, char *data, size_t size, unsi
 /* writer wake up is done outside of libtty if wake_writer is not NULL */
 unsigned char libtty_getchar(libtty_common_t *tty, int *wake_writer)
 {
-	unsigned char c = fifo_pop_back(tty->tx_fifo);
+	unsigned char c = libtty_popchar(tty);
 
 	if (wake_writer != NULL) {
-		*wake_writer = fifo_freespace(tty->tx_fifo) >= TX_FIFO_NOTFULL_WATERMARK;
+		*wake_writer = lf_fifo_free(&tty->tx_fifo) >= TX_FIFO_NOTFULL_WATERMARK;
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist as mentioned, you're mistaking things. Get your facts together, please.

tty/libtty/libtty_disc.c
```diff
@@ -357,13 +369,14 @@ void libtty_wake_reader(libtty_common_t *tty)
...
+	mutexUnlock(tty->tx_mutex);
 	mutexUnlock(tty->rx_mutex);
 }
 
 
 int libtty_putchar_unlocked(libtty_common_t *tty, unsigned char c, int *wake_reader)
 {
-	return libtty_putchar_helper(tty, c, wake_reader, 0);
+	return libtty_putchar_helper(tty, c, wake_reader, false);
 }
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist as the name suggests, it is the UNLOCKED version, so we assume the caller handles the synchronization there by calling the `libtty_putchar_{lock|unlock}` functions!

julianuziemblo:
drafting for now to test this on DTR and other projects

tty/libtty/libtty.c
```diff
@@ -419,14 +473,17 @@ void libtty_flush(libtty_common_t *tty, int type)
 		/* leaving one char in TX fifo should allow us to avoid */
 		/* undefined behaviour if writer is in the middle of operation */
 		mutexLock(tty->tx_mutex);
-		fifo_remove_all_but_one(tty->tx_fifo);
+		/* TODO: transfer to lf-fifo.h: lf_fifo_clear_leave_last */
+		unsigned int head = atomic_load_explicit(&tty->tx_fifo.head, memory_order_relaxed);
+		atomic_store_explicit(&tty->tx_fifo.tail, head == 0 ? tty->tx_fifo.mask : head - 1, memory_order_relaxed);
 		mutexUnlock(tty->tx_mutex);
```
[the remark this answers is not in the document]
julianuziemblo: @gemini-code-assist we always lock TX lock on producer as there are multiple threads. UART thread doesn't lock the lock, so that end is a single consumer (usually -- PTY now locks externally-available libtty_getchar_lock). I changed this mechanism to atomically acquire the tail and then release the head as `tail + 1 & mask`. It's still not the best solution, but at least we're now writing the head as the producer.

There's a TODO for transferring this to `lf-fifo.h`, which is in another repo.

nalajcie:
not sure: maybe we should instead add MPMC or MPSC FIFO implementations instead of using lock-free fifo with additional mutexes? MPMC implementation for non-SMP is easy, for SMP needs additional care to avoid cache-related problems.

julianuziemblo:
> not sure: maybe we should instead add MPMC or MPSC FIFO implementations instead of using lock-free fifo with additional mutexes? MPMC implementation for non-SMP is easy, for SMP needs additional care to avoid cache-related problems.

That was my first thought also, decided to first implement the easier way because of deadlines, but will look into if MPMC would also be doable

tty/libtty/libtty.c
```diff
@@ -219,6 +238,18 @@ int libtty_init(libtty_common_t *tty, libtty_callbacks_t *callbacks, unsigned in
...
+	}
+	else {
+		/* create the mutex */
+		if (mutexCreate(&tty->lock) < 0) {
+			return -1;
+		}
+		tty->lockCreated = true;
+	}
+
 	memset(tty, 0, sizeof(*tty));
```
adamgreloch: memset after already setting fields

multi/imxrt-multi/uart.c
```diff
@@ -320,7 +321,7 @@ static void uart_intrThread(void *arg)
 
 		/* RX */
 		while (lf_fifo_pop(&uart->rxFifoCtx, &c) != 0) {
-			libtty_putchar(&uart->tty_common, c & mask, NULL);
+			_libtty_putchar(&uart->tty_common, c & mask, NULL);
 		}
```
adamgreloch, line 323: no tty->lock, unsafe?

tty/libtty/libtty-lf-fifo.h
```diff
@@ -1,105 +0,0 @@
...
-{
-	atomic_init(&f->headPos, 0);
-	atomic_init(&f->tail, 0);
-	f->sizeMask = size - 1;
-	f->data = data;
-}
-
-
-/* return 0 in case of overrun, 1 otherwise */
-static inline int lf_fifo_push(lf_fifo_t *f, uint8_t byte)
```
adamgreloch, line 57: Note that libalgo `lf_fifo_push` has different semantics - it doesn't overwrite old data. You likely want to use `lf_fifo_ow_push` if the change was unintended

tty/libtty/libtty.c
```diff
@@ -359,38 +416,47 @@ int libtty_txready(libtty_common_t *tty)
...
 }
 
+
 int libtty_rxready(libtty_common_t *tty)
 {
 	return !fifo_is_empty(tty->rx_fifo);
 }
 
+
 int libtty_poll_status(libtty_common_t *tty)
```
adamgreloch: libtty_poll_status is still unsynchronized

tty/libtty/libtty.c
```diff
@@ -399,35 +465,39 @@ void libtty_signal_pgrp(libtty_common_t *tty, int signal)
...
 
 	/* check for breakchars, etc. */
 	termios_optimize(tty);
+
+	return 0;
 }
 
-int libtty_ioctl(libtty_common_t *tty, pid_t sender_pid, unsigned int cmd, const void *in_arg, const void **out_arg)
+
+int _libtty_ioctl(libtty_common_t *tty, pid_t sender_pid, unsigned int cmd, const void *in_arg, const void **out_arg)
```
adamgreloch: I very dislike how ioctl hands out internal pointers through `out_arg`, because this creates a race between pool threads handling ioctls (RTOS-1450 issue). Since you are already messing with the API to fix races, perhaps you could already fix this as well?

We likely need to reverse the handout, so that it is the caller-provided buffer, not the tty's one.

tty/libtty/libtty.h
```diff
@@ -107,27 +128,39 @@ int libtty_ioctl(libtty_common_t *tty, pid_t sender_pid, unsigned int cmd, const
...
+/* protected by libtty_lock */
+/* writer/reader wake up is done outside of libtty if wake_{writer|reader} is not NULL */
+int _libtty_putchar(libtty_common_t *tty, unsigned char c, int *wake_reader);
+unsigned char _libtty_popchar(libtty_common_t *tty);
+unsigned char _libtty_getchar(libtty_common_t *tty, int *wake_writer);
+
+
 int libtty_txready(libtty_common_t *tty); /* at least 1 character ready to be sent */
 int libtty_txfull(libtty_common_t *tty);  /* no more place in the TX buffer */
 int libtty_rxready(libtty_common_t *tty); /* at least 1 character ready to be read out */
```
adamgreloch: synchronization undocumented, drivers seem to synchronize the calls however they want. Worth fixing as well.

Same goes for `libtty_wake_writer`

tty/pc-tty/ttypc_bioskbd.c
```diff
@@ -34,11 +34,12 @@ static void ttypc_bioskbd_ctlthr(void *arg)
...
 
-		mutexLock(ttypc->lock);
-		mutexLock((cvt = ttypc->vt)->lock);
+		cvt = ttypc->vt;
+
+		mutexLock2(ttypc->lock, cvt->lock);
 
 		/* Add the keystroke and update head position */
-		libtty_putchar(&ttypc->vt->tty, *(volatile unsigned char *)((uintptr_t)ttypc->kbd + 0x400 + head), NULL);
+		_libtty_putchar(&ttypc->vt->tty, *(volatile unsigned char *)((uintptr_t)ttypc->kbd + 0x400 + head), NULL);
```
adamgreloch:
```suggestion
		_libtty_putchar(&cvt->tty, *(volatile unsigned char *)((uintptr_t)ttypc->kbd + 0x400 + head), NULL);
```

tty/pc-tty/ttypc_vt.c
adamgreloch: requires external sync now, prefix with `_`

tty/imx6ull-uart/imx6ull-uart.c
```diff
@@ -534,7 +534,7 @@ static void uart_process_tx(void)
 		}
 
 		/* FIXME: potential data race on tx_fifo (lock-free access) */
```
adamgreloch: FIXME still valid?

tty/libtty/libtty.c
```diff
@@ -335,17 +380,29 @@ ssize_t libtty_write(libtty_common_t *tty, const char *data, size_t size, unsign
 #endif
```
adamgreloch: `tx_mutex` no longer exists.

Suggestion: `#define LIBTTY_DEBUG 0` and replace `if 0` in libtty with `if LIBTTY_DEBUG` if it makes sense. If it doesn't, just drop the dead code

multi/imxrt-multi/cm4.c
```diff
@@ -21,7 +21,7 @@
 #include <sys/interrupt.h>
 #include <sys/threads.h>
 
-#include <libtty-lf-fifo.h>
+#include "lf-fifo.h"
```
adamgreloch: This is now a corelibs header, so should be `#include <lf-fifo.h>`. Applies to the rest as well

multi/mcxn94x-multi/uart.c
```diff
@@ -182,12 +183,12 @@ static void uart_intrThread(void *arg)
 
 		/* RX */
 		while (lf_fifo_pop(&uart->rxFifoCtx, &c) != 0) {
-			libtty_putchar(&uart->tty_common, c & mask, NULL);
+			_libtty_putchar(&uart->tty_common, c & mask, NULL);
```
ziemleszcz, line 184: Unprotected, lock released at  L182.

multi/imxrt-multi/uart.c
ziemleszcz: Extend lock released at L320 ro RX/TX sections?

tty/libtty/libtty.c
ziemleszcz: Yep, copy the data to user buffer.

tty/libtty/libtty.c
julianuziemblo: doesn't make much sense for some as e.g. this one is an untested feature (which I'll probably add back in the next PR), so I'll drop it for those cases

tty/libtty/libtty.h
```diff
@@ -107,27 +128,39 @@ int libtty_ioctl(libtty_common_t *tty, pid_t sender_pid, unsigned int cmd, const
...
+int _libtty_ioctl(libtty_common_t *tty, pid_t sender_pid, unsigned int cmd, const void *in_arg, const void **out_arg);
+
+
 /* internal (HW) interface */
-int libtty_putchar(libtty_common_t *tty, unsigned char c, int *wake_reader);
-void libtty_putchar_lock(libtty_common_t *tty);
-void libtty_putchar_unlock(libtty_common_t *tty);
+
+
 void libtty_wake_reader(libtty_common_t *tty);
```
ziemleszcz: Make these two protected by `libtty_lock`? Almost all calls are already protected. The missing ones are `tty_dmahandleRx()`, `tty_dmahandleTx()` and `signal_txready()`.

tty/libtty/libtty_disc.c
```diff
@@ -96,37 +92,40 @@
...
-#define CTL_ALNUM(c) (((c) >= '0' && (c) <= '9') || \
-	((c) >= 'a' && (c) <= 'z') || ((c) >= 'A' && (c) <= 'Z'))
+#define CTL_ALNUM(c) (((c) >= '0' && (c) <= '9') || ((c) >= 'a' && (c) <= 'z') || ((c) >= 'A' && (c) <= 'Z'))
+
 
 /* writing chars to TX buffer without any futher processing */
-static int tx_write_ifspace(libtty_common_t *tty, const char *data, size_t len)
+static int _libttydisc_txFeedback(libtty_common_t *tty, const char *data, size_t len)
 {
 	/* WARN: no locking */
```
ziemleszcz: Obsolete comment.

tty/libtty/libtty-lf-fifo.h
julianuziemblo: the libalgo lf_fifo OW API seems to have some quirks that are smoothed out in https://github.com/phoenix-rtos/phoenix-rtos-corelibs/pull/68 which is not a priority change. I think I'll drop this change for now and open a draft PR with it (https://github.com/phoenix-rtos/phoenix-rtos-devices/pull/698)

tty/libtty/libtty.c
```diff
@@ -399,50 +465,52 @@ void libtty_signal_pgrp(libtty_common_t *tty, int signal)
...
 
 	/* check for breakchars, etc. */
 	termios_optimize(tty);
+
+	return 0;
 }
 
-int libtty_ioctl(libtty_common_t *tty, pid_t sender_pid, unsigned int cmd, const void *in_arg, const void **out_arg)
+
+int _libtty_ioctl(libtty_common_t *tty, pid_t sender_pid, unsigned long cmd, const void *in_arg, void *out_arg)
```
adamgreloch, line 504: add `out_len` and pass `IOCPARM_LEN(request)` in callers? seems like `out_arg` can point to msg o.raw or o.data, no?

multi/grlib-multi/uart.c
```diff
@@ -331,12 +327,13 @@ static void uart_ioctl(msg_t *msg, int dev)
...
+	void *outData = NULL;
 
-	inData = ioctl_unpack(msg, &req, NULL);
+	inData = ioctl_unpackEx(msg, &req, NULL, &outData);
 	pid = ioctl_getSenderPid(msg);
 
-	err = libtty_ioctl(&uart_common.uart[dev].tty, pid, req, inData, &outData);
+	err = libtty_ioctl(&uart_common.uart[dev].tty, pid, req, inData, outData);
 
 	ioctl_setResponse(msg, req, err, outData);
```
adamgreloch, line 336: note that when taking outData from ioctl_unpackEx, ioctl_setResponse essentially does memcpy from outData to itself. Applies to all uses

tty/libtty/libtty.c
```diff
@@ -43,59 +44,91 @@
...
-#define log_debug(fmt, ...)     do { if (0) printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
-#define log_ioctl(fmt, ...)     do { if (0) printf(COL_CYAN LOG_TAG "IOCTL: " fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
-#define log_info(fmt, ...)      do { if (0) printf(COL_CYAN LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
-#define log_warn(fmt, ...)      do { if (1) printf(COL_YELLOW LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
-#define log_error(fmt, ...)     do { if (1) printf(COL_RED  LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_debug(fmt, ...) do { if (0) printf(LOG_TAG fmt "\n", ##__VA_ARGS__); } while (0)
+#define log_ioctl(fmt, ...) do { if (0) printf(COL_CYAN LOG_TAG "IOCTL: " fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_info(fmt, ...)  do { if (0) printf(COL_CYAN LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_warn(fmt, ...)  do { if (1) printf(COL_YELLOW LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
+#define log_error(fmt, ...) do { if (1) printf(COL_RED  LOG_TAG fmt "\n" COL_NORMAL, ##__VA_ARGS__); } while (0)
```
adamgreloch: these can deadlock in libtty_ioctl (because of obtaining tty->lock now) with a single pool thread if a driver writes to its own stdout. Please add a comment warning about this, and maybe defer the log_warn/log_error usages outside of tty->lock sections

tty/libtty/libtty.c
adamgreloch, line 504: add `out_arg` null check

tty/libtty/libtty_disc.c
```diff
@@ -325,68 +315,39 @@ static int libtty_putchar_helper(libtty_common_t *tty, unsigned char c, int *wak
...
-{
-	mutexLock(tty->rx_mutex);
-}
-
-
-void libtty_wake_reader(libtty_common_t *tty)
+/*
+ * this doesn't need synchronization now, but in the future we
+ * might need it to e.g. check the RX queue capacity
+ */
```
adamgreloch, line 332: comment is contradictory with the declaration in libtty.h

multi/grlib-multi/uart.c
adamgreloch: dead code

tty/imx6ull-uart/imx6ull-uart.c
adamgreloch, line 569: dead code

adamgreloch (commented):
the "fix style and formatting" commit is not style and formatting only

tty/libtty/libtty.c
julianuziemblo: we pass the request (cmd) anyway so we can do IOCPARM_LEN ourselves. We assume that `out_arg` is a writeable buffer of exactly that size

multi/grlib-multi/uart.c
julianuziemblo: hmm, maybe I'll introduce `ioctl_setResponse` which only sets the error and `ioctl_setResponseEx` which actually copies the provided data to an output buffer? this way we don't have to pass the temporary pointer to `libtty_ioctl`.

tty/libtty/libtty.c
adamgreloch: You cannot assume that, as this is provided by the IPC sender. ioctl_unpackEx may set the outData to msg->o.raw (64 B), and on TCGETS this will overflow. The driver is thus vulnerable to clients that intentionally undersize msg size

adamgreloch: Previously this was not an issue, because the ioctl result was copied back through ioctl_setResponse

multi/grlib-multi/uart.c
adamgreloch: you may also handle NULL instead to mean "skip the copy" so that you can write:
```
ioctl_setResponse(msg, req, err, NULL);
```

tty/libtty/libtty.c
julianuziemblo: IOCPARM_LEN is sometimes too little (IOC_NESTED ioctls). Maybe another check should be added to `ioctl_unpackEx` and/or `ioctl_setResponse`

tty/libtty/libtty_disc.c
julianuziemblo: yes - the API lets the user know that we need synchronization with the lock to use this function. Whether that's true is in the library's hands - for now it is not, but we might want to implement a mechanism similar to `libtty_getchar` where the reader is woken up in the future, that's why this comment is here for future maintainers.

adamgreloch: I'm not against the comment, only its phrasing. Maybe rephrase it in a way that acknowledges the contract, something like "Currently, this would be safe without synchronization, but the API sync contract is future-proofed to allow e.g. (insert predictable reasons)"?

Just to stress the "it works without sync but it's nowhere guaranteed and you should sync it anyways" part

tty/libtty/libtty-lf-fifo.h
adamgreloch: Ack

tty/libtty/libtty.c
julianuziemblo: why? i thought we always want to handle the `out_arg` in `libtty_ioctl`, so if it's not provided with buffer of the exact size wanted by `cmd`, we shouldn't handle that case?

tty/pc-tty/ttypc_vt.c
```diff
@@ -670,7 +670,8 @@ int ttypc_vt_ioctl(ttypc_vt_t *vt, pid_t pid, unsigned int cmd, const void *idat
 			vt->fbmode = mode;
 			break;
 		case FBCONGETMODE:
-			*odata = (const void *)&vt->fbmode;
+			mode = vt->fbmode;
+			memcpy(odata, &mode, sizeof(mode));
```
ziemleszcz, line 674: `odata` will be NULL for `FBCONGETMODE` because it's incorrectly defined as `_IO()`. Probably requires fix in libphoenix.

Before this PR odata was local variable.

multi/grlib-multi/uart.c
```diff
@@ -331,12 +325,13 @@ static void uart_ioctl(msg_t *msg, int dev)
...
```
ziemleszcz: That's not the correct usage of `ioctl_setResponse()`. `outData` should point to a local buffer or be set to NULL if the buffer was taken by `ioctl_unpackEx()`. As it is, you're essentially copying the message buffer to itself. The same applies to the other locations.

tty/libtty/libtty.c
```diff
@@ -43,59 +44,95 @@
...
-/* NOT supported: IXON|IXOFF|IXANY|PARMRK|INPCK|IGNPAR */
-#define TTYSUP_IFLAG (IGNBRK | BRKINT | ISTRIP | INLCR | IGNCR | ICRNL | IMAXBEL)
-
-#define TTYSUP_OFLAG (OPOST | ONLCR | TAB3 | OCRNL | ONOCR | ONLRET)
-/* NOT supported: TOSTOP|FLUSHO|NOFLSH|ECHOPRT */
-#define TTYSUP_LFLAG (ECHOKE | ECHOE | ECHOK | ECHO | ECHONL | ECHOCTL | ISIG | ICANON | IEXTEN)
+/* NOT supported: IGNBRK|BRKINT|IMAXBEL|IXON|IXOFF|IXANY|PARMRK|INPCK|IGNPAR */
+#define TTYSUP_IFLAG (ISTRIP | INLCR | IGNCR | ICRNL)
+/* NOT supported: ONOCR|ONLRET|NLDLY|CRDLY|TABDLY|BSDLY|VTDLY|FFDLY */
+#define TTYSUP_OFLAG (OPOST | ONLCR | TAB3 | OCRNL | ONLRET)
```
ziemleszcz: `ONLRET` is listed as not supported?

tty/pc-tty/ttypc_vt.c
julianuziemblo: already done in https://github.com/phoenix-rtos/libphoenix/pull/527, will merge parallel to this

tty/libtty/libtty.c
julianuziemblo: I don't see where is it supported

ziemleszcz: It's literally added to `TTYSUP_OFLAG` and and the same time commented as not supported.

tty/libtty/libtty_disc.c
```diff
@@ -316,7 +306,7 @@ static int libtty_putchar_helper(libtty_common_t *tty, unsigned char c, int *wak
 			if (wake_reader != NULL) {
 				*wake_reader = 1;
 			}
-			if (lock != 0) {
+			else {
 				condSignal(tty->rx_waitq);
```
ziemleszcz: Use `_libtty_wake_reader() ` (safer if it changes in the future).

tty/libtty/libtty.c
julianuziemblo: ah okay, yeah it shouldn't be there, thanks, didn't see this

tty/libtty/libtty_disc.c
```diff
@@ -325,68 +315,41 @@ static int libtty_putchar_helper(libtty_common_t *tty, unsigned char c, int *wak
 		if (wake_reader != NULL) {
 			*wake_reader = 1;
 		}
-		if (lock != 0) {
+		else {
 			condSignal(tty->rx_waitq);
```
ziemleszcz: Use `_libtty_wake_reader() ` (safer if it changes in the future).

tty/libtty/libtty.c
```diff
@@ -43,59 +44,95 @@
...
 		if (tty->cb.cb_name != NULL) \
 			tty->cb.cb_name(tty->cb.arg, ##__VA_ARGS__); \
 	} while (0)
 
-#if 0
+#define TX_FIFO_NOTFULL_WATERMARK 16 /* amount of free space in fifo before we will wake up the writer */
+
+
+#if LIBTTY_DEBUG
 #define DEBUG_CHAR(c) \
```
ziemleszcz: Do we need this? There is no `debug` field in `tty`.

tty/pc-tty/ttypc_vt.h
```diff
@@ -151,19 +151,19 @@ extern ssize_t ttypc_vt_write(ttypc_vt_t *vt, int mode, const char *buff, size_t
...
+extern int _ttypc_vt_respond(ttypc_vt_t *vt, const char *buff);
 
 
 /* Polls virtual terminal status */
 extern int ttypc_vt_pollstatus(ttypc_vt_t *vt);
 
 
 /* Virtual terminal ioctl */
-extern int ttypc_vt_ioctl(ttypc_vt_t *vt, pid_t pid, unsigned int cmd, const void *idata, const void **odata);
+extern int ttypc_vt_ioctl(ttypc_vt_t *vt, pid_t pid, unsigned int cmd, const void *idata, void *odata);
```
ziemleszcz: Nitpick: unsigned long cmd

tty/pc-tty/ttypc_vt.c
```diff
@@ -651,7 +651,7 @@ int ttypc_vt_pollstatus(ttypc_vt_t *vt)
 }
 
 
-int ttypc_vt_ioctl(ttypc_vt_t *vt, pid_t pid, unsigned int cmd, const void *idata, const void **odata)
+int ttypc_vt_ioctl(ttypc_vt_t *vt, pid_t pid, unsigned int cmd, const void *idata, void *odata)
```
ziemleszcz: Nitpick: unsigned long cmd

tty/pc-tty/ttypc_vt.h
julianuziemblo: GH shows that you're on outdated version somehow - this was fixed a few days ago

ziemleszcz: Strange.

tty/libtty/libtty.c
ziemleszcz: Bump.

julianuziemblo: sorry, just removed it

ziemleszcz (approved):

